PR #37, opened 2014-08-28
Use H5BP server configs

austinpray:
This has been tested locally but I would like to try to deploy it on digital ocean to see if it runs alright on a remote server (shouldn't be an issue)

Any feedback on this?

---

Checks out a stable version of h5bp/server-configs-nginx and applies it
to the /etc/nginx folder

http://goo.gl/pBfB4W

Misc:
Adds "restart nginx" handler

closes https://github.com/roots/bedrock-ansible/issues/31

### DIFF
--- a/roles/nginx/defaults/main.yml
+++ b/roles/nginx/defaults/main.yml
@@ -1,0 +1,4 @@
+---
+logs_root: /var/log/nginx/
+nginx_user: www-data
+strip_www: true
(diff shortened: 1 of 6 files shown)


teohhanhui:
Would be good to have the `expires` issue sorted out... Is there any downside when including it?

swalkinshaw:
@teohhanhui the downside is that you might not get the latest version of the files if you don't have cache busting in place. This [comment](https://github.com/roots/bedrock-ansible/pull/37#issuecomment-53863182) still applies.

PR #45, opened 2014-08-31
HHVM provisioning and toggling between php-fpm

nathanielks:
* Renamed `php-55` to `php-fpm` as it made sense ( to me ) to remove the
version number from the role as it may change in the future, and
php-fpm is a more accurate representation of what we’re doing (
especially considering we don’t have version numbers in any of the
other roles ).

* Added base php role. This will load proper php implementation in
`meta/main.yml` based upon the variable `php_implementation` located in
`vars/common.yml`. The tasks in the php role are executed after said role.

* Added `nginx-php-fpm.conf` and `nginx-hhvm.conf` to allow decoupling of
how nginx passes the request back to php. I believe the hhvm
implementation can be used for both, though we’d have to change from a
socket connection to a tcp connection. Should we choose, hhvm can be
used over a unix socket and we can standardize the file name structure
as well so as to not have to keep changing the nginx php location block.


nathanielks:
@swalkinshaw I forget where we landed on this...

roles/hhvm/tasks/main.yml
```diff
@@ -0,0 +1,21 @@
...
+  apt_repository: repo="deb http://dl.hhvm.com/ubuntu {{ ansible_distribution_release }} main"
+                  state=present
+
+- name: Update apt
+  apt: update_cache=yes
+
+- name: Install HHVM
+  apt: pkg="{{ item }}" state=latest
+  with_items:
+    - hhvm
```
swalkinshaw: No need for `with_items` here. Just hardcore hhvm in there.

swalkinshaw:
@nathanielks me too. I'm thinking we should just ditch separate php-fpm confs/sockets per WP site. Then HHVM and PHP-FPM can both be treated in the same way with 1 socket.

swalkinshaw:
Closing in favour of https://github.com/roots/bedrock-ansible/pull/136

nathanielks:
Whoop whoop!

PR #95, opened 2014-11-16
Ansible based site deploys

swalkinshaw:
This is a work in-progress to deploy sites with Ansible instead of Capistrano through Bedrock itself. The main benefits are:

1. Removes Ruby requirements
2. Ansible can run the deploy itself easily
3. Re-using hosts/settings from Ansible configs

Todo:

- [x] `README`/docs update
- [x] Improve separation of `wordpress-sites` role
- [x] Figure out `.env` generation. Create on provision + on each deploy
- [x] Figure out workflow for installing remote WP sites after deploys
- [x] sudoers for running commands like restarting php5-fpm
- [x] Consistent `logs` folder location/creating for dev/remote
- [x] Update `deploy_user` to `web_user`
- [x] Proper solution for cloning private repos


sugarskull:
@swalkinshaw, is there a possible merge happening here? Also, seems to me that the Readme file needs to be modified to include a note that composer will be failing at this point and that the user should deploy using capistrano before proceeding. Seems like having the task included for production doesn't make sense... errr why have a production script that we know will fail? Thanks for the hard work in keeping this repo up.

Still getting (when using the master branch):

```TASK: [wordpress-sites | Install Dependencies with Composer] ******************    

stdout: Composer could not find a composer.json file in /srv/www/website.com/current```

**Comment by swalkinshaw** [comment]
@sugarskull there's still a few big issues to figure out before this is merged. I'll try and get a REAMDE update in the meantime to document the current workflow gap.

**Comment by retlehs** [comment]
:shipit:

File: roles/users/defaults/main.yml [hunk 1]
```diff
@@ -1,5 +1,3 @@
 deploy_user: deploy
 deploy_group: deploy
 deploy_groups: www-data
```
fullyint: If `deploy_user` etc. are defined in `group_vars/all`, could you remove `roles/users/defaults/main.yml`?

roles/wordpress-install/tasks/main.yml
```diff
@@ -0,0 +1,47 @@
+---
+- include: directories.yml
+
+- name: Create .env file
+  template: src="env.j2"
+            dest="{{ www_root }}/{{ item.key }}/current/.env"
+            owner="{{ item.value.user }}"
+            group="{{ item.value.group }}"
```
fullyint: I'm pretty sure `item.value.user` and `item.value.group` will be undefined now that `user` and `group` are now removed from the `wordpress_sites` dict in `group_vars/development`.

roles/users/defaults/main.yml
swalkinshaw: Yeah I probably can. That was left over since I needed those vars to be globally available.

roles/wordpress-install/tasks/main.yml
swalkinshaw: Good catch, I removed these in a few other places but missed this one. I'll do a search for any other occurrences.

fullyint:
Now that you've removed all instances of `default(user)`, I'm pretty sure you can remove `user: 'vagrant'` from `ansible.extra_vars = {}` in the `Vagrantfile`. I don't see `user` used anywhere else.

fullyint:
I can't find where `github_ssh_keys` is defined. Maybe it's not defined so as to not trigger the `github-ssh-keys` role. How about defining it (in `group_vars/all`?), but commenting it out, just so people have an example demonstrating structure? Or maybe you'd prefer the example to be in the README.

fullyint:
Most importantly, THANK YOU so much for all your work on this @swalkinshaw !

When `github_ssh_keys` is not defined...

**1. `github-ssh-keys | Add SSH keys` task fails.** Despite the `when` condition on the `github-ssh-keys` role: 
`fatal: [192.168.50.5] => subelements lookup expects a dictionary, got 'github_ssh_keys'`. [Try this](http://grokbase.com/t/gg/ansible-project/148mbx1ykv/skip-a-task-with-subelements-if-its-dictionary-is-undefined):
```yaml
  with_subelements:
     - github_ssh_keys | default([])
     - authorized
```

**2. `ssh-known-hosts` role fails.** I think it's because the `/home/deploy/.ssh` directory doesn't yet exist:
`IOError: [Errno 2] No such file or directory: '/home/deploy/.ssh/known_hosts'` 
I suppose you could create the `.ssh` directory at the end of `roles/users/tasks/main.yml`:
```yaml
- name: Create deploy user .ssh directory
  file: path="/home/{{ deploy_user }}/.ssh"
        owner="{{ deploy_user }}"
        group="{{ deploy_group }}"
        mode=0700
        state=directory
```
Or, make the `ssh-known-hosts` role also conditional on `when: github_ssh_keys is defined`, but maybe that wouldn't fit with your plans.  

I assume `group_vars/all` may end up including something like this?
```yaml
github_ssh_keys:
  - username: myGithubUser
    authorized:
      - deploy
```

swalkinshaw:
@fullyint pushed up a few little tweaks/fixes. Thanks for your help.

Oddly enough I never get a failure for `ssh-known-hosts` event without `github_ssh_keys` defined.

But I do need to document its usage. You guessed correctly at what it would look like.

fullyint:
Staging/production lack logs directory, maybe because d49beb4 removed logs directory from `project_shared_children`. Could move 'Create logs folder of sites' task from
`roles/wordpress-install/tasks/directories.yml` to
`roles/wordpress-setup/tasks/main.yml`.

roles/deploy/tasks/main.yml
```diff
@@ -0,0 +1,76 @@
...
+  with_items: project_post_build_commands
+  environment: project_environment
+
+- name: Finalize the deploy
+  deploy_helper: path={{ project_root }} release={{ deploy_helper.new_release }} state=finalize
+  when: project_finalize
+
+- name: Reload php-fpm
+  service: name=php5-fpm state=reloaded
+  sudo: True
```
fullyint: Does deploy user need to be granted passwordless sudo for `service php5-fpm reload`? Or maybe you want to add `sudo_user: admin` and users will include `--ask-sudo-pass`? Sorry if I'm missing something.

deploy.yml
```diff
@@ -0,0 +1,24 @@
+---
+- name: Deploy WP site
+  hosts: web
+  remote_user: deploy
```
fullyint: I wonder if you actually want this:
`remote_user: "{{ deploy_user }}"`
but I may be getting ahead of myself.

roles/deploy/tasks/main.yml
swalkinshaw: This is the main thing that still needs to be done before this is merged. Ran into issues with this on roots.io because of how Ansible runs `sudo` commands. Normally you could add a rule like `deploy ALL=(root) NOPASSWD: /usr/sbin/php5-fpm` and it would be fine.

Problem is Ansible runs the command like `sudo -k && sudo -H -S -p "[sudo via ansible, key=qdpnqhzmilpnttuggtaksahbdfzzsstz] password: " -u root /bin/sh -c '"<command>'"`.

Which is insane and I couldn't figure out how to do this.

deploy.yml
swalkinshaw: Yeah good catch. Realized last night we may want to rename `deploy_user` to something a little more generic like `web_user`. Because in development there is no "deploy" and it's just the `vagrant` user.

fullyint:
**1. `remote_user: root` in `server.yml`.** Won't it fail to connect? (`secure-root.yml` sets `PermitRootLogin no`)

**2. Private repos.** Cloning private repos via https, Ansible hangs waiting for password. Clone via ssh?
```diff
wordpress_sites:
  staging.example.com:
-    repo: https://github.com/roots/bedrock.git
+    repo: git@github.com/roots/bedrock.git
```
Some seem to need this in `bedrock-ansible/ansible.cfg` (I seem to). Maybe worth mentioning. 
```
[ssh_connection]
ssh_args = -o  ForwardAgent=yes
```
Could note `ForwardAgent yes` in `~/.ssh/config`.

**3. `known_hosts` as `root:root`.** I think I noticed the deploy_user's `known_hosts` file as `root:root`. If so, would the deploy user ever need to add known_hosts independently but not be able to?

swalkinshaw:
@fullyint 

1. `secure-root.yml` is still kind of optional right now so we definitely need better integration with it.

2. I ran into issues getting forward agent to work as well and I believe for roots.io we just ended up using a deploy key but that isn't really ideal. I added a todo about this but we need to figure out one complete method and document it.

3. Yes, good find! Originally I was using the deploy_module role from Galaxy and ran into this [issue](https://github.com/f500/ansible-project_deploy/issues/12) so that known hosts role was needed. Now that I've just included the deploy role and customized it, it's no longer needed.

roles/deploy/tasks/main.yml
fullyint: Oh. DeHaan ([Apr '14](https://groups.google.com/forum/#!msg/ansible-project/YHtTkipp1Ts/FQ8s5BOiPKoJ)): "I'm sorry, you can't define sudo access to do specific commands this way."

swalkinshaw: That's really terrible. Wonder if there's a more direct way if you use `shell` or `command` and do:

`sudo service php5-fpm reload`

fullyint: I was just trying to figure that out. :smile: Some people store a sudo password in ansible vault, but seems less than ideal.

swalkinshaw: I got it working with this:

Sudoers file:

`web ALL=(root) NOPASSWD:/usr/sbin/service php5-fpm *`

Ansible:

`shell: "sudo service php5-fpm reload"`

fullyint: :tada: :tada: :tada:

fullyint:
**Rsyncing compiled assets.** I plan to compile my theme locally and will probably add a custom [rsync](http://docs.ansible.com/synchronize_module.html) task instead of specifying `project_files` for the "Copy project files" task. That way I can use filename matching patterns. I haven't checked whether the `copy` module can do that. I suspect my approach doesn't belong in the official project, but just to share...

**1. Add `rsync_theme` to `group_vars/<staging|production>`.** It's optional, fine to omit.
```diff
  wordpress_sites:
    staging.example.com:
      repo: git@github.com:roots/bedrock.git
+     rsync_theme: ../mytheme # path targeting local theme directory (relative to bedrock-ansible root)
```
I don't store/commit the theme in my bedrock repo. They have separate repos. I symlink local theme into local bedrock and Vagrant is none the wiser. I do the same with the client's custom plugin. 
```
├── sites
    ├── client_name
        ├── example.dev (bedrock)
        ├── bedrock-ansible
        ├── mytheme
        └── myplugin
```
There are probably better ways. Feedback welcome.

**2. Add custom rsync task.** Right after "Copy project files" task (`roles/deploy/tasks/main.yml`):
```
- name: Rsync theme files
  synchronize: src="{{ project.rsync_theme }}"
               dest="{{ deploy_helper.new_release_path }}/web/app/themes"
               rsync_timeout=30
               archive=no
               recursive=yes
               delete=yes
               rsync_opts="--include-from={{ project.rsync_theme }}/.rsync"
  when: project.rsync_theme is defined
```
Maybe I'll precede this with a task to compile theme assets locally with a [local_action](http://docs.ansible.com/playbooks_delegation.html#delegation).

**3. Add `.rsync` file to `mytheme` root.** A bunch of exclude rules, followed by "include all the rest" (`+*`).
```
- .git
- assets
- bower_components
- node_modules
- .bowerrc
- .editorconfig
- .jshintrc
- .jscsrc
- .travis.yml
- bower.json
- composer.json
- CONTRIBUTING.md
- package.json
- .gitignore
- .gitkeep
- .DS_Store

+ *
```

mAAdhaTTah:
@swalkinshaw Is this PR at a testable state? I've got a side-project website that I am going to be deploying soon, which would be a good opportunity to test the full deployment and provisioning process on a non-mission critical server.

@fullyint I could use something like that for the same side-project actually, so if you want any help, I'd love to pitch in.

swalkinshaw:
@mAAdhaTTah other people testing it would be great. I think the only known "big" issue right now is a solution for cloning private repos. So other than doing that manually (either via SSH keys or get SSH forwarding working), it should be mostly good.

The only thing I haven't really documented is that the new `server.yml` playbook won't actually install WordPress like `dev.yml` does. Obviously it can't until the site is deployed.

fullyint:
@swalkinshaw These steps work for me in cloning a private bitbucket repo. Don't they work for you?

**1.** Switch `wordpress_sites` to clone via ssh: `repo: git@github.com:roots/my-private-bedrock.git`
**2.** Amend `bedrock-ansible/ansible.cfg` file (my comment above about this was muddled):
```
[ssh_connection]
ssh_args = -o ForwardAgent=yes -o ControlMaster=auto -o ControlPersist=60s
```
**3.** And maybe you need something like this in your `~/.ssh/config` locally:
```
Host example.com
  User myusername
  ForwardAgent yes
```

@mAAdhaTTah The sync task I mentioned (for theme assets compiled locally) is working using the steps I mentioned. If you beat me to writing a task to compile theme locally, I won't complain.

### Notes for Testing
This PR seems totally testable. I've only tested on DigitalOcean (not vagrant), and not yet with commits since d49beb4. For me, testing a staging site looks about like this:
- `ansible-galaxy install -r requirements.yml`
- In `server.yml` switch to `sudo: yes` and `remote_user: admin`
- Adjust various `group_vars` (leave `deploy_user: deploy` for now, but var may change to `web-user`)
- Add `github_ssh_keys` to `group_vars/all` (formatted like bottom of [this comment](https://github.com/roots/bedrock-ansible/pull/95#issuecomment-85285761)) 
- do numbers 1, 2, 3 above re: private repos
- optional: add the rsync task I mentioned for locally compiled theme (see steps above) 
- Spin up DO droplet. Clear IP from `known_hosts`. Add IP to `hosts/staging` and `wordpress_sites`.
- `ansible-playbook secure-root.yml -i hosts/staging -vvvv`
- `ansible-playbook server.yml -i hosts/staging -vvvv -K` (`-K` is short for `--ask-sudo-pass`)
- `ansible-playbook deploy.yml -i hosts/staging -vvvv -e "site=staging.example.com"`
(`-e` is `--extra-vars`)

I wonder if a [prompt](http://docs.ansible.com/playbooks_prompts.html) would be a convenient alternative to specifying that `site` variable for the `deploy.yml` playbook (vs. using the `--extra-vars "site=example.com"`).

swalkinshaw:
@fullyint thanks, that's good to know it's mostly working. I'll test out SSH forwarding as well and then just make the small change + document it.

I agree that Ansible's `extra-vars` is clunky. I was thinking of just creating a simple bash wrapper script so you'd just do: `deploy production example.com` for example.

group_vars/development
```diff
@@ -1,12 +1,13 @@
 mysql_root_password: devpw
 
+deploy_user: vagrant
+
 wordpress_sites:
-  - site_name: example.dev
+  example.dev:
     site_hosts:
       - example.dev
+    repo: https://github.com/roots/bedrock.git
```
mAAdhaTTah: Does the repo need to be included in development?

fullyint: I think this `repo` variable is only referenced in the `deploy.yml` playbook, and `deploy.yml` isn't used for development. Development continues using Vagrant sync folders for its bedrock.

This means there is a little less parity between development and production environments, in how they are provisioned. But I'm guessing it's better than problems that might arise by trying for parity on this issue.

mAAdhaTTah:
I ran through a deploy and everything is looking pretty good overall, although I'm getting a hang on the Reload php-fpm step of the deploy process.

group_vars/development
swalkinshaw: Removed

swalkinshaw:
If it's hanging I'm guessing it's asking for a sudo password and waiting. Did you have the latest updates? Specifically this one: https://github.com/roots/bedrock-ansible/commit/6fc553e4057dcf041446da8ec2c06f8f9338300b

roles/wordpress-setup/tasks/database.yml
```diff
@@ -0,0 +1,36 @@
...
+              login_host="{{ item.value.env.db_host | default('localhost') }}"
+              login_user="{{ mysql_user }}"
+              login_password="{{ mysql_root_password }}"
+  with_dict: wordpress_sites
+  when: item.db_create | default(True)
+
+- name: Copy database dump
+  copy: src="{{ item.value.db_import }}" dest=/tmp
+  with_dict: wordpress_sites
+  when: item.db_import | default(False)
```
mAAdhaTTah: This has to be `item.value.db_import` or it'll never fire. It appears a couple places in this file.

mAAdhaTTah:
@swalkinshaw Yeah, I just pulled it down this morning, so I've got that. I ran it with `-vvvv` and didn't see anything.

group_vars/development
fullyint: Sorry @mAAdhaTTah  I didn't read your comment carefully. I had thought `repo` was already missing from development and you were asking why. I was wrong about what you were saying, and wrong about what was in the code.

roles/wordpress-setup/tasks/database.yml
swalkinshaw: Thanks, fixed them all.

group_vars/development
mAAdhaTTah: @fullyint lol that explains why I was so confused. No worries.

mAAdhaTTah:
Besides that, everything works very well. The bash wrapper would be a plus, as there are a lot of commands to run. Using that to combine `secure-root.yml` then `server.yml` for provisioning and run `deploy production example.com` for deployment would make this perfect.

mAAdhaTTah:
One more thing: @fullyint's solution does work for cloning private repos, with the addition to `ansible.cfg`.

mAAdhaTTah:
@fullyint Also, the rsync task works as well. Is that definitely not being included in core?

fullyint:
The theme and other files can currently be synced using the deploy role's "Copy project files" task. However, I don't think it allows syncing only files matching filename patterns . Maybe the task could add options like those in that rsync task (merging the tasks). That might mean using the synchronize module instead of the copy module. I need to refresh my memory on the differences between the two modules.

mAAdhaTTah:
> I symlink local theme into local bedrock and Vagrant is none the wiser.

This doesn't work for me; did you do anything to make Vagrant follow the symlinks?

fullyint:
@mAAdhaTTah Uh oh. I was wrong again! I originally had played with symlinking the theme into bedrock. Looks like I actually create additional vagrant sync folders. I gotta run (can't write much, or carefully), but I replace the sync stuff in my Vagrantfile with something like this:
```ruby
# Sync bedrock directory
  config.vm.synced_folder '../example.dev', '/vagrant-nfs/bedrock', type: 'nfs'
  config.bindfs.bind_folder '/vagrant-nfs/bedrock', '/srv/www/example.dev/current', u: 'vagrant', g: 'www-data'

  # Optionally sync plugin on host INSTEAD of having Composer install it on guest from remote repo.
  config.vm.synced_folder '../example', '/vagrant-nfs/plugin', type: 'nfs'
  config.bindfs.bind_folder '/vagrant-nfs/plugin', '/srv/www/example.dev/current/web/app/plugins/example', u: 'vagrant', g: 'www-data'

  # Optionally sync theme on host INSTEAD of having Composer install it on guest from remote repo.
  config.vm.synced_folder '../exampletheme', '/vagrant-nfs/theme', type: 'nfs'
  config.bindfs.bind_folder '/vagrant-nfs/theme', '/srv/www/example.dev/current/web/app/themes/exampletheme', u: 'vagrant', g: 'www-data'
```

swalkinshaw:
@mAAdhaTTah I think for now there will just be a script for deploying. Next step after merging this would be to merge `secure-root.yml` into `server.yml`. It's confusing to have it optional right now and leads to inconsistencies. Plus it's just a good practice anyway.

@fullyint forgot to reply about your project setup. You didn't mention if your theme/plugin are being re-used across different projects. My recommendation is always to bundle your theme in your project repo. The only exception is if it's being across multiple sites/projets (and not just MU). And even then you're probably likely to have a base parent theme as a Composer dependency and then small site specific themes in their repos.

And for @fullyint and @mAAdhaTTah this leads into the discussion about what will be included by default in deploys. @fullyint's rsync process is "non-standard" to me so it won't be included.

What's currently missing from this project (and the new deploys) is a solution for asset compilation. There's a couple of options:

1. Add packages like Node.js + npm + bower + gulp/grunt globally by default and make `project_post_build_commands` more flexible to accept a `chdir` for each command (I'll probably do this anyway).

2. Include a way for people to easily run **local** actions during the deploy to generate assets and then copy/upload them to the remote server.

3. Let the two options above be toggable. Although I'd rather stick to one.

swalkinshaw:
@fullyint @mAAdhaTTah added better pre/post commands in https://github.com/roots/bedrock-ansible/commit/2623132b0d5f7e3258c5f3e88b223d69b2808d98. Now you can pass an optional sub directory to run the command in :)

swalkinshaw:
Added the wrapper script.

Usage: `./deploy production example.com`

mAAdhaTTah:
@fullyint The mount bind works, thanks.

> My recommendation is always to bundle your theme in your project repo. 

All the themes I use on my personal sites are open-sourced, so if there was a way to do this such that it's not bundled with the project, that would be awesome.

Definitely prefer option 2. I think the less we can install on the server, the better (even that's what I ended up doing for one of my sites).

roles/deploy/defaults/main.yml
```diff
@@ -0,0 +1,83 @@
...
+project_environment: {}
+
+# There are a few moments in this role where arbitrary command(s) can be run. These commands receive
+# the "project_environment" so deploys for different stages can be done by changing this.
+# Each command is a dict with a `cmd` and an optional `path` which is relative to the `new_release_path`.
+# Example:
+# project_post_build_commands:
+#   - path: web/app/themes/mytheme
+#     cmd: npm install && gulp --productino
+#   - cmd: wp db import posts.sql
```
fullyint: gulp --productino

roles/deploy/tasks/main.yml
```diff
@@ -0,0 +1,80 @@
...
+
+- name: Run post_build_commands in the new_release_path
+  command: "{{ item.cmd }}"
+  args:
+    chdir: "{{ [deploy_helper.new_release_path, item.path] | select('defined') | join('/') }}"
+  with_items: project_post_build_commands
+  environment: project_environment
+
+- name: Finalize the deploy
+  deploy_helper: path={{ project_root }} release={{ deploy_helper.new_release }} state=finalize
```
fullyint: Want quotes on "{{ vars }}" in "Finalize the deploy"? Some possible others...
- above in "Copy vendor dir if exists to speed up composer".
- `roles/deploy/defaults/main.yml` (in comments for `project_finalize`)
- `roles/ferm/tasks/main.yml` (various)
- `roles/nginx/tasks/main.yml` (in "Copy over h5bp configuration")

swalkinshaw: Thanks I'll fix these up. Just a note that I only use quotes when the value *starts* in `{{`.

From their docs:

> Further, Ansible uses “{{ var }}” for variables. If a value after a colon starts with a “{”, YAML will think it is a dictionary, so you must quote it, like so:

It actually doesn't matter too much right now we're using `key=val` everywhere. But once this is merged I'm going to convert all tasks to YAML.

Example:

Before
```yaml
deploy_helper: path={{ project_root }} release={{ deploy_helper.new_release }} state=finalize
```

After
```yaml
deploy_helper:
  path: "{{ project_root }}"
  release: "{{ deploy_helper.new_release }}"
  state: finalize
```

fullyint: Ah ha! I get it now. I was hazy on when quotes were needed.

swalkinshaw:
Renamed `deploy_user` to `web_user`. Default user name is also now `web`.

roles/deploy/tasks/main.yml
```diff
@@ -0,0 +1,80 @@
...
+
+- name: write unfinished file
+  file: path="{{ project_source_path }}/{{ deploy_helper.unfinished_filename }}" state=touch
+
+- name: Copy files to new build dir
+  command: cp -pr {{ project_source_path }} {{ deploy_helper.new_release_path }}
+
+- name: Move site into root
+  shell: mv {{ deploy_helper.new_release_path }}/{{ project_subtree }}/* {{ deploy_helper.new_release_path }}
+  when: project_subtree is defined
```
fullyint: I don't understand this "Move site into root" task. For me, it just skips because `project_subtree` is not defined.

swalkinshaw: A common setup including one that we use for roots.io is this dir tree:

```
roots.io/
  ansible/
  roots.io/
```

ie: your repo has two top level folders: ansible/playbooks (customized bedrock-ansible), and your actual site/Bedrock project. So we have `project_subtree: roots.io` and it just moves that up a level.

fullyint: Thanks. That makes sense. I hadn't imagined the cloned repo would be anything but bedrock alone (vs. including ansible etc.). Thanks for this additional insight into project setup, and for your comments earlier re: committing theme to bedrock repo. I need to digest what you said, and reread  the various discourse threads on that because I feel drawn to modularity of keeping them separate (even though I'm not reusing across multiple sites/projects).

roles/deploy/tasks/main.yml
```diff
@@ -0,0 +1,80 @@
...
+  git: repo="{{ project_git_repo }}"
+       dest="{{ project_source_path }}"
+       version="{{ project_version }}"
+       accept_hostkey=yes
+
+- name: write unfinished file
+  file: path="{{ project_source_path }}/{{ deploy_helper.unfinished_filename }}" state=touch
+
+- name: Copy files to new build dir
+  command: cp -pr {{ project_source_path }} {{ deploy_helper.new_release_path }}
```
fullyint: I'll think about it more carefully in a bit, but could you just clone the project straight into `deploy_helper.new_release_path` instead of cloning to one dir then copying to another? I think f500's version may have cloned to `project_source_path` just to coordinate with `project_deploy_strategy: "synchronize"`, but you've omitted that.

swalkinshaw: I kind of want to keep this role as close as possible to the original. It's going to be included in Ansible core modules sometime soon and a few changes I've made might be done upstream as well.

fullyint: I understand that [f500/ansible-project_deploy_module](https://github.com/f500/ansible-project_deploy_module) is the proposed deploy_helper module for ansible-modules-extra. However, the tasks in question (Clone project files, Copy files to new build dir) come from [f500/ansible-project_deploy](https://github.com/f500/ansible-project_deploy).

I wasn't under the impression that this latter was also destined for ansible-modules-extra. But even if not, perhaps you'd like to maintain similarity. If in fact you're not concerned with similarity, you may want to prune a lot of unused stuff out of `/roles/deploy/defaults/main.yml` (which isn't part of 'deploy_helper').

swalkinshaw: Oh yeah good point. I completely forgot about this distinction. I probably will go through the `deploy` role and customize/clean it up a bit more then.

roles/users/templates/sudoers.d.j2
```diff
@@ -0,0 +1,5 @@
+# {{ ansible_managed }}
+
+{% for service in web_sudoers %}
+{{ web_user }} ALL=(ROOT) NOPASSWD: {{ service }}
+{% endfor %}
```
fullyint: The deploy role's "Reload php-fpm" task only works for me if run as `root` instead of `ROOT`.
@mAAdhaTTah This task was hanging for you too. Does this change resolve the issue?
```diff
- {{ web_user }} ALL=(ROOT) NOPASSWD: {{ service }}
+ {{ web_user }} ALL=(root) NOPASSWD: {{ service }}
```

swalkinshaw: Ah didn't run into that since I've been mostly testing deploys lately with `vagrant` user which already has sudoers access. Fixed.

fullyint:
@swalkinshaw Any deploy history log file built in?

fullyint:
I don't have as wide of perspective as you do @swalkinshaw, but the `vars` in `deploy.yml` feel a little out of place to me. If they're variables that won't change all that much, maybe they'd fit better in `deploy/defaults/main.yml`. Or, If they will change, presumably they may change on a site-by-site basis, so maybe they fit in `wordpress_sites` (vs having users `deploy.yml` per deployment).

Also, I'd be tempted to drop a few vars that feel like unnecessary middlemen to other vars: `project`, `project_root`, `project_git_repo`, `project_version`. For example, `{{ project_git_repo }}` is only used once, only as alternative to `{{ wordpress_sites[site].repo }}` (not much advantage).

roles/deploy/tasks/main.yml
```diff
@@ -0,0 +1,80 @@
...
+  shell: mv {{ deploy_helper.new_release_path }}/{{ project_subtree }}/* {{ deploy_helper.new_release_path }}
+  when: project_subtree is defined
+
+- name: Remove unwanted files/folders from new release
+  file: path="{{ deploy_helper.new_release_path }}/{{ item }}" state=absent
+  with_items: project_unwanted_items
+
+- name: Copy project files
+  copy: src="{{ item.src }}" dest="{{ deploy_helper.new_release_path }}/{{ item.dest }}" mode="{{ item.mode | default('0644') }}"
+  with_items: project_files
```
fullyint: #### Optional compile locally
This works for me, right before the "Copy project files" task.
```yaml
- name: Run pre_build_commands_local on Ansible host
  local_action: "command {{ item.cmd }}"
  args:
    chdir: "{{ item.path }}"
  with_items: project_pre_build_commands_local
```
...with the following example definition in defaults or `wordpress_sites`
```yaml
project_pre_build_commands_local:
  - path: ../example.dev/web/app/themes/sage
    cmd: gulp --production
```
#### Switch "Copy project files" to rsync
I'd like to switch "Copy project files" to [synchronize](http://docs.ansible.com/synchronize_module.html) to use its `.rsync-filter`:
```yaml
- name: Copy project files
  synchronize: src="{{ item.src }}" dest="{{ deploy_helper.new_release_path }}/{{ item.dest }}"
  with_items: project.sync_files  | default([])
```
...used with this optional specification in `wordpress_sites`:
```yaml
wordpress_sites:
  example.com:
    sync_files:
      - name: compiled_theme_assets
        src: ../example.dev/web/app/themes/sage/dist
        dest: web/app/themes/sage
``` 
@mAAdhaTTah this could replace the rsync task discussed earlier.

swalkinshaw:
@fullyint I think I might leave the variables in `deploy.yml` as is for now. I don't know why but in my mind it maps more directly to a Capistrano config. Which I realize is probably irrational. We can always clean it up after.

re: log file. Another thing that's needed but won't hold up this PR.

re: compiling assets. Funny that by advocating the use of a VM for development makes it harder to locally compile assets. Tools like Node, npm, gulp, etc should be running on the VM. But you run Ansible from your host machine which won't have access to them. That being said, your solution is very simple anyway.

fullyint:
>Tools like Node, npm, gulp, etc should be running on the VM.

You install and run those tools in a Vagrant VM? Huh, I've only put them on my host machine, so I hadn't considered that a person's Ansible host might not have the tools installed. I see that in `roles/deploy/defaults/main.yml` you gave an example of installing the tools and compiling on the remote server (using `project_pre_build_commands`). I think I'll compile locally though.

swalkinshaw:
@fullyint I was talking with @retlehs and I think we'll just go with the locally compiling by default. I don't want to get into installing Node, npm, Gulp etc. It's also really slow on a VM. Your suggestion looks great. Do you want to do a PR against this branch?

austinpray:
>It's also really slow on a VM.

Care to share your test results aka numberzz aka hard data? :sunglasses:

swalkinshaw:
@austinpray nope. I mostly meant the provisioning process (compiling latest Node from source).

austinpray:
Also: agree with locally compiling. You need to be compiling from production with the same process you are using to dev. It would not make sense to do all of your dev on local and then hope it works inside the VM when deploying. If you are working with a team then you should be using some sort of CI process/compiling on the server.

roles/deploy/tasks/main.yml
fullyint: I asked 
>could you just clone the project straight into `deploy_helper.new_release_path` instead of cloning to one dir then copying to another?

I now see the benefit of idempotence in cloning into `project_source_path` (`example.com/shared/source`). It will skip the cloning task if it already has the specified version. If cloning straight into the new release path as I proposed, it would perform the cloning every time.

retlehs:
:shipit:
:shipit:
:shipit:
:shipit:

server.yml
```diff
@@ -0,0 +1,20 @@
+---
+- name: "WordPress Server: Install LEMP Stack with PHP 5.6 and MariaDB MySQL"
+  hosts: web
+  remote_user: root
```
fullyint: I anticipated you'd change this to `sudo: yes` and `remote_user: admin`, given that `secure-root.yml` blocks ssh access for root.

swalkinshaw: Not until `secure-root.yml` is either explicitly added to our instructions or merged into `server.yml` though.

README.md
```diff
@@ -2,95 +2,86 @@
...
 
-You probably don't need to edit your `Vagrantfile` and should avoid it unless you need more advanced configuration.
+## Deploying to remote servers
 
-Tip: Assuming you leave the default `type: 'nfs'` for `config.vm.synced_folder` (in `Vagrantfile`), see the vagrant docs section ["Root Privilege Requirement"](https://docs.vagrantup.com/v2/synced-folders/nfs.html) for how to avoid having to type your password with every `vagrant up`.
+1. Run `./deploy <environment> <site name>`
+2. To rollback a deploy, run `ansible-playbook -i hosts/<environment> rollback.yml --extra-vars="site=<site name>"`
 
-Note: `hosts/development` is there for completeness sake only as Vagrant automatically generates and uses its own.
+<<<<<<< HEAD
```
fullyint: <<<<<<< HEAD

swalkinshaw: Updated!

swalkinshaw:
Took forever but thanks for all the help. @mAAdhaTTah @fullyint @austinpray @retlehs :+1:

el-rotny:
Good job!

fullyint:
### Major congratulations!!! 🎆 
Thank you for the tons of work you've put in!

minor things
- maybe define/describe `github_ssh_keys` somewhere (did I miss it?)
- could drop the first 17 lines of `roles/deploy/defaults/main.yml`. It includes `project_version`, but you've got that defined in `deploy.yml`.

el-rotny:
@swalkinshaw , what does this mean for the rest for Bedrock w/ Capistrano based deploys? Will that be deprecated at some point. (probably a better forum for Q&A. Apologies in advance).

swalkinshaw:
@el-rotny yes ideally the Capistrano configs will be removed from Bedrock and probably just put into their own separate one.

Docs will be updated in Bedrock as well. We'll also have to document how you can use the `deploy.yml` playbook without using the rest of this project if someone only wants deploys.

mAAdhaTTah:
This is siiiiiick!

austinpray:
Siqqqqq

fullyint:
@mAAdhaTTah I edited my [comment](https://github.com/roots/bedrock-ansible/pull/95#issuecomment-87300154) above (re: bindfs mount) to use one instead of three.

mAAdhaTTah:
@fullyint That's awesome, thanks. Personally, I think the only significant thing missing from Bedrock-Ansible is the DO integration. Then every time we make changes to BA, I can just spin up a new server, no hassle. :) Thanks for all the hard work!

austinpray:
:sunglasses:

mAAdhaTTah:
@fullyint Made a slight mod given the new Vagrantfile:

```ruby
config.vm.synced_folder '../themes/sage', File.join(nfs_path(name), 'web/app/themes/sage'), type: 'nfs'
```

fullyint:
@mAAdhaTTah Nice! I like how easily that integrates with the official project. I was too hasty when I replaced the official slick sync stuff with my hardcoded stuff. Thanks.

PR #113, opened 2014-12-21
Add mysql_user var

swalkinshaw:
cc @emaildano 

Adds new `mysql_user` var instead of hardcoding `root`. This can be vary by group or host obviously.


emaildano:
Getting better all the time. :beers: :

PR #116, opened 2014-12-24
Enable subdomain multisite installations

nathanielks:
Dropped dependency on `vagrant-hostsupdater` in favor of [`landrush`](https://github.com/phinze/landrush). `landrush` spins up a small dns server that allows us to use wildcard subdomains, a requirement for subdomain multisite installs. Let me know what you think!


swalkinshaw:
Finally getting back to this.

I'm thinking the best solution might just to be to keep the necessary subdomain changes but leave the hosts/DNS up to the user?

nathanielks:
mmm that could be a solid option. or have landrush be an optional thing? "suggested default"?

swalkinshaw:
Yeah just add a note about it or they can always just manually add the hosts they need. Odds are it won't be many subdomains.

nathanielks:
:+1:

retlehs:
:eyes:

retlehs:
https://github.com/roots/bedrock-ansible/wiki/Multisite

retlehs:
see #168

Issue #122, opened 2015-01-16
SMTP server request


rstormsf:
With current setup, you have no way to use Contact Form or forgot password link because there is no mail server being setup. 

Actually, there are bunch of SMTP plugins so this could be delegated to some other service which I think is the best option which rewrites PHPMailer function

swalkinshaw:
We haven't installed any SMTP clients because we're in favour is using a 3rd party service like Mandrill, Sendgrid, Mailgun etc.

But I do think we should still provide a simple client by default that's only for sending. msmtp and ssmtp both seem simple so I'll leave this issue open until I check them out.

rstormsf:
Do you mind to explain briefly how do you integrate it? Do you using any plugin? How do you reset password from wp-admin?

swalkinshaw:
All of those services are very similar and usually work in 2 ways: either via their API (raw HTTP calls or a PHP client library), or via SMTP (which you'd need a client like you mentioned).

To hook into WP mail there's various plugins. Here's an example of one: https://wordpress.org/plugins/wpmandrill/

This one uses curl in the background for their API so not SMTP needed at all.

In general you should always use an email service for sending external email. It's very hard and not really worth the trouble to setup your own. If it's just for "internal" emails like signups/resets for developers/admin then you don't need to worry.

rstormsf:
but it doesn't solve the problem for reset password link

swalkinshaw:
What doesn't? The Mandrill plugin would work for example.

But as I said I'm leaving this issue open and we'll add an SMPT client which will fix this problem as well. Then emails will work regardless of what solution people use.

JulienMelissas:
I love the Mandrill plugin and service, but :+1: for for SMPT client just in case.

nathanielks:
Can I vote against it? SMTP would feel more like an awkward dongle that will probably fall behind in maintenance. Reason I say this is it appears everyone uses an external service, so why bake it in?

austinpray:
@nathanielks SMTP should be opt-out and not opt-in. We should definitely recommend people use an external service but email not working out of the box by default is unacceptable.

mAAdhaTTah:
Yeah, you need something. I think there was one instance where I added a user, not realizing there was no mail system installed, and the adding process timed out (it worked, the user was added, but it was a very odd experience).

PR #123, opened 2015-01-17
Add SMTP role

swalkinshaw:
Ref #122

This uses ssmtp as a simple mail delivery tool for forwarding mail to an external SMTP server.

/cc @austinpray @nathanielks @rstormsf


austinpray:
Gonna try this out on one of my servers to test

swalkinshaw:
Closing this. Realized this solution isn't enough since email should work out of the box on development too without requiring a 3rd party SMTP service.

swalkinshaw:
and we're back! Going to stick with as anyone can use their Gmail (for example) in development.

austinpray:
:sunglasses:

rstormsf:
:+1:

PR #136, opened 2015-03-03
HHVM implementation

swalkinshaw:
This is an updated and slightly simplified version of #45.

Most of the credit goes to @nathanielks since a lot of this was copied directly from that work.

This does *not* support magically toggling between the two implementations though. I don't think the extra complexity that adds is worth it. If anyone wants to switch between the two on an existing server, it's not much extra work to do the manual cleanup.

Note: this also removes the per-site php-fpm worker pools that existed before. It was needless complexity for the little/no benefits it added and would have made the setup diverse more from HHVM.

Todo:

- [x] Add nightly cron to restart HHVM due to [memory leaks](https://github.com/facebook/hhvm/issues/4398)


louim:
Reading trough HHVM issues, I came across: facebook/hhvm#4398. I'm not currently using HHVM, but I'm looking into it for performance reason. Do you think such mechanism (reloading HHVM automatically when memory usage is high or juste in the middle of the night) should be added? Maybe just a comment in the readme or an option to enable it?

Thoughts?

swalkinshaw:
@louim good catch. Tough to tell how pervasive this problem is but I'm more inclined to include a nightly cron job by default since the memory leak source has been [found](https://github.com/facebook/hhvm/issues/4250). And obviously this won't change anytime soon since WP won't switch to anonymous functions.

louim:
:+1: for the nightly cronjob.

swalkinshaw:
@louim added the daily restart cron job

louim:
@swalkinshaw Excellent, do you plan to merge this into the project soon?

swalkinshaw:
@louim yep, I'll try and do it tonight.

PR #138, opened 2015-03-06
php5-xdebug role

JacobDorman:
Added php5-xdebug role from https://galaxy.ansible.com/list#/roles/1133 (as mentioned this [this thread](https://discourse.roots.io/t/found-an-xdebug-ansible-role-installer-wondering-how-it-might-work-with-bedrock/2901))

Change ```xdebug_install: false``` in group_vars/development to enable.

Fixes https://github.com/roots/bedrock-ansible/issues/42

### DIFF
--- a/group_vars/development
+++ b/group_vars/development
@@ -29,3 +29,12 @@ php_display_startup_errors: 'On'
 php_track_errors: 'On'
 php_mysqlnd_collect_memory_statistics: 'On'
 php_opcache_enable: 0
+
+xdebug_install: false
+php_xdebug_remote_enable: true
+php_xdebug_remote_connect_back: true
+php_xdebug_remote_host: localhost
+php_xdebug_remote_port: 9000
+php_xdebug_remote_log: /tmp/xdebug.log
+php_xdebug_idekey: XDEBUG
+php_max_nesting_level: 200
--- a/roles/php5-xdebug/.travis.yml
+++ b/roles/php5-xdebug/.travis.yml
@@ -1,0 +1,21 @@
+language: php
+
+php:
+  - 5.5
+
+before_install:
+  - sudo apt-get update  -y -q
+
+install:
+  - sudo apt-get install python-pip -y
+  - sudo pip install ansible==1.4.0
+
+script:
+  - cd tests
+  - ansible-playbook -i inventory playbook.yml --syntax-check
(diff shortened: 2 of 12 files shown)


swalkinshaw:
Is there anything changed from the galaxy role? Or is it a straight up copy?

If so we should just switch to requiring installing galaxy roles as part of the install process.

JacobDorman:
It's a straight-up copy. The other roles aren't installed from galaxy so I didn't want to introduce a new requirement / step.

Are you thinking requirements.yml, or just adding a "To use xdebug, ```ansible-galaxy install MaximeThoonsen.php5-xdebug``` "?

Are any of the other roles able be be installed from or replaced by ansible galaxy roles?

swalkinshaw:
I'm thinking of adding a `requirements.yml` or `galaxy.yml` or whatever (is there a consensus name?).

There's a couple candidates:

* this one
* ntp
* mariadb
* memcached
* ssmtp

fullyint:
Maybe logrotate as another candidate galaxy role (#141)

swalkinshaw:
@fullyint I was just looking at that. Problem is our logfiles for WP sites are dynamically defined :(

JacobDorman:
Re-using roles from galaxy wherever possible is definitely a good thing. Do you want to merge this in the meantime, and create a new issue for switching to a ```requirements.yml``` ? or should we introduce it along with this role?

Not sure if there is a consensus on naming, i'm just going off http://docs.ansible.com/galaxy.html#advanced-control-over-role-requirements-files.

It looks like you can specify version numbers in the yml, but not version constraints like npm/composer/bundler. Should be enough. Anything else we need to know?

swalkinshaw:
@JacobDorman I'd rather just switch to using Galaxy roles now. The first one can be for this role and including the version number would be good :+1: 

We'll need `README` changes as well to update requirements/install process and a note on Galaxy roles.

JacobDorman:
okay, sounds good. I'll have a crack at it if this is still open in a couple of weeks.

swalkinshaw:
@JacobDorman we're now requiring installing galaxy roles as part of installation so this PR can be redone safely now. See https://github.com/roots/bedrock-ansible/blob/dd74bc493f812e1fe017c86997164482b7009b0e/requirements.yml.

PR #151, opened 2015-03-31
SSL support

enricodeleo:
This fork introduce a basic SSL support. 

The certificates are required and need to be on the host machine (I use the same 2 auto-signed certificates for all my development virtual machines). Yo can specify the local path and certificates' names in group_vars/all

### DIFF
--- a/roles/mariadb/defaults/main.yml
+++ b/roles/mariadb/defaults/main.yml
@@ -1,3 +1,3 @@
 keyserver_fingerprint: "0xcbcb082a1bb943db"
-mirror: nyc2.mirrors.digitalocean.com
+mirror: "ftp.nluug.nl/db"
 version: "10.0"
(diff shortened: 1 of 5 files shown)


roles/mariadb/defaults/main.yml
```diff
@@ -1,3 +1,3 @@
 keyserver_fingerprint: "0xcbcb082a1bb943db"
-mirror: nyc2.mirrors.digitalocean.com
+mirror: "ftp.nluug.nl/db"
```
mAAdhaTTah: Was this intended to be committed?

enricodeleo: I had to change the mirror this morning due to an error, but I think this could be left as it was, it was probably a temporary downtime of the DO's mirror.

austinpray: They had some hickups: https://status.digitalocean.com/

Def change it back though.

swalkinshaw:
@enricodeleo thanks so much for getting this started again. I wanted to this to handle this myself though since there's a lot more Nginx configuration needed.

See https://github.com/roots/bedrock-ansible/pull/172

enricodeleo:
@swalkinshaw thanks to your attention and your great work! I'll try to contribute further in the near future since as I said before I am an heavy user of Bedrock :)

PR #153, opened 2015-03-31
Improvements

enricodeleo:
Working on Deploys branch that seems more lean and will make easier to deploy siege.

* Ported the SSL support thing I've pull-requested this morning for the master branch
* Added a swap role from ansible galaxy 
* Since I had "bad gateway" issues both with php and hhvm with a fresh copy of the repo I switched both from unix socket to tcp (such a thing could be used for a _fallback-to-php-if-hhvm-fails_ strategy in a near future)
* HHVM now is executed with the web_user user so you won't have directory permissions anymore


swalkinshaw:
@enricodeleo thanks for all this work. Most of this is needed and is good but I'm going to hold on these changes until the `deploys` branch is merged (hopefully very soon).

enricodeleo:
@swalkinshaw my pleasure, I'm a big fan of Bedrock Ansible, I thought I should help in some way :)

Something I'd find useful would be to optionally reintroduce Capistrano for deployment. I loved the ansible-way idea at the beginning but now I find that in some cases Capistrano would be a more reliable solution and it has definitively more features. Dunno, the current solution seems too bash-ish.

swalkinshaw:
@enricodeleo I kind of ruined this PR with a lot of updates and rebases in #95 :(

Do you mind submitting a new PR with just the changes to HHVM and php-fpm? I already commented on the swap PR and we'll handle the SSL one separately as well.

PR #154, opened 2015-04-02
Add rollback role

fullyint:
No pressure to use this. I just thought I'd try a draft. Might be nice with #95 

**Roll back.** This is for rolling back to the previous deploy/release (default), or to any existing deploy you specify. It is **not** for when a deploy fails because the deploy role deploys nothing if it fails at any point.

**Use.** We could write a bash wrapper script, but for now...
- **Roll back to previous:** `ansible-playbook -i hosts/staging rollback.yml --extra-vars="site=staging.example.com"`
- **Roll back/forward to existing release:** `ansible-playbook -i hosts/staging rollback.yml --extra-vars="site=staging.example.com release=20150401173849"` If a site runs on multiple servers, I suspect release names (time stamps) would vary slightly across servers. If so, the deploy role could maybe put a single time stamp into deploy_helper's `release` parameter at initiation.

**Assumptions.** The janky "Get list position of current symlinked release" task assumes the `releases` directory has no contents but releases, in chronological list order. In the future, the task could grab release names/sequence from the deploy history log.

### DIFF
--- a/roles/rollback/tasks/main.yml
+++ b/roles/rollback/tasks/main.yml
@@ -1,0 +1,18 @@
+---
+- include: user-release.yml
+  when: release is defined
+
+- include: prior-release.yml
+  when: release is not defined
+
+- name: Check whether target release was from a successful deploy
+  stat: path="{{ new_release_path }}/DEPLOY_UNFINISHED"
+  register: target
+
+- name: Fail if target release was from failed deploy
+  fail: msg="Cannot switch to release at {{ new_release_path }}. It is from an unfinished deploy.
+        You may manually specify a different release using --extra-vars='release=12345678901234'."
+  when: target.stat.exists | default(False)
+
+- name: Link 'current' directory to target release
+  file: path="{{ project_root }}/current" src="{{ new_release_path }}" state=link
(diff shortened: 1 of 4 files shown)


fullyint:
I started wondering if this role isn't needed because the symlink can be changed in a single command:
`sudo ln -sfn releases/20150401173849 current`

But there are still a few advantages to the role:
- doesn't require sudo
- reduces chance errors by preventing `current` from being symlinked to
   - non-existent release (e.g., due to error while typing release name)
   - a faulty release (e.g., a release that still has the DEPLOY_UNFINISHED file)
- doesn't require entering a release name (when using default of rolling back to previous)
- will write the rollback to deploy history log (once it exists)

mAAdhaTTah:
+1 for including, imo. Both capistrano and rocketeer (a php-based deployment solution I've also used with Bedrock) have rollback commands, which would be a lot easier than trying to go in and symlink things yourself.

austinpray:
Include. If I'm rolling back something has gone terribly wrong and I need a fast command.

retlehs:
nice man!

swalkinshaw:
:+1:

PR #155, opened 2015-04-02
Add option for local build commands

fullyint:
Adds option for local build commands like compiling theme assets. For #95

**Does not run local build commands by default.** I felt like running build commands on the user's local assets should be opt-in, so I couldn't get myself to define `project_pre_build_commands_local` by default. But it seems likely people will want it, so for convenience, I put an example (commented-out) definition in `deploy.yml`. I can remove that if you want, because I also put an example definition in comments in `roles/deploy/defaults/main.yml`. Or, maybe it should be defined in `wordpress_sites`, if theme paths will differ between sites.

**Does copy up dist directory by default.** Adds a definition of `project_files` to `deploy.yml`, making it the default to copy up the theme's`dist` directory. Maybe you'd prefer this definition of `project_files` to be 
- commented out, but still in `deploy.yml`, or
- just an example in comments in `roles/deploy/defaults/main.yml`, or
- defined in `wordpress_sites` if theme paths will differ between sites.

**Switched from copy to synchronize module.** I changed the "Copy project files" task to use the [synchronize](http://docs.ansible.com/synchronize_module.html) module which allows use of a `.rsync-filter` file (like in this [comment](https://github.com/roots/bedrock-ansible/pull/95#issuecomment-86280712) - item 3). Although users wouldn't likely need/use the `.rsync-filter` when copying up a theme's compiled `dist` directory, it could come in handy for other files. However, the synchronize module requires rsync be installed on local and remote. Foresee that being a problem?

### DIFF
--- a/roles/deploy/defaults/main.yml
+++ b/roles/deploy/defaults/main.yml
@@ -37,9 +37,9 @@ project_composer_vendor_path: vendor
 #   - name: "some_file"           // <- optional, for your own reference and readability
 #     src: "local-path-to-file"   // <- relative or absolute, just like Ansible
 #     dest: "remote-path-to-file"
-#   - name: "some_other_file"
-#     src: "local-path-to-other-file"
-#     dest: "remote-path-to-other-file"
+#   - name: compiled theme assets
+#     src: ../example.dev/web/app/themes/sage/dist
+#     dest: web/app/themes/sage
 project_files: []
 
 # All the templates to copy to the remote system on deploy. These could contain config files.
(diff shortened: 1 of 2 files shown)


swalkinshaw:
Let's remove the commented out variables in `deploy.yml` for now. It exists in defaults and will be in the `README` as well.

About synchronize, for the existing copy task, rsync is only needed on the remote server I assume? If so, that's fine since it's already installed on the VM.

deploy.yml
```diff
@@ -8,6 +8,13 @@
     project_root: "{{ www_root }}/{{ site }}"
     project_git_repo: "{{ project.repo }}"
     project_version: "{{ branch | default('master') }}"
+    project_files:
+      - name: compiled theme assets
+        src: ../example.dev/web/app/themes/sage/dist
+        dest: web/app/themes/sage
```
swalkinshaw: Think I'd rather this just be commented out in `defaults` for now as well.

fullyint:
As requested, I removed the vars from `deploy.yml`. They appear commented out in `defaults`.  

Ansible re: [synchronize](http://docs.ansible.com/synchronize_module.html):
>rsync must be installed on both the local and remote machine.

swalkinshaw:
But it shouldn't be needed locally for a completely remote operation? I'm hoping :(

fullyint:
I understood that message to mean that because synchronize is rsyncing files from your local machine to remote, the local machine in fact needs rsync in order to pull it off.

I'm not certain, but I think all *nix machines will have rsync ([ref](http://en.wikipedia.org/wiki/Rsync#Uses)). I think our Windows users will be running their Ansible from a *nix machine somewhere (probably in their Vagrant VM), so they should be fine too.

I can switch it back to the copy module if you want. Maybe that's safest. The project doesn't need to support my non-standard usage plans with the beloved `.rsync-filter`, or my being suckered by comments like these:
>Not sure why copy fails, but the solution was to use synchronize instead. ([ref](http://stackoverflow.com/a/23813630))
 
>The “copy” module recursively copy facility does not scale to lots (>hundreds) of files. For alternative, see synchronize module, which is a wrapper around rsync. ([ref](http://docs.ansible.com/copy_module.html))
 
>It is really amazing how bad copy is compared to synchronize. For bulk copies the difference is orders of magnitude. ([ref](http://stackoverflow.com/questions/23795117/why-doesnt-the-ansible-copy-module-copy-this-dir#comment44880610_23795117))

swalkinshaw:
No, I think it's fine. I realized the same thing about Windows users so it's not as big a deal as I originally thought.

swalkinshaw:
:+1:

Issue #134, opened 2015-03-01
Add NTP package


retlehs:

swalkinshaw:
See https://github.com/roots/bedrock-ansible/pull/159

PR #159, opened 2015-04-03
Add ntpd role

swalkinshaw:
Ref: #134


jeroenvisser101:
@swalkinshaw this seems to break things, but I might be doing something wrong. I get the following editor:

```
ERROR: cannot find role in ~/ansible/roles/ntp or ~/ansible/ntp or ~/ansible/vendor/roles/ntp
```

(I ran `vagrant up`, and it showed me this right after it launched Ansible)

PR #160, opened 2015-04-06
Fix changing of root password in mariadb task

teohhanhui:
Host was left out resulting in some root logins with no password.


swalkinshaw:
Wow that `with_items` was kind of useless without this. Thanks :+1:

PR #163, opened 2015-04-07
Added missing command of install for ansible-galaxy in README.md instruc...

brettwise:
Command wasn't working for me when copy pasting over. Turns out it was missing the install command.


retlehs:
thanks!

PR #164, opened 2015-04-07
Add deploy logs and improve rollbacks

fullyint:
Improves on #154. Writes log entries for deploys and rollbacks done with bedrock-ansible. For rollback, it checks available releases on server and parses their most recent entries in `logs/revisions.log`, aborting/warning if there may be trouble. I've tested on DigitalOcean, but not development on Vagrant.

This may not fit your plans. I didn't realize how complicated it would get. A next iteration could extract most of `rollback/defaults/main.yml` into a module for integrated logs and rollbacks.

## Example Log
```
                                                   -----------  NEW  --------------      ----------  PRIOR  -------------
  DATE       TIME      USER       EVENT    STATUS    RELEASE        SHA     VERSION        RELEASE        SHA     VERSION
----------  --------  ----------  ------   ------  --------------  -------  -------      --------------  -------  -------
2015-04-07  04:33:48  fullyint    deploy    ok     20150407043348  4127081  master    <- ??????????????  ???????  ???????
2015-04-07  05:31:45  fullyint    deploy    ok     20150407053145  4127081  master    <- 20150407043348  4127081  master 
2015-04-07  05:33:40  otheruser   deploy    fail   20150407053340  4127081  master  X <- 20150407053145  4127081  master 
2015-04-07  05:34:48  otheruser   rollback  ok     20150407053145  4127081  master    <- 20150407053340  4127081  master 
2015-04-07  05:38:21  fullyint    deploy    ok     20150407053821  e3167f1  1.3.3     <- 20150407053145  4127081  master 
2015-04-07  05:39:06  fullyint    deploy    ok     20150407053906  deb1dbd  1.3.4.1   <- 20150407053821  e3167f1  1.3.3  
2015-04-07  05:39:57  another     deploy    fail   20150407053956  4127081  master  X <- 20150407053906  deb1dbd  1.3.4.1
2015-04-07  05:41:05  yetanother  rollback  ok     20150407053906  deb1dbd  1.3.4.1   <- 20150407053956  4127081  master 
2015-04-07  05:42:33  fullyint    rollback  force  20150407053956  4127081  master    <- 20150407053906  deb1dbd  1.3.4.1
2015-04-07  05:43:23  fullyint    deploy    ok     20150407054323  4127081  master    <- 20150407053956  4127081  master 
```

## Basic Usage
**Deploy.** 
- Basic: `./deploy.sh <environment> <site name>`
- Specific version from repo (default is 'master'): `   ansible-playbook -i hosts/<environment> deploy.yml --extra-vars "site=<site name> branch=<branch>"` where branch is branch name, sha, or tag for your git repo. (This was already built in to `deploy.yml`. Just pointing out that the log will pick up on it.)

**Rollback.** `ansible-playbook -i hosts/<environment> rollback.yml --extra-vars "site=<site name>"`

## --extra-vars for rollback.yml
_**site=\<site name\>**_ (required, no default)

_**version=\<RELEASE\>**_ (default is auto-rollback; also accepts SHA and VERSION). Choose a release from the columns labeled "NEW". Can roll forward to existing release.

_**version=list**_ (default is auto-rollback) Doesn't rollback. Just lists most recent log entry for each release.

_**priorby=\<activation | creation\>**_ (default is 'activation') Determine whether auto-rollback selects the prior version according to recency of activation or creation. Imagine you create release A then B then C, then rollback to B, then create D, and now want to auto-rollback.
 - summary: A -> B -> C -> B -> D -> ?
 - which prior release should be used for auto-rollback?
   - B: prior in terms of activation (sort by DATE & TIME in log), obtained via `priorby=activation`
   - C: prior in terms of creation (sort by RELEASE in log) obtained via `priorby=creation`

_**markfail=true**_ (default is 'false') Set the status of the currently active release to 'fail' in the deploy log before rollback. This is useful if the current release had a problem and you don't want it to be a candidate for future auto-rollback. This isn't the only way a release gets marked with 'fail' status. Each deploy/rollback starts by writing an entry to the log with status 'fail', which is overwritten to 'ok' at the end (unless it fails).

_**force=true**_ (default false) Override all the error-checking and just create the damn symlink.

## Error Checking
Diagnoses and gives user specific feedback.
- When `revisions.log` appears corrupted in formatting (which could lead to failed rollbacks).
- When no prior version is available for auto-rollback. Can also warn if the current version is somehow not in the log, making it impossible to know which version would be prior.
- When user specifies a version but...
  - version doesn't exist
  - version is non-unique (e.g., version=master would probably match multiple)
  - version exists but isn't in log (i.e., it isn't 'trusted')
  - version's status is not 'ok' (e.g., was marked 'force' or 'fail' in the log)
  - version is already the currently active version
- Output from `version=list` will notify of directories in `example.com/releases` that are not in log.

## Time Stamps
I switched to naming releases using a UTC timestamp that should result in releases having same name across multiple servers ([credit](https://groups.google.com/d/msg/ansible-project/lPmj6Y5XxUg/tzgcKPvORYgJ)). This should make things easier for rollbacks. Log entries and release names would also then be strictly chronological even if people deploy from multiple timezones. But TBH, I haven't tried any of this across multiple servers.


fullyint:
I think this is on the right track in terms of features, but not in terms of implementation. The scripting type issues ought to be handled in a module, not in a maze of jinja filters in `defaults/main.yml`. I started with jinja, not anticipating how it all would keep growing and growing. So, I could learn a little python, and take another stab at it, doing it right, but I'm guessing you already have your own plans for implementing these features.

swalkinshaw:
@fullyint not really :)

I've left this here for a bit because it is complicated. It's just a lot of code to go through. But yeah I agree that a library/module would be better suited for a lot of this.

Let me know what you want to do with this in the short term. Is there anything smaller to break out of it first?

fullyint:
As it's your project/repo, I respect anything you decide. Part of me thinks the features and logic are super smart, but part of me fears that it could be just a beginner's silly contraption in code. I simply don't have broad enough experience and perspective to know.

I'll have to rely on your judgment as to whether this has merit, or any place in your project. My view is that it works, and is really convenient, and I like the logs. I did enough testing to share it for others to respond to, but I didn't hammer on it hard, so I expect we'd find bugs.

> It's just a lot of code to go through.

That's the fault of my bad implementation. Had I put all the logic in a python-based library/module, it'd be easier (of course). I bet that it'd be easiest to evaluate while test driving, trying to stump it, then looking at the code to see its strategy.

>Is there anything smaller to break out of it first?
 
It could merged as is and require zero change in how people use bedrock-ansible. It uses the same commands for deploy and rollback as at present. It just has the extra features available in the `--extra-vars` and in the error-checking. I'm not sure why break out a smaller piece of it, unless you see pieces in my description above that you know you'd never want to include in the final version anyway.

A big argument for not including ANY of it would be if you expect that a next iteration/implementation would be totally different. In that case, maybe there's no reason to adding something you'll just turn around and rip out. Another reason not to include it might be to avoid having its weird jinja implementation in your repo history, kind of like, "don't ship code that is doing it wrong."

Some notes I have to myself of what I would add/change, if it's worth pursuing:
- add a dry-run option
- rename `logs/revisions.log` to `logs/{{site}}.revisions.log` (like the access and error logs)
- use multiline variables for the messages, like I did in the [`logrotate_scripts`](https://github.com/roots/bedrock-ansible/blob/3f5c44d097ef0c71d81bee705b37042e65178aba/group_vars/all#L48-L51) variable
- maybe write a bash wrapper script

fullyint:
Closing. A PR is not the place for "hey, what do you think of my poorly-implemented idea?" Wish I had time right now to re-do it with a library/module, but it'll have to wait. Anyone is welcome to beat me to it.

Issue #174, opened 2015-04-09
Set php fpm user to web_user


fullyint:
#95 changed the php-fpm `listen.owner` and `listen.group` ([lines](https://github.com/roots/bedrock-ansible/commit/373fe77e90e5208fc96a9def301bb9e17b66f2ac#diff-97894dad9044d33182d3b1ad613c6e20L3)) but I wonder if it should have changed the `user` and `group` instead. I think it caused the issues with permissions on the uploads folder [here](https://discourse.roots.io/t/unable-to-create-directory-uploads-2015-03-is-its-parent-directory-writable-by-the-server/3304) and [here](https://discourse.roots.io/t/cant-upload-images/3450) and likely in #150 as well.

Here's my best guess at what's going on. When someone tries to upload a file, I'm guessing php-fpm acts on their behalf to write a new file to the uploads directory. The permissions on the uploads directory are 0755 so only the owner can write, and the owner is probably `web` (the default `web_user`). However, php-fpm is acting as `www-data` (`user` in `roles/php/templates/php-fpm.conf.j2`), so it is denied permission and can't write.

So, when stefanlindberg ([here](https://discourse.roots.io/t/unable-to-create-directory-uploads-2015-03-is-its-parent-directory-writable-by-the-server/3304)) added `user = {{ web_user }}` it resolved the problem. I propose a slightly different revision, which works in my tests (`roles/php/templates/php-fpm.conf.j2`): 
```diff
- listen.owner = {{ web_user }}
- listen.group = {{ web_group }}
+ listen.owner = www-data
+ listen.group = www-data
- user = www-data
- group = www-data
+ user = {{ web_user }}
+ group = {{ web_group }}
```

```
                          ------- user --------
                              web    www-data 
                          ---------------------
listen.owner  |      web |   Works     fails  
              | www-data |   Works     fails 
```
(got the results above whether `nginx_user` is `web` or `www-data` in `roles/nginx/templates/nginx.conf.j2`).

swalkinshaw:
Yeah this makes sense. Want to do a PR?

fullyint:
I've invited [stefanlindberg](https://discourse.roots.io/t/unable-to-create-directory-uploads-2015-03-is-its-parent-directory-writable-by-the-server/3304/6) to submit the PR after having tracked down the issue. I'll submit it after a while if no response. If it starts to feel urgent, let me know.

swalkinshaw:
[...]

PR #175, opened 2015-04-09
Changed permissions for PHP issue #174

slackday:
- Changed listen.group and listen.owner to www-data in roles/php/templates/php-fpm.conf.j2

- Changed user to {{ web_user }} and group to {{ web_group }} to solve issue with php not being able to create files in Wordpress (related to upload issue)

https://github.com/roots/bedrock-ansible/issues/174


swalkinshaw:
Thanks!

PR #176, opened 2015-04-11
xdebug role from ansible-galaxy

JacobDorman:
Load from ansible-galaxy with requirements.yml instead of including role.

replace #138


swalkinshaw:
:+1:

PR #177, opened 2015-04-11
Set ansible roles_path relative to ansible directory in Vagrantfile

JacobDorman:
Ansible can't find the galaxy roles in vendor/roles after moving the Vagrantfile to a project directory per [wiki instructinons](https://github.com/roots/bedrock-ansible/wiki/Vagrantfile).

ansible.cfg isn't loaded and roles_path would be relative to current directory anyway.

So instead, set roles_path in Vagrantfile relative to ANSIBLE_PATH with ANSIBLE_ROLES_PATH environment variable.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -5,6 +5,9 @@ require 'yaml'
 
 ANSIBLE_PATH = '.' # path targeting Ansible directory (relative to Vagrantfile)
 
+# Set Ansible roles_path relative to Ansible directory
+ENV['ANSIBLE_ROLES_PATH'] = File.join(ANSIBLE_PATH, 'vendor', 'roles')
+
 config_file = File.join(ANSIBLE_PATH, 'group_vars/development')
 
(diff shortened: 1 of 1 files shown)


Vagrantfile
```diff
@@ -14,6 +14,9 @@ else
   raise "#{config_file} file not found. Please set `ANSIBLE_PATH` in Vagrantfile"
 end
 
+# Set Ansible roles_path relative to Ansible directory
+ENV['ANSIBLE_ROLES_PATH'] = File.join(ANSIBLE_PATH, 'vendor/roles')
```
swalkinshaw: Could you move this up under `ANSIBLE_PATH`? Rather just have them grouped.

Also: `File.join(ANSIBLE_PATH, 'vendor', 'roles')`

JacobDorman:
Ahh, sorry about that. Haven't done much Ruby.

swalkinshaw:
Thanks!

PR #180, opened 2015-04-12
Deploy Improvements

swalkinshaw:
* Use `local_path` for easier `pre_build_commands_local` paths
* Create `project_copy_folders` as a generic method of copying folders
  between releases
* Move copying Composer's `vendor` dir into `project_copy_folders`
* Rename `project_files` to `project_local_files`
* Remove Composer specific commands
* Add Composer install to `project_post_build_commands`
* Auto set project_environment with `WP_ENV`
* Add `project_post_finalize_commands`

/cc @fullyint @retlehs

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -16,7 +16,7 @@ Available environments:
 `( IFS=$'\n'; echo "${ENVIRONMENTS[*]}" )`
 
 Examples:
-  deploy staging example.dev
+  deploy staging example.com
   deploy production example.com
 "
 }
--- a/group_vars/development
+++ b/group_vars/development
@@ -3,10 +3,11 @@ mysql_root_password: devpw
 web_user: vagrant
 
 wordpress_sites:
-  example.dev:
+  example.com:
     site_hosts:
       - example.dev
-    local_path: '../example.dev' # path targeting local project directory (relative to root/Vagrantfile)
+    local_path: '../example.com' # path targeting local Bedrock project directory (relative to Ansible root)
+    repo: git@github.com:roots/bedrock.git
     site_install: true
     site_title: Example Site
     admin_user: admin
@@ -23,7 +24,6 @@ wordpress_sites:
       db_name: example_dev
       db_user: example_dbuser
       db_password: example_dbpassword
-      domain_current_site: example.dev # required for multisite
 
 php_error_reporting: 'E_ALL'
 php_display_errors: 'On'
--- a/group_vars/production
+++ b/group_vars/production
@@ -5,6 +5,7 @@ wordpress_sites:
     site_hosts:
       - example.com
       - 192.168.50.5
+    local_path: '../example.com' # path targeting local Bedrock project directory (relative to Ansible root)
     repo: git@github.com:roots/bedrock.git
     system_cron: true
     multisite:
@@ -17,7 +18,6 @@ wordpress_sites:
       db_name: example_prod
       db_user: example_dbuser
       db_password: example_dbpassword
-      domain_current_site: example.dev # required for multisite
       auth_key: "generateme"
       auth_salt: "generateme"
       logged_in_key: "generateme"
--- a/roles/deploy/defaults/main.yml
+++ b/roles/deploy/defaults/main.yml
@@ -1,12 +1,8 @@
-# These variables must be set
-# project_root: "path_to_project_on_the_target_machine"
-# project_deploy_strategy: "git" or "synchronize"
-
 # If you use the "git" strategy:
 # - you must set a repository (no default)
-# project_git_repo: "git_repository"
+project_git_repo: "{{ project.repo }}"
 # - you can set the git ref to deploy (can be a branch, tag or commit hash)
-project_version: "master"
+project_version: "{{ branch | default('master') }}"
 
 # If you use the "rsync" strategy:
 # - you can set a timeout for the synchonize module
@@ -20,49 +16,50 @@ project_deploy_synchronize_delete: false
 project_source_path: "{{ project_root }}/shared/source"
 
 # Files or folders to remove from the source when deploying
-project_unwanted_items: [ '.git' ]
+project_unwanted_items: ['.git']
 
-project_has_composer: true
-
-# Default values to run composer install
-project_composer_binary: composer
-project_command_for_composer_install: "{{ project_composer_binary }} install --no-ansi --no-dev --no-interaction --no-progress --optimize-autoloader --no-scripts"
-# When set to true the old vendor dir is copied to the new release to speed up composer install
-project_copy_previous_composer_vendors: false
-project_composer_vendor_path: vendor
+# There are certain folders you'll want to copy from release to release to speed up deploys.
+# Examples: Composer's `vendor` folder, npm's `node_modules`.
+# These should not be part of project_shared_children since dependencies need to be atomic and tied to a deploy.
+project_copy_folders:
+  - vendor
 
 # All the files to copy to the remote system on deploy. These could contain config files.
 # Example:
-# project_files:
+# project_local_files:
 #   - name: "some_file"           // <- optional, for your own reference and readability
 #     src: "local-path-to-file"   // <- relative or absolute, just like Ansible
 #     dest: "remote-path-to-file"
 #   - name: compiled theme assets
-#     src: ../example.dev/web/app/themes/sage/dist
+#     src: "{{ project.local_path }}/web/app/themes/sage/dist"
 #     dest: web/app/themes/sage
-project_files: []
+project_local_files: []
 
-# All the templates to copy to the remote system on deploy. These could contain config files.
-# Works the same as the project_files
-project_templates: []
+# All the templates to process on the remote system on deploy. These could contain config files.
+# `src` and `dest` paths work the same as project_local_files.
+project_templates:
+  - name: .env config
+    src: roles/deploy/templates/env.j2
+    dest: .env
 
 # The shared_children is a list of all files/folders in your project that need to be linked to a path in "/shared".
-# For example a sessions directory or an uploads folder.. They are created if they don't exist, with the type
-# specified in the "type" key (file or directory).
+# For example a sessions directory or an uploads folder. They are created if they don't exist, with the type
+# specified in the `type` key (file or directory).
 # Example:
 # project_shared_children:
 #   - path: "/app/sessions"
 #     src: "sessions"
 #     type: "file" / "directory"  // <- optional, defaults to "directory"
-#   - path: "/web/uploads"
-#     src: "uploads"
-project_shared_children: []
+project_shared_children:
+  - path: /web/app/uploads
+    src: uploads
 
 # The project_environment is a list of environment variables added to the various *_commands
 # Example:
 # project_environment:
 #   WP_ENV: "production"
-project_environment: {}
+project_environment:
+  WP_ENV: "{{ project.env.wp_env }}"
 
 # There are a few moments in this role where arbitrary command(s) can be run. These commands receive
 # the "project_environment" so deploys for different stages can be done by changing this.
@@ -71,7 +68,7 @@ project_environment: {}
 # `new_release_path` on the remote server (for `project_pre_build_commands` and `project_post_build_commands`).
 # Examples:
 # project_pre_build_commands_local:
-#   - path: ../example.dev/web/app/themes/sage
+#   - path: "{{ project.local_path }}/web/app/themes/sage"
 #     cmd: gulp --production
 # project_post_build_commands:
 #   - path: web/app/themes/mytheme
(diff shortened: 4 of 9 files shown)


roles/deploy/defaults/main.yml
```diff
@@ -79,7 +77,13 @@ project_environment: {}
 #   - cmd: wp db import posts.sql
 project_pre_build_commands_local: []
```
fullyint: There's another `- path: ../example.dev/...` just above in the comments, to be replaced with the new `- path: {{ project.local_path }}/...`

fullyint:
smart stuff, man

fullyint:
I like how you folded the composer install into `project_post_build_commands`. I'd keep it there.

Note, however, that there is an [ansible composer module](http://docs.ansible.com/composer_module.html) which you'll be forgoing, and ansible galaxy advises "How To Review Roles You’ve Tested -> [Code Quality](https://galaxy.ansible.com/intro):" 
> Does it use relevant ansible modules as opposed to relying on the ‘shell’, ‘command’, or ‘script’ tasks?

You can decide whether that's a concern. I like it  the way you have it now.

group_vars/production
```diff
@@ -5,6 +5,7 @@ wordpress_sites:
     site_hosts:
       - example.com
       - 192.168.50.5
+    local_path: '../example.dev' # path targeting local project directory (relative to root/Vagrantfile)
```
fullyint: The README [installation section](https://github.com/roots/bedrock-ansible/blob/3f5c44d097ef0c71d81bee705b37042e65178aba/README.md#installation) implies naming your Bedrock directory `example.com`, so maybe this could avoid potential confusion if it were`.com` instead of `.dev`. Same for other `group_vars` files, especially `group_vars/development` because it is the one that new users will encounter first.

swalkinshaw:
@fullyint good point about the path names. Updated them all to `example.com`.

I'm going to leave the Composer command as is simply because it lets us keep it under `project_post_build_commands`.

swalkinshaw:
@fullyint took your advise and finally moved more variables into defaults.

Also added a little message at the end of deploys:

```
TASK: [deploy | debug ] *******************************************************
ok: [vagrant] => {
    "msg": "master@e0f47a8 deployed as release 20150412053432"
}
```

group_vars/development
```diff
@@ -3,10 +3,10 @@ mysql_root_password: devpw
 web_user: vagrant
 
 wordpress_sites:
-  example.dev:
+  example.com:
```
fullyint: However, for `group_vars/development`, I think site name should remain `.dev` because maybe you do have a live `.com` even while working on the `.dev`. You wouldn't want vagrant-hostsupdater messing with the hosts file entry for `.com`. So, in `group_vars/development`, I'd suggest the defaults be site name `example.dev` but local path `example.com`.

roles/deploy/tasks/main.yml
```diff
@@ -82,5 +76,17 @@
...
+  with_items: project_post_finalize_commands
+
+- name: Get current git commit
+  command: git rev-parse --short HEAD
+  args:
+    chdir: "{{ project_source_path }}"
+  register: git_commit
+
+- debug:
+    msg: "{{ project_version }}@{{ git_commit.stdout }} deployed as release {{ deploy_helper.new_release }}"
```
fullyint: Instead of adding a new task to "Get current git commit", try just adding `register: git_commit` to the existing "Clone project files" task. Then get the SHA with something like `{{ git_commit.after | truncate(7, true, '') }}`.

group_vars/development
```diff
@@ -3,10 +3,10 @@ mysql_root_password: devpw
...
     site_hosts:
       - example.dev
-    local_path: '../example.dev' # path targeting local project directory (relative to root/Vagrantfile)
+    local_path: '../example.com' # path targeting local project directory (relative to root/Vagrantfile)
```
fullyint: Thinking about it more, I realize I'm _still_ sometimes confused about the required name for my bedrock directory. Of course, the bedrock directory doesn't require any name in particular, so long as `local_path` matches.

It may minimize confusion to refer to `my-bedrock`. So, in `group_vars/<environment>`:
`local_path: '../`**my-bedrock**` # path targeting local`**BEDROCK**`directory'`

and in the README:
```
project/                   - Primary folder for the project
├── my-bedrock-ansible/    - This repo
└── my-bedrock/            - A Bedrock-based site (could name it 'example.com')
```

I realize that the bedrock directory is no longer just bedrock, it is a full-on "bedrock-based site", as you've called it, because of customizations, installation of dependencies, etc. So, technically you've been more accurate in calling it a 'bedrock-based site' or 'local project directory'. But I'm not sure that those phrases will always immediately translate in people's minds to 'my bedrock directory'. You're abstract and correct, but when I'm new, I need concrete.

roles/deploy/defaults/main.yml
```diff
@@ -71,7 +68,7 @@ project_environment: {}
 # `new_release_path` on the remote server (for `project_pre_build_commands` and `project_post_build_commands`).
 # Examples:
 # project_pre_build_commands_local:
-#   - path: ../example.dev/web/app/themes/sage
+#   - path: "{{ project.local_path }}"/web/app/themes/sage"
```
fullyint: quotes

group_vars/development
swalkinshaw: Only `site_hosts` get added as hosts, not the key. I had staging as `staging.example.com` but it seems like the site **name** should be consistent across all the env files. After all it's the same project. It's just the settings that vary.

This name is only descriptive and is never used as an actual "domain". The project is called "example.com".

roles/deploy/tasks/main.yml
swalkinshaw: That's awesome. Never knew about `before`/`after` (which I swear aren't anywhere in the docs).

fullyint: 😊 I tried it one day, just out of sheer hope. Haven't seen it documented.

swalkinshaw:
Did some naming updates + git commit.

group_vars/development
fullyint: Ah I see. Man, I still have some things to sort out in my mind on this. Maybe this confusion is my personal problem, not a general problem. If so, it may annul my other comment re: `my-bedrock`.

swalkinshaw: I reverted some naming updates after talking with @retlehs but it's still a little now with more descriptive comments.

@fullyint think of it in the context of deploying:

`./deploy.sh staging example.com`

`./deploy.sh production example.com`

You're still deploying the same project/site, just doing it to a different host.

Rather than: `./deploy.sh staging staging.example.com`

fullyint: Ok, yeah, this discussion gives me clarity on the distinction between site name and host names in group_vars. I had them all in one hazy mental mess.

deploy.sh
```diff
@@ -16,7 +16,7 @@ Available environments:
 `( IFS=$'\n'; echo "${ENVIRONMENTS[*]}" )`
 
 Examples:
-  deploy staging example.dev
+  deploy staging staging.example.com
```
fullyint: So, this now will drop the `staging` subdomain and be...
`deploy staging example.com`

swalkinshaw: Good catch, updated.

mAAdhaTTah:
This is nice; that naming conversation makes me wonder: right now I have one BA repo for each website, but maybe it makes more sense to just have one, but is it possible to only bring up one site at a time? I found in the past that multiple sites on a single server made it run slow, but if I can run `vagrant up --site=example.com` and it only provisions the one site, that would make things a lot easier. This possible? Thoughts?

swalkinshaw:
@mAAdhaTTah yeah that would be quite easy via the `Vagrantfile`. I'm not sure I'll include it in this repo but I'll take a look at what it would take.

mAAdhaTTah:
Yeah, not really doable, I don't think, because ansible provisions all of the sites by default, and you can't provision a site by a flag (unless you know something I don't). You could do a big find->replace to make the sites swappable, but then it's either one site at a time or all sites all the time.

swalkinshaw:
@mAAdhaTTah oh you're right actually. For some reason I was only thinking about it from Vagrant's point of view (as in skipping synced dirs). Didn't think about the Ansible side.

One alternative solution: use one BA repo + multiple Vagrant's. You could set a different [group name](https://github.com/roots/bedrock-ansible/blob/fcb5d55e5822f8da4ed4861cc0ea465434ad2a3c/Vagrantfile#L57-L60) and then create more `group_vars/<group>` for them.

fullyint:
Yesterday I secretly spent time (then aborted after failure) trying to pull values out of ARGV inside Vagrantfile, pass them to a var in `ansible.extra_vars` (in `Vagrantfile`), grab the values in `group_vars/<environment>` and use them to subset the `wordpress_sites` dict. No luck. 

An alternative might have been to use those ARGV values is a var `selected_sites` and add something to the conditional for each task using the `with_dict` option, a la
```
with_dict: wordpress_sites
when: item.key in selected_sites
```
Not sure if there's actually potential there. It's all still a little hazy to me.

mAAdhaTTah:
I got it working with @swalkinshaw's group name and an environment variable ([suggested here](https://github.com/mitchellh/vagrant/issues/2064)). Run with `VSITE=example.com vagrant up`.

```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :

require 'yaml'

ANSIBLE_PATH = '.' # path targeting Ansible directory (relative to Vagrantfile)

# Set Ansible roles_path relative to Ansible directory
ENV['ANSIBLE_ROLES_PATH'] = File.join(ANSIBLE_PATH, 'vendor', 'roles')

if ENV['VSITE'].nil?
  config_file = File.join(ANSIBLE_PATH, 'group_vars/development')
else
  config_file = File.join(ANSIBLE_PATH, 'group_vars', ENV['VSITE'])
end

if File.exists?(config_file)
  wordpress_sites = YAML.load_file(config_file)['wordpress_sites']
  raise "no sites found in #{config_file}" if wordpress_sites.to_h.empty?
else
  raise "#{config_file} file not found. Please set `ANSIBLE_PATH` in Vagrantfile"
end

Vagrant.require_version '>= 1.5.1'

Vagrant.configure('2') do |config|
  config.vm.box = 'roots/bedrock'
  config.ssh.forward_agent = true

  # Required for NFS to work, pick any local IP
  config.vm.network :private_network, ip: '192.168.50.5'

  hostname, *aliases = wordpress_sites.flat_map { |(_name, site)| site['site_hosts'] }
  config.vm.hostname = hostname

  if Vagrant.has_plugin? 'vagrant-hostsupdater'
    config.hostsupdater.aliases = aliases
  else
    puts 'vagrant-hostsupdater missing, please install the plugin:'
    puts 'vagrant plugin install vagrant-hostsupdater'
  end

  if Vagrant::Util::Platform.windows?
    wordpress_sites.each do |(name, site)|
      config.vm.synced_folder site['local_path'], remote_site_path(name), owner: 'vagrant', group: 'www-data', mount_options: ['dmode=776', 'fmode=775']
    end
  else
    if !Vagrant.has_plugin? 'vagrant-bindfs'
      raise Vagrant::Errors::VagrantError.new,
        "vagrant-bindfs missing, please install the plugin:\nvagrant plugin install vagrant-bindfs"
    else
      wordpress_sites.each do |(name, site)|
        config.vm.synced_folder site['local_path'], nfs_path(name), type: 'nfs'
        config.bindfs.bind_folder nfs_path(name), remote_site_path(name), u: 'vagrant', g: 'www-data'
      end
    end
  end

  config.vm.provision :ansible do |ansible|
    ansible.playbook = File.join(ANSIBLE_PATH, 'dev.yml')
    ansible.extra_vars = { ansible_ssh_user: 'vagrant' }

    if ENV['VSITE'].nil?
      ansible.groups = {
        'web' => ['default'],
        'development' => ['default']
      }
    else
      ansible.groups = {
        'web' => ['default'],
        ENV['VSITE'] => ['default']
      }
    end
  end

  config.vm.provider 'virtualbox' do |vb|
    # Give VM access to all cpu cores on the host
    cpus = case RbConfig::CONFIG['host_os']
      when /darwin/ then `sysctl -n hw.ncpu`.to_i
      when /linux/ then `nproc`.to_i
      else 2
    end

    # Customize memory in MB
    vb.customize ['modifyvm', :id, '--memory', 1024]
    vb.customize ['modifyvm', :id, '--cpus', cpus]

    # Fix for slow external network connections
    vb.customize ['modifyvm', :id, '--natdnshostresolver1', 'on']
    vb.customize ['modifyvm', :id, '--natdnsproxy1', 'on']
  end
end

def nfs_path(site_name)
  "/vagrant-nfs-#{site_name}"
end

def remote_site_path(site_name)
  File.join('/srv/www/', site_name, 'current')
end
```

swalkinshaw:
If you want it a little simpler:

```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :

require 'yaml'

ANSIBLE_PATH = '.' # path targeting Ansible directory (relative to Vagrantfile)

# Set Ansible roles_path relative to Ansible directory
ENV['ANSIBLE_ROLES_PATH'] = File.join(ANSIBLE_PATH, 'vendor', 'roles')

SITE_GROUP = ENV.fetch('VSITE', 'development')

config_file = File.join(ANSIBLE_PATH, 'group_vars', SITE_GROUP)

...
  config.vm.provision :ansible do |ansible|
    ansible.playbook = File.join(ANSIBLE_PATH, 'dev.yml')
    ansible.extra_vars = { ansible_ssh_user: 'vagrant' }

    ansible.groups = {
      'web' => ['default'],
      SITE_GROUP => ['default']
    }
    end
```

Good work :+1:

PR #184, opened 2015-04-21
Adding text about not needing relative folder paths

kalenjohnson:
Since it's not a requirement to use one bedrock-ansible VM per project, a quick note about how the folder structure is not strict would be helpful.


kalenjohnson:
Not necessary as the README was updated today.

PR #189, opened 2015-04-27
Use site instead of example.com

merchantguru:
To be consistent with Readme's example and the Roots Example Project example.


retlehs:
thanks!

PR #190, opened 2015-04-27
Specify Virtualbox as provider

merchantguru:
To be consistent with Roots Example Project. `vagrant up` fails otherwise because of the `mv ansible/Vagrantfile .` step.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,6 +1,9 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
+# Specify Vagrant provider as Virtualbox
+ENV['VAGRANT_DEFAULT_PROVIDER'] = 'virtualbox'
+
 require 'yaml'
 
 ANSIBLE_PATH = '.' # path targeting Ansible directory (relative to Vagrantfile)


swalkinshaw:
We can't do this by default though since out of the box this `Vagrantfile` wouldn't work anymore.

merchantguru:
What do you recommend to fix it? Both those points will make the Roots Example fail out of the box as is.

Foxaii:
Unless something has changed recently, Virtualbox is already the default Vagrant provider, so that shouldn't need to be stated (and afaik not stating it hasn't caused any issues to date). Can you expand on how it is "failing out of the box" for you?

The Roots example project documentation should show that if the Vagrantfile is moved, the Ansible path should be updated accordingly. This makes more sense than a customisation that would break every other usage.

merchantguru:
The provider definition is a simple fix for users who have a different default provider specified in their own environment. For example, I have parallels defined as my default provider. (I think Parallels did this to me). When I run `vagrant up`, it dies because `vagrant up` relies on an assumption that you will use the default provider unless you give it the --provider=virtualbox flag. This will help make the project work wherever. 

It's fine with me if the ansible path solution is to have it done in the Example project, but that creates another step in that project.

merchantguru:
Removed/withdrew the ansible path change from this pull request. Just suggesting the Vagrantfile change. (I didn't mean to mix the 2 commits in the first place)

swalkinshaw:
:+1:

PR #203, opened 2015-04-29
Windows compatibility

kalenjohnson:
This is getting the ball rolling on Windows compatibility.

I've incorporated the info in the wiki and fixed the shell script so that Ansible runs on the VM, the `ansible-galaxy` files get loaded correctly, and Ansible provisions the VM.

I've hit one snag where the [start php5-fpm service](https://github.com/roots/bedrock-ansible/blob/master/roles/php/tasks/main.yml#L20-21) errors, but the verbose output is basically useless. I'll need either some more Ansible or Windows experts to take a look.

Issues:
- [x] Ansible erroring on php5-fpm start: https://github.com/ansible/ansible/issues/10675
  - possible fix to use Ansible 1.9.0.1
- [x] Pass `windows.sh` the ansible directory in the vagrant box. Otherwise if the project places `bedrock-ansible` in a sub-directory, all calls to `/vagrant` need to be updated as well.
- [x] Decide how Windows users should be able to deploy. I can think of two options:
  - SSH in and run the same script everyone else does
  - since the `windows.sh` script is basically ready to go, they could potentially do something like `deploy=‘production' vagrant up`

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -52,16 +52,23 @@ Vagrant.configure('2') do |config|
     end
   end
 
-  config.vm.provision :ansible do |ansible|
-    ansible.playbook = File.join(ANSIBLE_PATH, 'dev.yml')
-    ansible.groups = {
-      'web' => ['default'],
-      'development' => ['default']
-    }
+  if Vagrant::Util::Platform.windows?
+    config.vm.provision :shell do |sh|
+      sh.path = File.join(ANSIBLE_PATH, 'windows.sh')
+      sh.args = ANSIBLE_PATH
+    end
+  else
+    config.vm.provision :ansible do |ansible|
+      ansible.playbook = File.join(ANSIBLE_PATH, 'dev.yml')
+      ansible.groups = {
+        'web' => ['default'],
+        'development' => ['default']
+      }
 
-    if vars = ENV['ANSIBLE_VARS']
-      extra_vars = Hash[vars.split(',').map { |pair| pair.split('=') }]
-      ansible.extra_vars = extra_vars
+      if vars = ENV['ANSIBLE_VARS']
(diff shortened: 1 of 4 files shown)


kalenjohnson:
I think the final issue is actually a bug in Ansible: https://github.com/ansible/ansible/issues/10675

People are experiencing an issue starting php5-fpm with version 1.9.1, which happens to be the latest version that gets installed on box, which also happens to be Ubuntu 14.04.

swalkinshaw:
Regarding deploys: rather them just SSH in and run commands.

austinpray:
>SSH in and run the same script everyone else does

This

kalenjohnson:
Ok I think we all agree on SSH in and follow the same instructions to deploy on Windows

austinpray:
Yep. Reduces the surface area we need to test.

Vagrantfile
```diff
@@ -52,16 +52,23 @@ Vagrant.configure('2') do |config|
...
-    }
-
-    if vars = ENV['ANSIBLE_VARS']
-      extra_vars = Hash[vars.split(',').map { |pair| pair.split('=') }]
-      ansible.extra_vars = extra_vars
+  if Vagrant::Util::Platform.windows?
+    config.vm.provision :shell do |sh|
+      sh.path = "#{ANSIBLE_PATH}/windows.sh"
+      sh.args = "#{ANSIBLE_PATH}"
+    end
```
swalkinshaw:
```ruby
config.vm.provision :shell do |sh|
  sh.path = File.join(ANSIBLE_PATH, 'windows.sh')
  sh.args = ANSIBLE_PATH
end
```

kalenjohnson:
Squashed

kalenjohnson:
Fully squashed

PR #210, opened 2015-05-06
Use inventory_hostname instead of ansible_hostname

merchantguru:
Fixes issue https://github.com/roots/bedrock-ansible/issues/191 where re-running the playbook can fail while setting the hostname's root user password for MariaDB.


swalkinshaw:
Thanks!

merchantguru:
Thanks for merging!

PR #211, opened 2015-05-07
Remove possibly confusing default option

louim:
`ferm_input_list` rule `disabled` default to false. including this in the `all` file might be confusing for the user, since the `ssh` rule (the case where the `disabled: false` was included) is one of the more important rule. if the user change it to `disabled: true`, they most likely will lock themselves out of their instance.


swalkinshaw:
Thanks!

PR #213, opened 2015-05-08
`server_name` in SSL and non-SSL blocks should mirror each other 

chriszarate:
In `roles/wordpress-setup/templates/wordpress-site.conf.j2`:

When TLS is enabled, the non-TLS server block adds `www.` to every item in `site_hosts`, whether you want it or not. This can lead to disparity between TLS and non-TLS server blocks, and makes using wildcard `site_hosts` impossible (e.g., `*.example.com`).

### DIFF
--- a/roles/wordpress-setup/templates/wordpress-site.conf.j2
+++ b/roles/wordpress-setup/templates/wordpress-site.conf.j2
@@ -45,10 +45,10 @@ server {
   include wordpress.conf;
 }
 
-{% for host in item.value.site_hosts if item.value.ssl is defined and item.value.ssl.enabled | default(False) %}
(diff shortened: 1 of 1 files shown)


swalkinshaw:
@chriszarate this is a lot better. Never liked that loop before.

Could you rebase this since the template has changed?

chriszarate:
Done. Thanks!

chriszarate:
Oops: Added `item.value.ssl is defined` check to match the template revisions you recently pushed.

swalkinshaw:
Thanks :+1:

fullyint:
Thanks @chriszarate. This solved some issues for me.

PR #214, opened 2015-05-08
Update wordpress-site.conf.j2

nathanielks:
In the event that item.value.ssl isn't defined, let's go ahead and check to see if it exists and if so, continue. This prevents errors when it's not defined, preventing the play from proceeding.


swalkinshaw:
:+1: squash

nathanielks:
squashed, good sir

PR #219, opened 2015-05-15
Fix conditional check for subtree in deploy

evanfuture:
Fixes "error while evaluating conditional" error when deploying multiple sites one by one.  Ansible needs to check if the variable "is defined" since its not a straight boolean true/false check.

### DIFF
--- a/roles/deploy/tasks/main.yml
+++ b/roles/deploy/tasks/main.yml
@@ -17,7 +17,7 @@
 
 - name: Move project subtree into root folder
   shell: mv {{ deploy_helper.new_release_path }}/{{ project_subtree }}/* {{ deploy_helper.new_release_path }}
-  when: project_subtree
+  when: project_subtree is defined
 
 - name: Remove unwanted files/folders from new release
   file: path="{{ deploy_helper.new_release_path }}/{{ item }}" state=absent


swalkinshaw:
Thanks!

slackday:
This doesn't seem to work for me. As I read it subtree defaults to False so it will always be defined?

I get this output:
```
=> {"changed": true, "cmd": "mv /mnt/persist/www/example.com/releases/20150517190042/False/* /mnt/persist/www/example.com/releases/20150517190042", "delta": "0:00:00.010619", "end": "2015-05-17 19:00:56.686640", "rc": 1, "start": "2015-05-17 19:00:56.676021", "warnings": []}
stderr: mv: cannot stat ‘/mnt/persist/www/example.com/releases/20150517190042/False/*’: No such file or directory

```

It seems to always output /False/* as the folder no matter what I change the variable to. I'm not so handy with Ansible but am trying to fiddle around a bit..

swalkinshaw:
@slackday thanks, should be fixed in 07047a2. No point in having the `default` anymore.

davekiss:
I'm getting this now with no subtree defined:

```
TASK: [deploy | Move project subtree into root folder] ************************ 
fatal: [ip.address] => One or more undefined variables: 'dict object' has no attribute 'subtree'
```

swalkinshaw:
@davekiss can you try with this patch:

```diff
- project_subtree: "{{ project.subtree }}"
+ project_subtree: "{{ project.subtree | default(omit) }}"
```

Found here: https://github.com/roots/bedrock-ansible/blob/4e93afbcd8ecafac76f2eeb41580dcea407b1056/roles/deploy/defaults/main.yml#L4

davekiss:
@swalkinshaw no go.

`stderr: mv: cannot stat ‘/srv/www/vimeography.com/releases/20150520132106/__omit_place_holder__0fae024f264debcf4f2677c603f189bd98dae9c8/*’: No such file or directory`

swalkinshaw:
@davekiss oops that makes sense. I'll figure it out.

davekiss:
For now, I reverted https://github.com/roots/bedrock-ansible/commit/07047a23840dd6da83acff3bee27ef67efd91d91 and added this condition instead:

`when: not project_subtree == 'False'`

swalkinshaw:
@davekiss thanks, updated in 56045fc

Issue #73, opened 2014-10-10
Notes about mysql passwords


fullyint:
Could the `README.md` could use [these notes](https://github.com/fullyint/bedrock-ansible/commit/154a355ca2289d6511ad0c32c9448c9bdeb88a67) about the mysql passwords?

### `mysql_root_password` with roots/bedrock box
The "Set root user password" task (`roles/mariadb/tasks/main.yml`) fails when I make any change to `mysql_root_password` (in `group_vars/development`) while `config.vm.box = 'roots/bedrock'`. I'm guessing the roots/bedrock box has the `mysql_root_password` set already.

### Special characters in db_password
When `db_import=true` and `db_password` contains any of the characters `$ & ( )`, the "Import database" task (`roles/wordpress-sites/tasks/database.yml`) fails with messages such as `mysql ERROR 1045 (28000): Access denied for user...`.

Some characters were fine: `a-z 0-9 ! % * # @ ^ - `. I didn't test any others. It also failed with [escaping](http://jinja.pocoo.org/docs/dev/templates/#escaping): `login_password={% raw %}"{{ item.env.db_password }}"{% endraw %}` (also without the double-quotes).

I don't see any adjustments to make aside from having the `README.md` warn users about special characters. I think the problem is with ansible's [`mysql_db`](http://docs.ansible.com/mysql_db_module.html) module ([source](https://github.com/ansible/ansible-modules-core/blob/devel/database/mysql_db.py)) when `state=import` (or maybe the problem is with the MySQLdb Python package being used).

The problem doesn't occur when `state=present`, as in the bedrock-ansible's task "Create database of sites" (e.g., when `mysql_root_password` contains special characters), nor when the `login_password` parameter is used with ansible's [`mysql_user`](http://docs.ansible.com/mysql_user_module.html) module ([source](https://github.com/ansible/ansible-modules-core/blob/devel/database/mysql_user.py); used several times throughout the bedrock-ansible playbook).

A few potentially related notes over at ansible: [1](https://github.com/ansible/ansible/issues/7053), [2](https://github.com/ansible/ansible/issues/8137), [3](https://github.com/ansible/ansible/issues/8161).

thiagodebastos:
Is this still supposed to be happening? I arrived here because I faced the same issue.

ramielbarsoum:
[...]

PR #227, opened 2015-05-26
Remove mysql_root_password from development config

louim:
As mentioned in #73 , changing this value to anything else in development, will cause the playbook to fail, since the base box MariaDB root password is already set. Removing the option will reduce the likeliness of someone having the error, until a permanent fix can be made.

### DIFF
--- a/group_vars/development
+++ b/group_vars/development
@@ -1,5 +1,3 @@
-mysql_root_password: devpw
-
 web_user: vagrant
 
 wordpress_sites:


swalkinshaw:
Won't this break things?

Right now `login_password` is using `mysql_root_password` in a bunch of places. Which I think would just result in failures.

louim:
My bad, I assumed there was a default set somewhere in the role. I plan on submitting a PR to make the mysql role idempotent.

PR #228, opened 2015-05-29
Disable exposing PHP version to the world

louim:
Do not expose PHP version via header, for security reasons.


swalkinshaw:
I'm good with this :+1:

PR #231, opened 2015-06-02
Option to allow remote connection to dev db

JacobDorman:
Not sure if this is wanted but was useful to me (allowing IDE to connect to the vagrant db)

### DIFF
--- a/group_vars/development
+++ b/group_vars/development
@@ -1,4 +1,5 @@
 mysql_root_password: devpw
+db_remote_allow: false
 
 web_user: vagrant
 
@@ -42,3 +43,8 @@ php_xdebug_remote_port: 9000
 php_xdebug_remote_log: /tmp/xdebug.log
 php_xdebug_idekey: XDEBUG
 php_max_nesting_level: 200
+
+ferm_input_group_list:
+  - type: "dport_accept"
+    dport: ["mysql"]
+    filename: "mysql_accept"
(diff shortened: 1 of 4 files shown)


retlehs:
why did you need this in order for your IDE to connect to the DB?

what's your IDE / does it not support SSH connections?

image

swalkinshaw:
@JacobDorman does the above work? I'd rather avoid this. I like that you made it development only but it's always good to avoid `lineinfile` when possible too.

louim:
I have this working from another project, but our use case is different. When more than one dev work on the same site, we use a remote database (usually the staging database). We allow remote access and we use the database with the local Vagrant setup.

it allow to pass a list of host to open the database to. ex:

```
db_allowed_hosts:
  - localhost
  - "123.123.123.123"
  - "%"
```

it is database dependant, so I think it is more elegant and allow finer grained control if you have many sites on the same server. I could prepare a pull request if it is something that you might consider adding. My intention was to release it as a Galaxy playbook like I did with https://galaxy.ansible.com/list#/roles/3714, but I haven't checked if that's actually doable.

swalkinshaw:
Closing for now until there's a reason why an SSH connection can't be used.

JacobDorman:
Sorry for the delay... SSH tunnel is a better solution. Here are the docs for [IntelliJ](https://www.jetbrains.com/idea/webhelp13/configure-ssh-tunnel-dialog.html) is case anyone needs.

Issue #186, opened 2015-04-24
Implement microcaching


louim:
Hello guys, 

I would to get feedback on implementing microcaching in Bedrock-Ansible. I recently implemented it for the first time on a Bedrock, following instruction from http://reviewsignal.com/blog/2014/06/25/40-million-hits-a-day-on-wordpress-using-a-10-vps/ and it does help tredemously with sudden traffic spikes. Since the cache time is really low, there is no need to implement a flushing mecanism.

The way I see this working could be two settings: 
- Enabled or disabled
- cache time (recommended amount would be around 30 seconds)

Would you be interested in such pull request? I think it will provide added value, with little drawback, but I wanted to have your opinions before doing such thing.

retlehs:
thanks for bringing this up, we are open to this! @swalkinshaw might follow up later with some more details.

louim:
I believe this can be closed :smile:

PR #234, opened 2015-06-05
Fastcgi caching

louim:
Hello, as mentioned in #186 this is a "microcaching" setup we have used in the past with Bedrock. 

I have a couple of question about where should I go from now.

* I implemented the cache globally, as in "enabled" or "disabled" for all sites on the server. Should it stays like that or should that be handled for each sites separately?
* I initially toyed with the cache residing in `/var/run` since it is in memory instead of being on the filesystem. However I decided to scrap the idea, because it would require changing the Nginx init script to  create the folder after each server reboot, since `/var/run` is cleared. What do you think?
* Cache size: should this be a fixed size, a size based on the number of sites on the server, or just user configurable?
* Should the cache be enabled by default? (I think it should)
* Should this feature be documented in the Readme or the wiki?

Comments are welcome.

I will be following shortly with another pull request that aim to prevent website DDOS using 404s. That was one of the weak point we saw when using the cache.

### DIFF
--- a/roles/nginx/tasks/main.yml
+++ b/roles/nginx/tasks/main.yml
@@ -14,7 +14,7 @@
 - name: Move h5bp configs
   command: creates=/etc/nginx/h5bp/ cp -R /etc/nginx/h5bp-server-configs/h5bp /etc/nginx/h5bp
 
-- name: Copy SSL conf files
+- name: Copy conf files
   copy: src="{{ item }}" dest=/etc/nginx/{{ item | basename }} mode=644
   with_fileglob: '*'
 
--- a/roles/nginx/templates/nginx.conf.j2
+++ b/roles/nginx/templates/nginx.conf.j2
@@ -36,6 +36,15 @@ http {
   # Hide nginx version information.
   server_tokens off;
 
+  # Setup the fastcgi cache.
+  fastcgi_cache_path {{ nginx_cache_path }} levels=1:2 keys_zone=wordpress:{{ nginx_cache_key_storage_size }} max_size={{ nginx_cache_size }} inactive={{ nginx_cache_inactive }};
+  fastcgi_cache_use_stale updating error timeout invalid_header http_500;
+  fastcgi_cache_lock on;
+  fastcgi_cache_key $scheme$host$request_uri$request_method;
+  fastcgi_ignore_headers Cache-Control Expires Set-Cookie;
+  fastcgi_pass_header Set-Cookie;
+  fastcgi_pass_header Cookie;
+
   # Define the MIME types for files.
   include       h5bp-server-configs/mime.types;
   default_type  application/octet-stream;
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Configure complete [Bedrock](https://roots.io/bedrock/)-based WordPress ready se
 bedrock-ansible will configure a server with the following and more:
 
 * Ubuntu 14.04 Trusty LTS
-* Nginx
+* Nginx (with optional FastCGI caching)
 * PHP 5.6 (or [HHVM](http://hhvm.com/))
 * [MariaDB](https://mariadb.org/) as a drop-in MySQL replacement (but better)
 * sSMTP (mail delivery)
@@ -116,6 +116,9 @@ For a complete, working example you can see our [example project](https://github
   * `enabled` - Multisite enabled flag (required, set to `false`)
   * `subdomains` - subdomains option
   * `base_path` - base path/current site path
+* `cache` - hash of cache options
+  * `enabled` - Cache enabled flag (required, set to `false`)
+  * `duration` - Duration of the cache (default: `30s`)
 * `env` - environment variables
   * `wp_home` - `WP_HOME` constant (required)
   * `wp_siteurl` - `WP_SITEURL` constant (required)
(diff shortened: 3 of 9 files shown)


roles/nginx/files/fastcgi_cache.conf
```diff
@@ -0,0 +1,19 @@
+set $skip_cache 0;
+
+# POST requests and urls with a query string should always go to PHP
+if ($request_method = POST) {
```
austinpray: I think your logic is backwards. The caching methods should be a whitelist

austinpray: In other words: you should only be caching GET calls.

louim: @austinpray You are totally right, in fact it is not even used as Nginx default cached method are only GET and HEAD. http://nginx.org/en/docs/http/ngx_http_fastcgi_module.html#fastcgi_cache_methods

I updated accordingly.

roles/nginx/files/fastcgi_cache.conf
```diff
@@ -0,0 +1,15 @@
...
+  set $skip_cache 1;
+}
+
+# Don't cache uris containing the following segments
+if ($request_uri ~* "/wp-admin/|/xmlrpc.php|wp-.*.php|/feed/|index.php|sitemap(_index)?.xml") {
+  set $skip_cache 1;
+}
+
+# Don't use the cache for logged in users or recent commenters
+if ($http_cookie ~* "comment_author|wordpress_[a-f0-9]+|wp-postpass|wordpress_no_cache|wordpress_logged_in") {
```
swalkinshaw: I think this string (and the above for require uri) should be stored in defaults as a variable. Then they can easily be overridden.

roles/nginx/tasks/main.yml
```diff
@@ -35,3 +35,9 @@
 
 - name: Create base WordPress subdirectory Multisite config
   template: src=wordpress_multisite_subdirectories.conf.j2 dest=/etc/nginx/wordpress_multisite_subdirectories.conf
+
+- name: Ensure Cache folder exist
```
swalkinshaw: Ensure cache folder exists

roles/nginx/templates/nginx.conf.j2
```diff
@@ -36,6 +36,17 @@ http {
   # Hide nginx version information.
   server_tokens off;
 
+  {% if nginx_use_fastcgi_cache %}
+  # Setup the fastcgi cache.
+  fastcgi_cache_path {{ nginx_cache_path }} levels=1:2 keys_zone=wordpress:10m max_size=1024m inactive=1h;
```
swalkinshaw: These times and sizes should also be in `defaults` as variables.

swalkinshaw:
@louim thanks for this. It's looking mostly good. For your questions:

1. I'm okay with just making it globally enabled for now and not per-site.
2. I think I'd prefer `/var/run`. I don't mind that the cache won't be persisted across restarts since it's just a temporary micro cache anyway. Does this directory really need to be created? I think Nginx just handles it automatically. But there's a bigger problem with `/var/run`. I think it's set to 10% of RAM which means that the Nginx cache would need to be set as less than that. So might just be best to avoid all this.
3. Let's pick a good default and make it a variable. I'd probably go with less than 1GB though. 
4. I'll have to think about making it default.
5. Mentioned in README as a feature + explained in Wiki I think.

louim:
@swalkinshaw thanks for the comments, I updated the PR to fix what you mentioned. Just to follow up on you answers:

1. It would actually be easy to add as per site if that's what you think is best. Juste need to add a key in the config dict to turn it on, and tweak the global condition for the `nginx.conf`
2. I would too prefer `/var/run` for performances reason. But I tested it and no folder are persisted in `/var/run`. Most Deamon create their own folders when starting. See http://askubuntu.com/questions/303120/how-folders-created-in-var-run-on-each-reboot. The solution would imply having the directory creation added to the init script. As you mentioned, the size is also an issue. For small VPS with 1GB RAM, this could be an issue as the cache will be really small.
3. 250MB would be good I think. HTML doesn't really take many space.
4. I actually pondered about it and I think it should not be enabled by default. Even if the cache is only a small period of time, It can make people not aware of it wonder why their modifications don't appear right away. Best leave the choice to the user to knowingly enable it.
5. :+1:

swalkinshaw:
@louim:

1. per site would be better so let's do that :+1:. Maybe the `keys_zone` should have the site name in it too?
3. 250MB is a good default
4. Yep, agreed.

louim:
@swalkinshaw I made the config site specific. It required me to move the `location ~ \.php$` block from the generic include to the site-specific file since I the content now change depending of the sites.

Settings can be used like : 
```yml
...
ssl:
  enabled: false
cache:
  enabled: true
  duration: 30s
  skip_cache_uri: "some string here"
  skip_cache_cookie: "some other string here"
system_cron: true
...
```

All params will take the default global value when not specifically defined in a site. I also added `fastcgi_cache_lock on;` so only one request hit the backend when generating a new cache.

Just let me know if everything looks good to you, and Ill write the documentation and squash my commits. 

As a side note, If we one day decide to compile NGINX from source (saw that mentioned in #235), it would be interesting to compile it with https://github.com/FRiCKLE/ngx_cache_purge, so the setup can be used with longer cache time. This would allow people to use plugin like https://wordpress.org/plugins/nginx-helper/ to purge their cache from the wordpress admin.

swalkinshaw:
@louim this looks great. Want to squash and update readme/wiki?

louim:
@swalkinshaw I'm unable to edit / create new page to the wiki. Would you mind opening the wiki for public edit or adding me as a collaborator so I can edit it?

swalkinshaw:
@louim it's public now

louim:
@swalkinshaw Updated the wiki and the readme, rebased against the current master. Should be good to go! :+1:

swalkinshaw:
Tested this out and it's working great. Thanks @louim

PR #238, opened 2015-06-09
Insert full path to service command, add hhvm restart minute

davenaylor:
This ensures that HHVM is restarted just once and not every minute between 2 and 3am. Also provides for full path to the service command.


swalkinshaw:
Well that was a terrible bug. Thanks!

PR #239, opened 2015-06-10
Make changing mysql root password possible.

### LINKED ISSUE #237: Make the MySQL (MariaDB) role idempotent
Currently, on the Vagrant `development` environment, it is impossible to change the mysql root password, because it is already set in the base box. On `staging`and `production`, it's only possible to set the password the first time because it is not already defined. Any attempt to change it afterward will fail.

To make the playbook idempotent, the root password should be stored in the `~/my.cnf` file as recommended by Ansible: http://docs.ansible.com/mysql_user_module.html#notes

See https://github.com/geerlingguy/ansible-role-mysql/blob/master/tasks/secure-installation.yml#L2-L22 for a working implementation, although I'm not sure if all of it is needed.

louim:
This pull request makes it possible to change the root password. it fixes #237. I also removed the option `mysql_user` because it wasn't really doing anything. (The user is hardcoded to `root` in the task and it's not possible to change the root username without running some SQL).

This should only be merged after @swalkinshaw update the base box, else it's gonna break in the development environment.

### DIFF
--- a/group_vars/all
+++ b/group_vars/all
@@ -1,7 +1,8 @@
 ---
 apt_cache_valid_time: 86400
 mariadb_dist: trusty
-mysql_user: root
+mysql_remote_database: false
+mysql_root_user: root
 www_root: /srv/www
 
 mail_smtp_server: smtp.mandrillapp.com:587
--- a/roles/mariadb/templates/my.cnf.j2
+++ b/roles/mariadb/templates/my.cnf.j2
@@ -1,0 +1,3 @@
+[client]
+user=root
+password={{ mysql_root_password }}
--- a/dev.yml
+++ b/dev.yml
@@ -10,7 +10,7 @@
     - { role: ferm, tags: [ferm] }
     - { role: ntp }
     - { role: sshd, tags: [sshd] }
-    - { role: mariadb, tags: [mariadb] }
+    - { role: mariadb, tags: [mariadb], when: not mysql_remote_database }
     - { role: ssmtp, tags: [ssmtp mail] }
     - { role: php, when: not hhvm, tags: [php] }
     - { role: hhvm, when: hhvm, tags: [hhvm] }
--- a/roles/mariadb/tasks/main.yml
+++ b/roles/mariadb/tasks/main.yml
@@ -20,8 +20,6 @@
               host="{{ item }}"
               password="{{ mysql_root_password }}"
               check_implicit_admin=yes
-              login_user="{{ mysql_user }}"
-              login_password="{{ mysql_root_password }}"
               state=present
   with_items:
     - "{{ inventory_hostname }}"
@@ -29,11 +27,17 @@
     - ::1
     - localhost
 
+- name: Copy .my.cnf file with root password credentials.
+  template:
+    src: my.cnf.j2
+    dest: ~/.my.cnf
+    owner: root
+    group: root
+    mode: 0600
+
 - name: Delete anonymous MySQL server users
   mysql_user: user=""
               host="{{ item }}"
-              login_user="{{ mysql_user }}"
-              login_password="{{ mysql_root_password }}"
               state=absent
   with_items:
     - localhost
@@ -43,5 +47,3 @@
 - name: Remove the test database
   mysql_db: name=test
             state=absent
-            login_user="{{ mysql_user }}"
-            login_password="{{ mysql_root_password }}"
--- a/roles/wordpress-setup/tasks/database.yml
+++ b/roles/wordpress-setup/tasks/database.yml
@@ -3,10 +3,10 @@
   mysql_db: name="{{ item.value.env.db_name | default(item.key) }}"
             state=present
             login_host="{{ item.value.env.db_host | default('localhost') }}"
-            login_user="{{ mysql_user }}"
+            login_user="{{ mysql_root_user }}"
             login_password="{{ mysql_root_password }}"
   with_dict: wordpress_sites
-  when: item.value.db_create | default(True)
+  when: not mysql_remote_database and item.value.db_create | default(True)
 
 - name: Create/assign database user to db and grant permissions
   mysql_user: name="{{ item.value.env.db_user }}"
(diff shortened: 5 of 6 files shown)


swalkinshaw:
I'll take a better look at this tomorrow but two quick points:

* I think `mysql_user` was there for people who wanted to use an external database like RDS for instance (which doesn't use `root`). Ideally the playbook supports both internal/external DBs.
* The current password solution was designed **not** to write a config file on the server with a plain text password in it. Not sure how important this is but it was at least done on purpose.

louim:
`mysql_user` as it was implemented wasn't working because it was referenced in the `Set root user password` task of the MariaDB playbook. The task will fail if the value of `mysql_user` is not `root`. I think creating `mysql_use_remote` and completely skipping the playbook when set to true would be the best solution.

I see your point. But it has the major flaw of only allowing you to set the password **once**. It will work the first time because of the `check_implicit_admin`. Any change after that will fail because it will try to log in as root with the new password that you want to set, instead of the old one that you want to replace. 

I also think that a file with `600`permission in the root home is pretty safe. It seems like this the standard procedure for doing it.

I will fix a couple of things tomorrow, including the remote database access. :+1:

louim:
I had a look during the weekend, and there are a couple of things that need need to be fixed, before the remote user work. Currently when the MySQL users are created, they are created only with localhost, and not remote ip access. It means that the web server wouln't be able to connect to the database server (RDS for example). Ill have a look this week. 

@swalkinshaw do you have a preference on how should I go about that? Wildcard the host to '%', grab the webserver IP or allow user to set the list themselves?

swalkinshaw:
@louim my original idea wasn't to actually create remote databases. I just wanted it flexible that if someone manually made an RDS database they could just enter the credentials under the sites `env` property and it would just work.

So the two options would be:

1. Everything is local and it creates the databases/user/etc.
2. Everything is remote and it just uses those DB credentials to connect through WP.

So there might be a less confusing way of achieving those options.

swalkinshaw:
Going to close this since it doesn't really apply after https://github.com/roots/trellis/pull/260. If there's any separate changes for the remote DB option, then another PR would be better :+1:

louim:
@swalkinshaw #260 didn't really fix the underlying problem that currently the root mysql password cannot be changed once set for the first time.

As I explained in #237, this caused problem on the Vagrant environment due to the already provisioned role, but it also affected `staging` and `production`. #260 fixed the first provisioning of the password, but not the subsequent ones. 

You can try it for yourself by provisioning locally, then changing `mysql_root_password` and trying to provision again.

swalkinshaw:
@louim oh yeah good point. I got all these issues/PRs confused a bit. Let's re-open this. The remote user option still needs to be addressed though.

louim:
@swalkinshaw Like you said, I opted for the simple route. When `mysql_enable_remote` is true, user and database creation will not occur (globally, not at site level). MariaDB role will not be run.

If running for the first time with an already provisioned VM / Server, it will fail because no `~/.my.cnf` will be found for the root user. Solution is to create one manually, or just destroy the VM and recreate it. 

If provisioning a brand new server, it will work correctly from the start without the need to create the file manually.

group_vars/all
```diff
@@ -1,7 +1,8 @@
 ---
 apt_cache_valid_time: 86400
 mariadb_dist: trusty
-mysql_user: root
+mysql_enable_remote: false
```
swalkinshaw: What about `mysql_remote_database`?

louim:
@swalkinshaw rebased against current master, renamed the variable and squashed.

austinpray:
Gonna do some testing to see if I can make this backwards compatible

BrandonShutter:
Note, this doesn't install `mysql` at all, so you automatically get a database error.  I could be wrong, but don't we need to install it for the remote connection to work?

louim:
@BrandonShutter Mysql is not needed for Remote connection. This is handled by the mysqli (or PDO) php extension. Note that when using `mysql_remote_database`, the remote database / users wont be created. It's left to the user to configure his remote instance.

roles/mariadb/templates/my.cnf.j2
```diff
@@ -0,0 +1,3 @@
+[client]
+user=root
```
austinpray: should this be {{ mysql_root_user }} ?

austinpray:
So I went ahead and tested:

```
git checkout master
vagrant up
git checkout -b louim-mysql_root_password master
git pull https://github.com/louim/bedrock-ansible.git mysql_root_password
vagrant provision
```

which gives

```
TASK: [mariadb | Set root user password] **************************************
failed: [default] => (item=default) => {"failed": true, "item": "default"}
msg: unable to connect to database, check login_user and login_password are correct or ~/.my.cnf has the credentials
failed: [default] => (item=127.0.0.1) => {"failed": true, "item": "127.0.0.1"}
msg: unable to connect to database, check login_user and login_password are correct or ~/.my.cnf has the credentials
failed: [default] => (item=::1) => {"failed": true, "item": "::1"}
msg: unable to connect to database, check login_user and login_password are correct or ~/.my.cnf has the credentials
failed: [default] => (item=localhost) => {"failed": true, "item": "localhost"}
msg: unable to connect to database, check login_user and login_password are correct or ~/.my.cnf has the credentials
```

I tried fooling around with doing things like moving the copy .my.cnf block above the set root user password but registering a variable as to what was previously there (or not there). I'm sure we could get it to work but seeing as though the current implementation is broken: let's just let this be a breaking change. 

A user wishing to patch their servers can just temporarily change their mysql root user role:

```diff
 - name: Set root user password
   mysql_user: name=root
               host="{{ item }}"
               password="{{ mysql_root_password }}"
               check_implicit_admin=yes
               state=present
+              login_user="root"
+              login_password="{{ mysql_root_password }}"
```

And then remove those lines when the server is patched. Either that or destroy their server and re provision.

roles/mariadb/templates/my.cnf.j2
louim: We could do that, but it wouldn't work if the user changer the value afterward, because MySQL save the users in the database, we would need to implement a way to modify the root user via a MySQL command. I looked around, and this is not recommended. (Places like Amazon RDS that lets you set the root user won't let you change afterwards. So the `mysql_root_user` is only used on external installs. )

austinpray:
I'm good with merging this today unless you have any reservations @louim

louim:
You have my blessing :+1:

PR #241, opened 2015-06-14
Provide a mechanism for custom Nginx includes.

chriszarate:
I'm in the process of moving our multisite (subdomain) WP network to Bedrock, and using Bedrock-Ansible for provisioning of dev (and eventually production) instances. Thank you so much for this project!

One thing that would be very useful is a stock way to extend the default Nginx server conf (`roles/wordpress-setup/templates/wordpress-site.conf`). In any large installation, there is always additional configuration needed—for legacy redirects, `proxy_pass`es to other services, etc.

It would be nice to stick with the stock `wordpress-site.conf` (rather than overwrite/adjust it with our own `template` or `lineinfile` tasks). I'd like to propose baking in a wildcard `include` statement that would provide a place for site-specific Nginx server directives.

This PR creates `/etc/nginx/includes.d/[site]/` for each site in `wordpress_sites`, then adds a wildcard include statement to `/etc/nginx/sites-available/[site].conf`. The placement, right above the `include/wordpress.conf`, is designed to allow site-level rewrites in `includes.d` to take precedence.

Of course, the user can ignore this folder and not provide any site-level includes, with no ill effects.

Happy to discuss alternative strategies, or perhaps you've considered something like this and ruled it out. Either way, thanks for Bedrock-Ansible.

### DIFF
--- a/roles/wordpress-setup/tasks/nginx.yml
+++ b/roles/wordpress-setup/tasks/nginx.yml
@@ -12,6 +12,10 @@
   with_dict: wordpress_sites
   when: item.value.ssl.enabled | default(False)
 
+- name: Create includes.d directories
+  file: path="/etc/nginx/includes.d/{{ item.key }}" state=directory mode=0755
+  with_dict: wordpress_sites
+
 - name: Create WordPress configuration for Nginx
   template: src="wordpress-site.conf.j2"
             dest="/etc/nginx/sites-available/{{ item.key }}.conf"
--- a/roles/wordpress-setup/templates/wordpress-site.conf.j2
+++ b/roles/wordpress-setup/templates/wordpress-site.conf.j2
@@ -42,6 +42,7 @@ server {
   ssl_certificate_key     /etc/nginx/ssl/{{ item.value.ssl.key | basename }};
   {% endif %}
 
+  include includes.d/{{ item.key }}/*.conf;
   include wordpress.conf;
 }
 


swalkinshaw:
This is definitely needed and I like this solution a lot.

Thanks :+1:

louim:
How would you provide such files to be loaded?

swalkinshaw:
@louim good question that I overlooked too. You'll always need to customize the tasks I guess but at least with this you can leave the base template alone. I'm not sure how much more we can do beyond that.

We could have a task that copies over templates/files into the `includes.d` folder although it would have to be aware of which site it belonged to somehow.

@chriszarate any ideas on this?

louim:
I'm thinking a task that would grab the content of the `files/includes.d/[site]/*` and copy it to the machine. It would be the users responsibility to name the folder the same way that the site key for it to be picked. It should be pretty easy to do in Ansible. Or maybe the same way the SSL file are loaded, and leave the option to the user? 

I prefer option one.

chriszarate:
I may be an outlier, but I provide my own roles that live alongside `bedrock-ansible`. These roles provide additional services (logging, monitoring) and files (site restores, scripts, dotfiles, and now these nginx site-level includes) that we need in our environment.

Then I copy bedrock-ansible's `dev.yml` and add my own roles, and that becomes the playbook I run. That way everything provided by `bedrock-ansbile` stays "stock."

chriszarate:
Happy to code up @louim's option 1 as a built-in method if that seems like the way to go—let me know.

swalkinshaw:
@louim's solution is good but I'd do both `files` and `templates`. No need for the option as `with_fileglob` should just work.

PR #243, opened 2015-06-16
Allow for custom permissions for shared sources

davenaylor:
Current deploys ensure that shared/uploads exists but the directory is left owned by web_user in the web_group group.  The permissions set are 0755 therefore, unless the owner of hhvm/php-fpm is the web_user, uploads aren't possible.  This pull request allows for the setting of looser permissions granting group writeable shared sources.


swalkinshaw:
Makes sense to me :+1:

PR #244, opened 2015-06-16
Wrap octal permission in quotes

davenaylor:
Had to wrap the octal permission in quotes (even though it was working!)


swalkinshaw:
@davenaylor I had already made a change so see here: 9a77834

PR #246, opened 2015-06-23
Merge secure-root.yml into server.yml

fullyint:
**Summary.** Instead of having a separate playbook to run (`secure-root.yml`), just set `secure_root: true` and run `server.yml`. Consolidates `sudoers` and `github_ssh_keys` roles into the existing `users` role.

**New `secure_root` variable.** It is in `group_vars/all` for parity between environments (note: doesn't do anything for development). Just change this to `true` on the first or any subsequent run of `server.yml` and that's all you need to do

**New `admin_user` variable.** Used in `users` list and `remote_user` parameter of `server.yml`. Convenient for people who might prefer `ubuntu` user (see [AWS discussion](https://github.com/roots/bedrock-ansible/pull/58)). There's some risk people could confuse this `admin_user` with `wordpress_sites['example.com']['admin_user']` in `group_vars/development`

**Conditional remote_user.** New task in `server.yml` checks which remote user to use (e.g., `root` vs. `admin`; inspiration: [this](https://groups.google.com/forum/#!topic/ansible-project/UMtmnBbJ6vk) and [this](http://stackoverflow.com/a/20546428)). You might be concerned that if/when you add a new host IP to `web` in your inventory, the new host wouldn't yet have root secure but the preexisting hosts would. In that case you'd need to use [--limit](http://docs.ansible.com/glossary.html#limit-groups) ([see also](http://stackoverflow.com/a/18199029)) to first run the playbook on just the new host so that is has root locked down like the other hosts (or maybe the [blacklist approach](https://groups.google.com/forum/#!topic/ansible-project/O1dpHTSGaL8) would be more convenient to your situation). The logic for the playbook's remote_user parameter tries to get the right user for you -- just checks whether it can ssh as root to the first host in the web group. As a different edge case, you can also [handle different machines needing different user accounts or ports](http://docs.ansible.com/faq.html#how-do-i-handle-different-machines-needing-different-user-accounts-or-ports-to-log-in-with)

**New `users` variable; SSH keys.** This PR is partly followup to discussion at https://discourse.roots.io/t/configure-domain-for-github-ssh-keys-to-point-at-a-gitlab-server/4066.  The new `users` variable subsumes the old `sudoers` and `github_ssh_keys` variables. I put it in `group_vars/all` because probably most people will want users consistent across staging and production. I left the entire list present and uncommented (except github keys) because I don't think any of it can hurt, and will be most likely to achieve successful ansible runs when people change `secure_root: true` (and do nothing else). By default it will attempt to add a key from `~/.ssh/id_rsa.pub` (like [it did before](https://github.com/roots/bedrock-ansible/blob/9a77834888d361988d8fcc3517494825efc87b10/roles/sudoers/defaults/main.yml#L3)). Note that since ansible 1.9 the [`key` parameter](http://docs.ansible.com/authorized_key_module.html) can accept a url, so we no longer need a separate task to fetch and store keys in `/tmp`before the ansible task that loads them.

**Ansible requirement.** This Bumps Ansible requirement to 1.9+. See 'SSH keys' section above for why.

**Quotes on path variables.** I removed quotes on `local_path` (in `wordpress_sites` lists) because our other vars with paths don't use quotes.

**Idempotency.** All tasks are idempotent except "Setup users" task, due to Ansible bug @swalkinshaw already noted [here](https://github.com/roots/bedrock-ansible/issues/218#issuecomment-104503820)

**Sudo issues.** I just left `sudo: yes` in `server.yml` always, after finding out that because it is [not currently templatable](https://github.com/ansible/ansible/pull/10075), I can't set it's value with conditional logic. Maybe in Ansible v2. As for `--ask-become-pass` (formerly `--ask-sudo-pass`, [ref](http://docs.ansible.com/become.html#new-ansible-variables)), you won't even need it the first time you run `server.yml` with `secure_root: true` because that run will still be as `root` (but you'll need it for subsequent runs). Did you want programmatic conditional `--ask-become-pass`? It is [not currently possible](https://github.com/ansible/ansible/issues/7742), but people can make ansible always ask for sudo pass by adding [`ask_sudo_pass=True`](http://docs.ansible.com/intro_configuration.html#ask-sudo-pass) to `ansible.cfg` (but that would just get annoying when running `dev.yml` and `deploy.yml` which don't need the sudo pass).

**Wikis**. If/when this PR is accepted, I have notes ready for wikis (ssh keys and security)


fullyint:
After some discussion with @swalkinshaw I see that it will be better to process these changes as separate pull requests, one that combines the `github_ssh_keys` role with the `users` role, and another that handles the potential merge of the `secure-root.yml` playbook into the `server.yml` playbook. So, I'll close now and prepare the revised PRs.

nathanielks:
Man Phil, you write some of the best documentation. Its seriously a pleasure to read what you write. Bravo, sir, bravo. 



—

PR #247, opened 2015-06-23
Move ssh key handling to users role

fullyint:
This is the subcomponent of #246 that combines the `github_ssh_keys` role with the `users` role, stemming from https://discourse.roots.io/t/configure-domain-for-github-ssh-keys-to-point-at-a-gitlab-server/4066

- Bumps Ansible requirement to 1.9+, to take advantage of revised [`key` parameter](http://docs.ansible.com/authorized_key_module.html) that accepts a url.
- Loads `vars/sudoer_passwords.yml` so `users` can be in the sudo group with a password, if desired.
- Removes quotes on `local_path` (in `wordpress_sites`) because other vars for paths don't use quotes.


roles/users/tasks/main.yml
```diff
@@ -1,12 +1,13 @@
 ---
-- name: Setup web user
-  user: name="{{ web_user }}"
-        group="{{ web_group }}"
+- name: Setup users
+  user: name="{{ item.name }}"
+        group="{{ item.groups[0] }}"
+        groups="{{ item.groups | join(',') }}"
+        password="{{ sudoer_passwords[ item.name ] | default(None) }}"
```
swalkinshaw: `sudoer_passwords[item.name]`

fullyint: Was copy-pasted from [`sudoers` role](https://github.com/roots/bedrock-ansible/blob/ef70af82afad5c48f921a89b19d5a17022ea6b7d/roles/sudoers/tasks/main.yml#L16), but I'll leave that one alone instead of changing it too.

swalkinshaw:
:+1: much better and clearler

PR #253, opened 2015-06-29
Protect against Logjam attack. Fixes #252

### LINKED ISSUE #252: SSL Logjam attack. Generate strong Diffie-Hellman group
We should use a strong Diffie-Hellman: https://weakdh.org/

What should be done: https://weakdh.org/sysadmin.html, We currently enforce ciphers, would shoud just generate a strong Diffie-Hellman group.

Reference SSL labs score for a website deployed using the current configuration: https://www.ssllabs.com/ssltest/analyze.html?d=strategiemaritime.gouv.qc.ca

I will provide a PR

louim:
Generate a 2048 bit prime and enforce its usage.


swalkinshaw:
Awesome :+1:. Think that "Create SSL directory" is better in the Nginx role anyway.

fullyint:
Thank you for the Diffie-Hellman! Any interest in a `when:` conditional like this?
```
when: true in [{% for key, value in wordpress_sites.iteritems() %}{{ value.ssl.enabled }},{% endfor %}]
```
While developing Trellis, it's an all the time operation (destroying and recreating servers constantly). I groan when provisioning churns on the Diffie-Hellman task because I forget to comment it out.

louim:
I see your point. I wasn't aware a condition like that worked. It must be because I have an SSD, or maybe that I just don't check the vagrant when it is building, that I never noticed it take much time. 

Out of curiosity, how many time does generating the group take on your machine?

fullyint:
Without doing formal tests, my recollection is that the Diffie-Hellman task varies widely in the time it takes, from 5 seconds to 5 minutes or more. I think others notice the slowness too. I am on a rather old system however, so maybe it's affecting me more: OSX 10.9.5, 3.6 GHz Intel Core 2 Duo (64 bit), 8 GB RAM, SSD (getting low on space).

retlehs:
image

swalkinshaw:
@louim I also suggested we just don't run it on development. Simplifies the logic.

louim:
@swalkinshaw that would be the better option I think.

austinpray:
We shouldn't hardcode the development exclusion. Just make it a config option that perhaps we set to "off" by default in the development vars.

fullyint:
No one needs to engage with me on this, but I don't understand why we'd limit the conditional to development. In standard use on staging/production, users should have SSL enabled and the Diffie should run. But some users won't have SSL enabled. Additionally, the process of testing non-ssl features in staging/production could be faster if the Diffie were conditional.

It's normal to avoid running slow processes when they aren't needed. Of course, it's not worth it if the conditional logic involved crosses the line into being too complex, risking development confusion or errors. Where you draw that "line" is probably a subjective matter. 

This conditional doesn't strike me as too complex, nor as a hack in purpose or implementation:
`when: true in [dynamically-created list of booleans]`

I think I'd even prefer the combination:
`when: "'development' not in group_names and true in [dynamically-created list of booleans]"`
It gets closer to crossing the line into risky complexity, but doesn't cross it for me.
I'd rather use this so that our code does all the things we want, then just put a comment next to it:
`# when not development and at least one site has ssl.enabled: true`

The above conditional could be altered to accommodate @austinpray's suggestion for a config option instead of hardcoding the development exclusion.

Perhaps the strongest argument against it is "The conditional would only serve edge cases so it's not worth it. Users should always use SSL. People testing Trellis can just comment out the Diffie task instead expecting complex conditionals built in just for dev purposes."

Again, no pressure to respond, but if one of you more-seasoned devs wants an eager pupil, I'm ready.

swalkinshaw:
@fullyint let's just do with `when: true in [dynamically-created list of booleans]`

PR #254, opened 2015-06-29
Update env.j2

nathanielks:
Jinja converts boolean values to "True" and "False," which php doesn't like. This converts those values to values php will interpret as booleans.

### DIFF
--- a/roles/wordpress-install/templates/env.j2
+++ b/roles/wordpress-install/templates/env.j2
@@ -1,3 +1,8 @@
 {% for key, value in item.value.env.iteritems() %}
+{% if value == 'True' %}
+{% set value = 1 %}
+{% elif value == 'False' %}
+{% set value = 0 %}
+{% endif %}
 {{ key | upper }}={{ value }}
 {% endfor %}


louim:
Testing:
`development.yml`

```
true_text: true
false_text: false
true_numeric: 1
false_numeric: 0
```
Give me (`.env`):
```
FALSE_TEXT=False
TRUE_TEXT=True
TRUE_NUMERIC=1
FALSE_NUMERIC=0
```
Which are loaded in PHP (var_dumping) with dotenv 1.x:
```
string(4) "True"
string(5) "False"
string(1) "1"
string(0) ""
```
With dotenv 2.x
```
string(4) "True"
string(5) "False"
string(1) "1"
string(1) "0"
```

I don't see problem on my end with the way the variable are loaded, unless I'm mistaken. But I think I will prepare a PR for Bedrock to use Dotenv v2.

Ansible version:
```
ansible 1.9.0.1
  configured module search path = None
```

swalkinshaw:
`getenv` always returns a string anyway. So you'd need to check the variable contents if its `True` or `1`. Is that correct @nathanielks?

nathanielks:
Hmmmm. The method I'm using at the moment is actually setting values in an ini using their env syntax:


```

some_value = ${SOME_ENV}

```




The value derived from that, if its True, is a string. If its 1, its interpolated as a Boolean. 





—

swalkinshaw:
@nathanielks but is that because of the context you're using it in? Since PHP does a lot of auto coercion.

Because http://php.net/manual/en/function.getenv.php says `string`

swalkinshaw:
Interesting but doesn't matter right now:

> template code now retains types for bools and Numbers instead of turning them into strings If you need the old behaviour, quote the value and it will get passed around as a string

That's for Ansible 2.0 in development

swalkinshaw:
@nathanielks I did some tests:

```yaml
env:
  foo1: true
  foo2: True
  foo3: 1
  foo4: 0
  foo5: 'true'
```

```php
var_dump(getenv('FOO1'));
var_dump(getenv('FOO2'));
var_dump(getenv('FOO3'));
var_dump(getenv('FOO4'));
var_dump(getenv('FOO5'));
var_dump(getenv('FOO6'));
```

Results in:

```
string(4) "True"
string(4) "True"
string(1) "1"
string(0) ""
string(4) "true"
string(4) "True"
```

swalkinshaw:
Going to close this since tests that both myself and @louim did show that PHP always loads env vars as strings.

nathanielks:
:+1:

Issue #251, opened 2015-06-29
PHPUnit


getdave:
Would it be possible to provision the VM with a copy of PHPUnit? As you need to be SSH'd into the box to run WPCLI it would help with the WordPress Unit Test scaffolding.

swalkinshaw:
Of course. It would be nice to have and should be a pretty simple role to add. Would you be interested in doing this?

getdave:
That's great. Interested yes but not quite up to speed on Ansible. Could you outline the approach you would take so I can absorb the workflow?

louim:
Shouldn't this just be added as a dev dependency of the project you want to run test on? I may be wrong since I never used phpunit. I would do it like this: `composer require-dev "phpunit/phpunit=4.7.*"`. What is the benefit of installing it globally?

swalkinshaw:
:+1: to what @louim said

austinpray:
You definitely want to install it locally. The phpunit binary should be available in /vendor/bin

kalenjohnson:
The benefit of installing globally is not needing phpunit required in your composer.json file, although if you have tests for each project it doesn't hurt. I would also say it's much easier to type `phpunit` than `./vendor/bin/phpunit` every time :joy:

nathanielks:
PATH could also be updated to include `./vendor/bin`.

—

kalenjohnson:
So it'll always work where your `composer.json` file is which has phpunit installed? I suppose that's an option too

nathanielks:
Yup! That way wherever you’re at you’re using that projects dependencies and not what’s available globally.


—

austinpray:
>The benefit of installing globally is not needing phpunit required in your composer.json file

Right, but phpunit can have breaking changes that affect your tests. You want to lock down the version just like any other package installed via composer.

nathanielks:
> You want to lock down the version just like any other package installed via composer.

This.

mAAdhaTTah:
> PATH could also be updated to include `./vendor/bin`.

+1 for including this as an update. Might be a good start for someone new to Ansible (like me :) ).

getdave:
Agree with keeping it as a local dependency and updating `PATH` to include `./vendor/bin`. Makes more sense. 

Ansible pros - what's the best method for updating `PATH`?

louim:
[...]

PR #256, opened 2015-07-02
Enable plugin unit testing

getdave:
A initial attempt at adding the ability to utilise the WP Unit Test library to Unit Test Plugins. Commits should outline work that has been completed. Basic work was:

* Create new Ansible role for `wordpress-tests`
* Configure Ansible to:
  * create new dir to store WP core test library
  * create test specific configuration file and store in `config` within Bedrock site
  * create a test database alongside main database
* Add PHPUnit as dependency within Composer
* Update README with basic instructions.

I'm very conscious that this is probably not up to the Roots team standards and so very keen to hear constructive feedback on how we feel this could be improved upon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,12 +152,35 @@ You can enable FastCGI caching on a per site basis. The cache is a low duration,
 
 The `secure-root.yml` playbook is provided to help secure your remote servers including better SSH security. See the Wiki for [Locking down root](https://github.com/roots/bedrock-ansible/wiki/Security#locking-down-root).
 
-## Unit Testing
+## Plugin Unit Testing
 
-You may wish to Unit Test your Plugins using the WP Core PHPUnit classes. To configure this ensure you have configured the test database correctly within the `group_vars/development` file. __This database will be dropped__ between each test so it is *critical* that you never use your main database as your test database.
+You may wish to Unit Test your Plugins using the WP Core PHPUnit classes. To configure this ensure you have configured the test database correctly within the `group_vars/development` file. 
+
+__Warning:__ *This database will be dropped* between each test so it is *critical* that you never use your main database as your test database.
 
 ### Testing a Plugin
 
-To test an individual plugin 
+To test an individual plugin we advise using WP CLI to scaffold the required files. To do this ssh into Vagrant and run:
+
+````wp scaffold plugin-tests my-plugin````
+
+...where `my-plugin` is the Plugin dir in which you wish to scaffold the test files. You should see several files generated. 
+
+We next need to edit the `tests/bootstrap.php` file, replacing...
+
+```
+$_tests_dir = getenv('WP_TESTS_DIR');
+if ( !$_tests_dir ) $_tests_dir = '/tmp/wordpress-tests-lib';
+```
+
+...with
+
+```
+$_tests_dir = '../../../../wp-tests-lib';
+```
+
+To run the Plugin's test suite cd into your plugin directory and run
+
+`../../../../vendor/bin/phpunit`
 
 
--- a/README.md
+++ b/README.md
@@ -127,6 +127,9 @@ For a complete, working example you can see our [example project](https://github
   * `db_user` - database username (required)
   * `db_password` - database password (required)
   * `db_host` - database hostname (default: `localhost`)
+  * `test_db_name` - test database name (__must__ be different to primary `db_name`)
+  * `test_db_user` - test database user (__must__ be different to primary `db_user`
+  * `test_db_password` - test database password (__must__ be different to primary `db_password`)
   * `domain_current_site` (required if multisite.enabled is `true`)
 
 ### Mail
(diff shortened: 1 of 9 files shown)


louim:
I feel like this would be nice as an Ansible Galaxy role that could just be added to the projet if one need to develop plugin or add test unit to his project. However, I believe it is a little bit too "opinionated" to fit in the standard bedrock-ansible use case. I would just set the path in bedrock-ansible like discussed in #251 and move the rest of the code to an Ansible Galaxy role. I did the same with https://github.com/louim/bedrock-site-protect 

My 2 cents.

austinpray:
I'm with @louim as far as making this into an Ansible Galaxy role. I definitely want to keep the core functionality of Trellis as general and decoupled as possible. As a maintainer I want to keep this repo focused on its primary mission: spinning up a rock solid WordPress server environments. Further setup inside this environment should be userland stuff. 

@swalkinshaw perhaps we should create a section in the README for these user contributed Galaxy addons. 

Also as a by-the-way: Curious to see how this setup compares with @johnpbloch's setup.

getdave:
Hi both

Ok that's fine. I've raised this only on encouragement from others in the Issue tracker.

Is there a general consensus that WordPress Unit Tests shouldn't be a core part of a setup for WordPress websites? 

More than happy to move to a plugin role. If that's the case then I'd suggest we definately need a recommended plugins section in the README.

Also would someone consider adding a commit to the PR which adds phpunit to PATH? Couldn't wrap my head around it yesterday for some reason.

louim:
I created a Wiki page https://github.com/roots/trellis/wiki/User-contributed-extensions so people that developed extensions could list them here. Maybe we could put a link in the readme to that page.

swalkinshaw:
@getdave thanks for this work.

I'd actually say this work itself is fine and up to our "standards". Nothing stands out as being wrong or anything. So while the implementation itself is good, it's the functionality I'm not sure about.

As a few others mentioned, this might be better as an external role. But at the same time I think there's more we should be doing to make testing easier and encourage it with Trellis since it's important.

I personally hate the default method of testing WP. It's pretty awful that we need to use SVN to pull down this other library which contains the testing library and use a completely different config file for it. 

A better method in my opinion is using something like http://giuseppe-mazzapica.github.io/BrainMonkey (see http://giuseppe-mazzapica.github.io/BrainMonkey/docs/wordpress-why-bother.html also).

Problem of course is that it doesn't test integration with WP core. You'd still need to do what this PR has.

mAAdhaTTah:
Something like WP_Mock could also do this: https://github.com/10up/wp_mock

johnpbloch:
Since @austinpray was interested esoterically in my thoughts on this, I figured I'd weigh in, albeit a bit late to the thread.

I tend to agree, personally, with @swalkinshaw regarding the use of the core testing framework to run plugin and theme unit tests. The tests may be called unit tests, and the SUT for any given test may be a "Unit", but they're really integration tests at heart. For true blue unit tests, you need to isolate your code as much as possible and both Brain Monkey and WP Mock are built for that purpose (*disclosure: I help maintain and develop WP Mock as part of my role at 10up*).

That being said, I think as much as possible we should be in the habit of not enforcing specific tools that we don't absolutely need to enforce. So whatever test runner, test helper, etc. a dev wants to use should be permitted as long as there's no real reason not to allow it. I don't think an environment management tool like trellis should be opinionated about such matters, personally. That's more in the purview of a CI server or plugin package.

My setup isn't really all that much to behold, but there's not a whole lot of intersection of concerns between deployment/environments and testing. Code that doesn't pass tests doesn't make it into master and code that's not on master doesn't get deployed. Plugin/theme tests run when I'm developing the plugin/theme, not when I'm editing the package versions in the site's `composer.json`.

Keep on keeping on with the tests, though—regardless of what anybody says about one method versus another. :fireworks:

nathanielks:
>  So whatever test runner, test helper, etc. a dev wants to use should be permitted as long as there's no real reason not to allow it.

:clap: :clap: :clap: :clap:

QWp6t:
I share the same opinion as the @10up guys. The WP Unit Tests are for testing WP core during core development. If you're not developing core, then it's unnecessary. There is no reason to use it for testing themes and plugins. For themes and plugins, you should mock the core components that are used by the theme/plugin.

getdave:
Ok thanks for the input everyone. So basically the consensus is

1. We don't want this in core Trellis - move to a Ansible plugin
2. We don't actually want WP Core Unit Testing FW at all
3. We want flexibility to use whatever testing FW we want to use.

Is that fair?

As a result of the above I'm not convinced this pull request is of any value beyond a proof of concept. 

Only exception might be if there's some base configuration regarding having a out of the box "test" environment config that's separate from the application config?

swalkinshaw:
@getdave summary seems correct. What I could be in favour of is automatically creating a test db in development. This could be done based on the `db_name` though and just append `_test`.

PR #257, opened 2015-07-02
Update the Vagrant file with dynamic ANSIBLE_PATH

louim:
`ANSIBLE_PATH` was relative to the folder where the command was called. This caused error when using Vagrant global commands, which can be called from anywhere. for example `vagrant halt 428b9b5` would produce this error (when in another folder):

```
There was an error loading a Vagrantfile. The file being loaded
and the error message are shown below. This is usually caused by
a syntax error.

Path: /Users/louim/repositories/bedrock-ansible-dev/bedrock-ansible/Vagrantfile
Message: ./group_vars/development file not found. Please set `ANSIBLE_PATH` in Vagrantfile
```

Using [`__dir__`](http://ruby-doc.org/core-2.0.0/Kernel.html#method-i-__dir__) we alway get the path relative to the file itself, not the current folder where the command is executed.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -6,7 +6,7 @@ ENV['VAGRANT_DEFAULT_PROVIDER'] = 'virtualbox'
 
 require 'yaml'
 
-ANSIBLE_PATH = '.' # path targeting Ansible directory (relative to Vagrantfile)
+ANSIBLE_PATH = __dir__ # path targeting Ansible directory (relative to Vagrantfile)
 
 # Set Ansible roles_path relative to Ansible directory
 ENV['ANSIBLE_ROLES_PATH'] = File.join(ANSIBLE_PATH, 'vendor', 'roles')


kalenjohnson:
:+1: haven't tested this, but I did miss being able to `vagrant up` and `vagrant halt` from whatever child directory I was in.

How would this work with using something like the roots-example-project.com project? just `__dir__ + '/site'` ?

kalenjohnson:
Whoops, I mean `__dir__ + '/ansible'`

louim:
@kalenjohnson Not exactly, it just evaluate as the full path to the folder where the `Vagrantfile` is. This doesn't change the current way this works. It just make it work correctly with Vagrant global commands. To do what you are trying to do you just have to use: `vagrant global-status` to get the id of the VM and you can run `vagrant (up | provision | halt | whathever) your_vm_id` from any folder.

kalenjohnson:
@louim actually it must be a nice side-effect, because Vagrant does traverse up folders to see if it can find a `Vagrantfile`. I just tested this and it does in fact work. Following this type of project: https://github.com/roots/roots-example-project.com

And changing https://github.com/roots/roots-example-project.com/blob/master/Vagrantfile#L6 to `ANSIBLE_PATH = __dir__ + '/ansible'`, does in fact allow you to issue vagrant commands from the `site` folder or any child folder of the project.

So.... :+1: :+1: :+1: :+1: :+1:

louim:
Oh, my bad, you are totally right, I hadn't see that the vagrantfile was out of the bedrock-ansible directory. Then yes it work like you mentioned.

swalkinshaw:
Good idea! Thanks.

PR #259, opened 2015-07-03
Add composer binary path to the default path.

louim:
This adds the composer binary path to the default $PATH variable as discussed in #256 and #251 . I didn't chose to implement a complex regex. If anyone has a better regex, (eg it allow to append before the last quote of the path), let me know. 

I did chose to append the path in the `/etc/environment` because it is the preferred place to do it (the file contain one line : the PATH definition). After looking in the `/etc/bash.bashrc` it would have been out of place in this file.

This is also implemented globally, not only for the `vagrant` user in development. I think is wouln't hurt to have the composer binary path available on production servers too.

Feedback appreciated. 
cc: @getdave if you where wondering how to do it.

### DIFF
--- a/roles/composer/tasks/main.yml
+++ b/roles/composer/tasks/main.yml
@@ -9,3 +9,10 @@
   command: composer self-update
   register: composer_update_result
   changed_when: composer_update_result.rc == 0 and 'already using composer version' not in composer_update_result.stderr
+
+- name: add Composer vendor binary path
+  lineinfile:
+    dest: /etc/environment
+    regexp: ^PATH="(((?!:./vendor/bin).)*)"
+    line: PATH="\1:./vendor/bin"
+    backrefs: yes


roles/composer/tasks/main.yml
```diff
@@ -9,3 +9,9 @@
   command: composer self-update
   register: composer_update_result
   changed_when: composer_update_result.rc == 0 and 'already using composer version' not in composer_update_result.stderr
+
+- name: add Composer vendor binary path
+  lineinfile:
+    dest: /etc/environment
+    regexp: "^PATH=.*"
+    line: PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:./vendor/bin"
```
austinpray: I don't like hardcoding all these values. Just add a line that is like `PATH="$PATH:./vendor/bin"`

swalkinshaw: :+1:

louim: Sadly, that wouldn't work, this file is not evaluated, so variables are not expanded. I'll refine the reflex so it just add at the end of the string instead of rewriting whole.

louim:
I had more time to poke around with the regex. Updated so it doesn't require keeping the full path in the role.

swalkinshaw:
Looks good. Can you do a squash first?

louim:
@swalkinshaw done!

swalkinshaw:
Thanks :+1:

Issue #126, opened 2015-01-26
Upgrade WP-CLI?


mAAdhaTTah:
I think the problem here is with the hosted box, not bedrock ansible, which comes with WP-CLI 0.15.0 installed by default. [0.18.0](http://wp-cli.org/blog/version-0.18.0.html) has `wp cli update`, which means you wouldn't have to keep it updated on the box going forward, as users can update it themselves (or we can run it during provision). Can the box be updated?

swalkinshaw:
It's outdated in many other ways as well :( I'll work on getting it updated soon and this will be part of it.

austinpray:
I thought WP CLI was installed when ansible runs? Why is it included in the box?

mAAdhaTTah:
Faster provisioning, I think? There's a couple things included by default in the box.

swalkinshaw:
Not valid anymore. See https://github.com/roots/trellis/pull/260.

PR #260, opened 2015-07-06
Change base box to stock Ubuntu 14.04

kalenjohnson:
We've been enjoying quick vagrant provisions using the `roots/bedrock` box, however it does have it's downfalls

1) Maintenance, a few things need to be manually updated like wp-cli
2) Dev/Prod parity - while this may not be a huge difference, using a stock Ubuntu box will ensure that all the same steps are being taken while provisioning both the Vagrant box and staging/production servers
3) `roots/bedrock` is the wrong name, and this is a big problem for me :trollface: 

So while initial `vagrant up` will take an extra few minutes, the trade-offs should be more than worth it.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -23,7 +23,7 @@ end
 Vagrant.require_version '>= 1.5.1'
 
 Vagrant.configure('2') do |config|
-  config.vm.box = 'roots/bedrock'
+  config.vm.box = 'ubuntu/trusty64'
   config.ssh.forward_agent = true
 
   # Required for NFS to work, pick any local IP


kalenjohnson:
I'll also note, this should take care of these issues:

https://github.com/roots/trellis/issues/126
https://github.com/roots/trellis/issues/73

nathanielks:
I concur. I actually do this on my own anyway =X


—

kalenjohnson:
I actually proposed this for you @nathanielks , I want to save you as much time as possible on all future projects :+1:

nathanielks:
I can always count on @kalenjohnson to have my back :+1:

nathanielks:
Not sure about that @austinpray fellow though...

austinpray:
@nathanielks Yeah I do this as well.

mAAdhaTTah:
Yeah, I haven't used the base box in a while because of the WP-CLI issue.

Issue #262, opened 2015-07-07
Dynamic ANSIBLE_PATH breaks Windows compatibility


jchapron:
It seems that #257 breaks windows compatibility. I tracked down the issue to this code in windows.sh : 

```ruby
if [ ! -d /vagrant/${ANSIBLE_PATH}/vendor ]; then
  echo "Running Ansible Galaxy install"
  ansible-galaxy install -r /vagrant/${ANSIBLE_PATH}/requirements.yml -p /vagrant/${ANSIBLE_PATH}/vendor/roles
fi
```

This code used to work with ANSIBLE_PATH being "." but breaks now that it's the absolute path to the file. You get something like "/vagrant/D:/development/ansible/vendor".

I'm not sure how to fix it without just rollbacking to the old value.
Thanks

louim:
Woops, that's my fix.

@kalenjohnson you have pioneered most of the windows compatibility, is there a reason the Ansible need to use `ANSIBLE_PATH` inside the VM?

kalenjohnson:
Hmm... yeah, I see how that's an issue now. I unfortunately don't work in Windows so I forgot about that.

The only reason we're passing the Ansible path to the windows script is to account for the example project modified directory structure.

So, it seems to me that perhaps we should simply have a bash variable set up in `windows.sh`, which points to `/vagrant`, if the Vagrantfile is in the topmost folder with the Ansible playbooks, then fine, it works out of the box. If you are following the project example, having the playbooks in an `ansible` folder and the `Vagrantfile` in the root directory, you would have to modify the variable on `windows.sh` to point to `/vagrant/ansible`.

PR #264, opened 2015-07-07
Windows compatibility fix

bezko:
*  ANSIBLE_PATH should be '' instead of \__dir__ or the windows path will be used inside  the vagrant box
* use --force-yes to install ansible (fails without)

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -59,7 +59,7 @@ Vagrant.configure('2') do |config|
   if Vagrant::Util::Platform.windows?
     config.vm.provision :shell do |sh|
       sh.path = File.join(ANSIBLE_PATH, 'windows.sh')
-      sh.args = ANSIBLE_PATH
+      sh.args = ''
     end
   else
     config.vm.provision :ansible do |ansible|
--- a/windows.sh
+++ b/windows.sh
@@ -20,7 +20,7 @@ if [ ! -f /usr/bin/ansible ]; then
   echo "Updating system..."
   sudo apt-get -y update
   echo "Installing Ansible..."
-  sudo apt-get -y install ansible
+  sudo apt-get -y --force-yes install ansible
 fi
 
 if [ ! -d /vagrant/${ANSIBLE_PATH}/vendor ]; then


kalenjohnson:
I'm not sure if I'm reading this correctly, but wouldn't passing in a blank string cause all references in `windows.sh` to also error? Like https://github.com/roots/trellis/blob/master/windows.sh#L26 , wouldn't that become `/vagrant//vendor` ?

Perhaps `sh.args = ANSIBLE_PATH` should become `sh.args = '.'` like how it was before, and if Trellis is installed to a subfolder, then this should be changed to the subfolders name.

That seems better actually than what I had proposed [here](https://github.com/roots/trellis/issues/262#issuecomment-119281682)

bezko:
sh.args = '.' works too , not sure which one is less ugly /vagrant//vendor or /vagrant/./vendor

This is just a band-aid, I think windows.sh is broken , $ANSIBLE_PATH is always an absolute path  ( as defined in the Vagrantfile) so /vagrant/${ANSIBLE_PATH}/vendor doesn't make much sens

louim:
I had a look at what could be done to fix this, while retaining the feature of finding the provisioning folder when it is nested.

Completely remove the `sh.args` parameter from the `Vagrantfile`.
Replace line https://github.com/roots/trellis/blob/master/windows.sh#L8 with:
```
ANSIBLE_PATH="$(find -name 'windows.sh' -printf '%h' -quit)"
```

This will look up the tree and find the `windows.sh` file. it will then assign the parent folder path to the `ANSIBLE_PATH` variable. The `-quit` parameter is to stop the lookup after a match has been found.

I  don't have a windows machine handy to test it at the moment, @bezko, would you be interested testing   ?

kalenjohnson:
Wouldn't it be, um... down, the tree, from the Vagrantfile?

louim:
Ahah yeah I think you are right! English is not my primary language :grin:

kalenjohnson:
Haha no worries, just wanted to make sure ;)

bezko:
@louim I'm not sure I understand the goal of this, windows.sh is looking for the location of itself relative to itself, isn't that always goint to be '.'?

Anyone I've got an error : 

$ find -name 'windows.sh' -printf '%h' -quit
find: invalid predicate `-quit'

swalkinshaw:
Thanks @bezko. Went with https://github.com/roots/trellis/pull/269

PR #268, opened 2015-07-09
Conditionally copy .env into web root

### LINKED ISSUE #261: .env should be regenerated based on content
1. Set WP_HOME to example.dev
2. Provision the server
3. WP_HOME in .env is example.dev
4. Set WP_HOME to mysite.dev
5. Provision the server
6. WP_HOME is still example.dev in .env

I would expect this to be regenerated when the values change.

fullyint:
closes #261
`.env` was always being created (in `/tmp`) when `env` vars changed in `group_vars/development`. However, `.env` was only being copied into web root if there wasn't already an `.env` file there.

With these changes, `.env` will always be copied into place if a new `.env` is created, any time `env` vars are changed in `group_vars/development`. Still idempotent.

---------
Edit: The next several comments were reacting to these original PR changes (now out-of-date):
```diff
 - name: Create .env file
   template: src="env.j2"
             dest="/tmp/{{ item.key }}.env"
             owner="{{ web_user }}"
             group="{{ web_group }}"
   with_dict: wordpress_sites
+  register: env
 
 - name: Copy .env file into web root
-  command: creates="{{ www_root }}/{{ item.key }}/current/.env" cp /tmp/{{ item.key }}.env {{ www_root }}/{{ item.key }}/current/.env
+  command: cp /tmp/{{ item.key }}.env {{ www_root }}/{{ item.key }}/current/.env
   with_dict: wordpress_sites
+  when: env.changed
```

### DIFF
--- a/roles/wordpress-install/tasks/main.yml
+++ b/roles/wordpress-install/tasks/main.yml
@@ -9,8 +9,10 @@
   with_dict: wordpress_sites
 
 - name: Copy .env file into web root
-  command: creates="{{ www_root }}/{{ item.key }}/current/.env" cp /tmp/{{ item.key }}.env {{ www_root }}/{{ item.key }}/current/.env
+  command: rsync -ac --info=NAME /tmp/{{ item.key }}.env {{ www_root }}/{{ item.key }}/current/.env
   with_dict: wordpress_sites
+  register: env
+  changed_when: env.stdout == "{{ item.key }}.env"
 
 - name: Install Dependencies with Composer
   command: composer install


louim:
Why is the `.env` file created in in `/tmp` instead of being copied directly? It seems to have been done with the Windows compatibility patch: https://github.com/fullyint/bedrock-ansible/commit/3d328563858518e75ee0ce60c6f34237f0c5bae0#diff-6dbc78b8392771e0501227e09777c762L6

@kalenjohnson do you remember what was the rationale behind this change?

fullyint:
There's something different in how vagrant handles the filesystem on Windows. The "Create .env file" task was failing on Windows hosts. Here's one [example discussion](https://discourse.roots.io/t/provisioning-failure-all-hosts-have-already-failed/3574/6?u=fullyint).

louim:
Wow, thanks, I didn't know that!

austinpray:
```bash
mkdir tmp
cd tmp
git clone git@github.com:roots/trellis.git
git clone git@github.com:roots/bedrock.git site
cd trellis
ansible-galaxy install -r requirements.yml
vagrant up
cat ../site/.env
```

gives the following in .env

```
WP_ENV=development
DB_USER=example_dbuser
DB_PASSWORD=example_dbpassword
WP_SITEURL=http://example.dev/wp
DB_NAME=example_dev
WP_HOME=http://example.dev
```

```bash
perl -pi -e 's/example\.dev/myexample.dev/g' group_vars/development
vagrant provision
cat ../site/.env
```

gives the following in .env

```
WP_ENV=development
DB_USER=example_dbuser
DB_PASSWORD=example_dbpassword
WP_SITEURL=http://example.dev/wp
DB_NAME=example_dev
WP_HOME=http://example.dev
```

uhoh! Not working! Let's try @fullyint's

```
git reset --hard
git pull https://github.com/fullyint/bedrock-ansible.git env
perl -pi -e 's/example\.dev/myexample.dev/g' group_vars/development
vagrant provision
cat ../site/.env
```

this gives the following in .env

```
WP_ENV=development
DB_USER=example_dbuser
DB_PASSWORD=example_dbpassword
WP_SITEURL=http://example.dev/wp
DB_NAME=example_dev
WP_HOME=http://example.dev
```

Not working :(

fullyint:
@austinpray and science have prevailed, transforming this PR from "seems right," to "yeah, that's it."

**Debrief.** The above test results mystified me till I caught my mistaken assumption that the template module would check the `env` vars against the `.env` file in the web root. But of course, the template module only checks the `env` vars against the file it creates, which is actually in `/tmp`. 

- Tests above start by provisioning: **create** original `.env` file (correct `env` values), **copy** it to web root
- Change `env` values, reprovision: **create** works, but **copy** never runs -- the original problem to solve
- Switch to original PR code, reprovision: **create** skips because `.env` in `/tmp` already matches modified `env` vars (leftover from steps above), **copy** skips too -- PR made it conditional on **create**
- Change to completely new `env` values, reprovision: fullyint code runs both the **create** and **copy**
- For science, directly edit web root `.env` so it differs from `env` vars, reprovision: **create** doesn't run because `env` vars still match the `.env` file in `/tmp`, and thus **copy** doesn't run. Problem persists.

**Moving on.** I revised the PR to run **copy** any time the `env` vars differ from the `.env` file in the web root. The tasks will continue to ensure that `/tmp` always has a `.env` file matching the `env` vars (**creating** one if necessary), and now will always **copy** it to the web root if it differs from what is in the web root.

In the beginning, @swalkinshaw casually said: "could do a local rsync," and that's where I ended up. I switched the **copy** task to use rsync, but it still uses the command module because I couldn't manage to adapt the synchronize or copy modules to have both src and dest on the remote server. The command module always runs, but the task remains idempotent, only `changed_when` the file is actually synced.

swalkinshaw:
:100: good work @fullyint

austinpray:
:100:

PR #269, opened 2015-07-09
Fix windows provisioning path problem

### LINKED ISSUE #262: Dynamic ANSIBLE_PATH breaks Windows compatibility

louim:
Fixes windows provisioning, while retaining the variable nesting of the provisioning folder relative to the Vagrantfile

I successfully tested it with standard Trellis, and the example project where the Vagrantfile is one level up. @bezko if you want to test this out, it should be working as intended. 

This would fix #262 and render #264 obsolete. More testing is appreciated, as I normally don't develop on Windows. Since the fix is related to the linux VM, Windows shouldn't have anything to do with it.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -59,7 +59,6 @@ Vagrant.configure('2') do |config|
   if Vagrant::Util::Platform.windows?
     config.vm.provision :shell do |sh|
       sh.path = File.join(ANSIBLE_PATH, 'windows.sh')
-      sh.args = ANSIBLE_PATH
     end
   else
     config.vm.provision :ansible do |ansible|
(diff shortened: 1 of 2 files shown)


austinpray:
baller

swalkinshaw:
Thanks

Issue #226, opened 2015-05-25
TASK: [deploy | Copy project folders] creates duplicates


metaarts:
During the Deploy process for remote servers (mentioned below) the source folder is copied to the destination folder on the selfsame location. This produced every time a duplicate.

```
- name: Copy project folders
  command: cp -rp {{ deploy_helper.current_path }}/{{ item.item }} {{ deploy_helper.new_release_path }}/{{ item.item }}
  with_items: project_folder_paths.results
  when: item.stat.exists == True
```

swalkinshaw:
Those are actually different paths:

`cp -rp {{ deploy_helper.current_path }}/{{ item.item }} {{ deploy_helper.new_release_path }}/{{ item.item }}`

Note `deploy_helper.current_path` and `deploy_helper.new_release_path`. "current" in this **during** the deploy is the last deploy. So it's coping the folders from the last/current release to the new release.

metaarts:
In my case it's coping the folders from the current release into the existent folder to the new release.

example:
```
cp path_to_current/vendor path_to_new/vendor
```
creates `path_to_new/vender/vendor/`

swalkinshaw:
Ah that's different. I'll look into it.

swalkinshaw:
@metaarts I can't duplicate this. Tried deploys on a test VM and on roots.io. You're using Ubuntu 14.04 I assume?

swalkinshaw:
Closing since there's no update

johnvh:
I'm seeing this too. I've got lots of nested vendor directories, 1 for every deploy I've made it seems:

```
root@tw-wp-prod-web-1:/srv/www/thrivewire/current/vendor# pwd
/srv/www/thrivewire/current/vendor
root@tw-wp-prod-web-1:/srv/www/thrivewire/current/vendor# find . -iname vendor -type d
./vendor
./vendor/vendor
./vendor/vendor/vendor
./vendor/vendor/vendor/vendor
./vendor/vendor/vendor/vendor/vendor
./vendor/vendor/vendor/vendor/vendor/vendor
./vendor/vendor/vendor/vendor/vendor/vendor/vendor
./vendor/vendor/vendor/vendor/vendor/vendor/vendor/vendor
./vendor/vendor/vendor/vendor/vendor/vendor/vendor/vendor/vendor
./vendor/vendor/vendor/vendor/vendor/vendor/vendor/vendor/vendor/vendor
[...]

PR #270, opened 2015-07-10
Stop recursive copying of vendor

austinpray:
fixes https://github.com/roots/trellis/issues/226

- [x] test on existing trellis install
- [ ] test on blank new DO server

## Happening right now:

vendor/vendor/vendor/vendor/

## What should happen:

vendor/


louim:
:+1:

PR #278, opened 2015-07-18
Fix PHP error handling

swalkinshaw:
These settings in our PHP-FPM conf help us achieve our intended goal of PHP error logging:

* development: log and display all errors
* production: log errors only

Setting `log_errors` is the key here.

/cc @nathanielks @austinpray


nathanielks:
Woohoo!

Issue #277, opened 2015-07-17
SSL: Option to generate snakeoil cert for non-production environments


austinpray:
## Problem

If your production environment has SSL, you should have SSL on your development and staging as well. Why? Because stuff like this will slip into production:

![screenshot 2015-07-17 11 06 58](https://cloud.githubusercontent.com/assets/2192970/8751678/2f76e556-2c74-11e5-8e18-ecddd09ba2ed.png)
![screenshot 2015-07-17 11 07 15](https://cloud.githubusercontent.com/assets/2192970/8751677/2f74a174-2c74-11e5-9c26-839715add2cb.png)

You should be catching errors like this in development not via a user submitting a ticket. You should not be deploying to staging and production crossing your fingers and hoping it works.

So why aren't you using SSL on dev? Probably because it's too much trouble.

## Solution

```diff
* `ssl` - enable SSL and set paths
  * `enabled` - `true` or `false` (required, set to `false`)
+  * `snakeoil` - `true` or `false` (default: `false`)
  * `key` - local relative path to private key
  * `cert` - local relative path to certificate
```

When `snakeoil` is set to `true` the playbooks will generate a snakeoil certificate and hand it over to nginx. 

### Alternative solutions

I understand that you could have the user generate a snakeoil cert on the host machine and just populate `key` and `cert` properties. But that doesn't make the repo very cloneable. Having to instruct everyone on the team to generate a cert in this particular way and put it in a exactly this folder with a particular name is too much trouble. Plus this probably is DOA for windoze users.

## Implementation

Could be as simple as:

```
apt-get install ssl-cert
```

And then we have this available to us:

```
SSLEngine on
SSLCertificateFile    /etc/ssl/certs/ssl-cert-snakeoil.pem
SSLCertificateKeyFile /etc/ssl/private/ssl-cert-snakeoil.key
```

This is what I have been doing manually, mind as well build it into a playbook.

## Snake Oil?

A "snake oil" cert might be not be the correct term, but that's the term I've always heard being used. Maybe you guys can advise me. 

## Conclusion

We need to make SSL dev by default as easy as possible because https://blog.mozilla.org/security/2015/04/30/deprecating-non-secure-http/ is looming. A+ SSL is a huge feature for trellis, we want to be making it as accessible as possible.

mAAdhaTTah:
I think the term is "self-signed certificate." Otherwise, :+1:

louim:
[...]

PR #279, opened 2015-07-19
Allow auto-generation of self signed SSL certificate.

louim:
As discussed in #277.

TLDR: When the SSL "enabled" key is set to true while "key" and "cert" are not provided, a self-signed certificate will be generated and used.

### How I've done it
1. First get all the already generated self-signed certificates, so we can prevent recreating those that already exists. Certs are named like their `wordress_sites` key instead of the domain to prevent recreating new files each time the domain is changed and to stay consistent with the folder naming conventions.
1. Get the domain name of the certs gathered in step 1.
1. Generate new SSL cert and key for each `wordress_sites` key "primary" domain  that is not in the list of domains gathered on step 2 and for which SSL is enabled
1. Assign generated cert and key in nginx config for the site.

### A word of warning
I encountered some difficulties while preparing this PR. The main one is that if we generate self signed certificates, we need to update them if / when the domain change (I use the first domain provided in the list if the user provide more than one). We also need to take in account that the user can have multiple sites. Handling of `wordress_sites` , which is a `dict` while trying to fetch all generated certificates files (to prevent recreating them on each provisioning) proved to be a hard task. Running `with_items` or  `with_dict` while `register`ing the output will alway return an `array` of results, and I didn't find an efficient way to make it correlate to the `wordress_sites` dict.

I think It came out OK in the end, but if you have a more elegant solution, or a more efficient way of doing this, I'm all ears!

### The one problem
There is only one situation which is unaccounted for in the current situation: if the user has multiples sites configured on the same environment (aka multiples entries in `wordress_sites`) and decide to swap the `site_hosts` between two sites in the same provisioning run, the certs [won't change](https://github.com/roots/trellis/compare/roots:master...louim:add-snake-oil-certs?expand=1#diff-ff830fc010a506c1a47b7aba05cd8bb7R23). 
Example: 

* site1 -> *example.dev*, site2 -> **mysite.dev**

changed to:

* site1 -> **mysite.dev**, site2 -> *example.dev*

I think it's an  edge-case incredibly unlikely to happen, but I wanted to have your opinions on it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ Before using Trellis, you must configure your WordPress sites. The `group_vars` 
 * `repo` - URL of the Git repo of your Bedrock project (required, used when deploying)
 * `branch` - the branch name, tag name, or commit SHA1 you want to deploy (default: `master`)
 * `ssl` - enable SSL and set paths
-  * `enabled` - `true` or `false` (required, set to `false`)
+  * `enabled` - `true` or `false` (required, set to `false`. Set to `true` without the `key` and `cert` options [to generate a *self-signed* certificate](https://github.com/roots/trellis/wiki/SSL) )
   * `key` - local relative path to private key
   * `cert` - local relative path to certificate
 * `site_install` - whether to install WordPress or not (*development* only, required)
@@ -126,7 +126,7 @@ Outgoing mail is handled by sSMTP. Configure credentials in `group_vars/all`.  S
 
 ## SSL
 
-Full SSL support is available for your WordPress sites. Our HTTPS implementation has all the best practices for performance and security. (Note: default configuration is HTTPS **only**.) See the [SSL wiki](https://github.com/roots/trellis/wiki/SSL).
+Full SSL support is available for your WordPress sites. Trellis will also *auto-generate* self-signed certificates for development purposes. Our HTTPS implementation has all the best practices for performance and security. (Note: default configuration is HTTPS **only**.) See the [SSL wiki](https://github.com/roots/trellis/wiki/SSL).
 
 ## Caching
 
--- a/roles/wordpress-setup/tasks/self-signed-certificate.yml
+++ b/roles/wordpress-setup/tasks/self-signed-certificate.yml
@@ -1,0 +1,28 @@
+---
+- name: Get existing self-signed certificates
+  shell: find . -name "*_self_signed.pem"
+  args:
+    chdir: /etc/nginx/ssl/
+  register: self_signed_certs
+  changed_when: false
+
+- name: Get self-signed certificates domains
+  shell: openssl x509 -noout -subject -in {{ item | quote }} | sed -n "/^subject/s/^.*CN=//p"
+  register: self_signed_domains
+  args:
+    chdir: /etc/nginx/ssl/
+  with_items: self_signed_certs.stdout_lines
+  changed_when: false
+
+- name: Generate self-signed certificates
+  shell: >
+    openssl req -subj "/CN={{ item.value.site_hosts | first }}" -new
+    -newkey rsa:2048 -days 3650 -nodes -x509 -sha256
+    -keyout {{ item.key }}_self_signed.key -out {{ item.key }}_self_signed.pem
+  args:
+    chdir: /etc/nginx/ssl/
+  with_dict: wordpress_sites
+  when: >
+    item.value.ssl.enabled and item.value.site_hosts | first
+    not in self_signed_domains.results | default([]) | map(attribute='stdout') | list
+  notify: reload nginx
--- a/roles/wordpress-setup/tasks/main.yml
+++ b/roles/wordpress-setup/tasks/main.yml
@@ -1,5 +1,6 @@
 ---
 - include: database.yml
+- include: self-signed-certificate.yml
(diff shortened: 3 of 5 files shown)


swalkinshaw:
These certs can be "wildcard" ones and support multiple domains. Common name should be the "main" domain (first one I guess) and then the others can be set under Alternate Names.

Example: http://blog.endpoint.com/2014/10/openssl-csr-with-alternative-names-one.html

louim:
@swalkinshaw I don't know if it's really needed. Because the certificate is self signed, it will always show as not valid in browsers. So I don't think it would add much to include the other domains. I think it's more to help debug the https problems in the code like assets being loaded on http or other commons problems than to provide real security (because the certs are self-signed anyway).

austinpray:
>I think it's more to help debug the https problems in the code like assets being loaded on http or other commons problems than to provide real security (because the certs are self-signed anyway).

I risk being pedantic here but this only half correct. A self-signed cert will absolutely provide increased security, just not trust. For instance: say you have a staging server with a self-signed cert. If you try and login to that site in a coffee shop with people snooping the traffic will not be able to see your traffic. 

The crypto is exactly the same as a CA cert. However the chain of trust is not there. But since it's only for development and staging servers, you and your team can just know to trust it.

README.md
```diff
@@ -126,7 +126,7 @@ Outgoing mail is handled by sSMTP. Configure credentials in `group_vars/all`.  S
 
 ## SSL
 
-Full SSL support is available for your WordPress sites. Our HTTPS implementation has all the best practices for performance and security. (Note: default configuration is HTTPS **only**.) See the [SSL wiki](https://github.com/roots/trellis/wiki/SSL).
+Full SSL support is available for your WordPress sites. You can also auto-generate *self-signed* certificates for development purpose. Our HTTPS implementation has all the best practices for performance and security. (Note: default configuration is HTTPS **only**.) See the [SSL wiki](https://github.com/roots/trellis/wiki/SSL).
```
austinpray: >You can also auto-generate *self-signed* certificates for development purpose.

Trellis will also auto-generate *self-signed* certificates for development purposes.

austinpray:
In casual testing locally:

```
# set ssl enabled: true
vagrant destroy -f
vagant up
```

Yields: 

```
TASK: [wordpress-setup | Get existing self-signed certificates] ***************
failed: [default] => {"changed": false, "cmd": "ls *_self_signed.pem", "delta": "0:00:00.003916", "end": "2015-07-21 04:55:12.301727", "rc": 2, "start": "2015-07-21 04:55:12.297811", "stdout_lines": [], "warnings": []}
stderr: ls: cannot access *_self_signed.pem: No such file or directory

```

Perhaps we have to ensure the `/etc/nginx/ssl/` directory exists before we make the check.

louim:
Thanks for test @austinpray! I know what the problem is, I'll fix is asap. :smile:

louim:
@austinpray if you want to try it again, the errors should be fixed!

roles/wordpress-setup/tasks/self-signed-certificate.yml
```diff
@@ -0,0 +1,28 @@
...
+  shell: openssl x509 -noout -subject -in {{ item | quote }} | sed -n "/^subject/s/^.*CN=//p"
+  register: self_signed_domains
+  args:
+    chdir: /etc/nginx/ssl/
+  with_items: self_signed_certs.stdout_lines
+  changed_when: false
+
+- name: Generate self-signed certificates
+  shell: >
+    openssl req -subj "/CN={{item.value.site_hosts | first}}" -new
```
swalkinshaw: Spaces inside the `{}`

swalkinshaw:
Tested and works for me :+1:

louim:
@swalkinshaw squashed and ready to go.

austinpray:
Tested and this works great for vagrant! I tried provisioning a staging instance with it and chrome would not let me visit it at all:

screenshot 2015-07-28 11 35 54

No option to bypass the message.

I'm not super heartbroken about this due to the fact that https://letsencrypt.org/ is so close. We can just get a legit certificate for staging and non-production environments here in september.

>On June 16, 2015, the final launch schedule for the service was announced, with the first certificate expected to be issued sometime in the week of July 27, 2015, followed by a limited issuance period to test security and scalability. General availability of the service is expected to begin sometime in the week of September 14, 2015, provided everything goes as planned.

https://letsencrypt.org/2015/06/16/lets-encrypt-launch-schedule.html

So should we add a note about self-signed being only for development? We could also provide instructions on how to add the snakeoil cert to your trusted certificates but this might be too much trouble.

louim:
@austinpray This is really weird, I have staging sites protected by self-signed certs with Trellis currently and I dont get the warning. 

I think it's because roots.io is in the [hsts preload list](https://code.google.com/p/chromium/codesearch#chromium/src/net/http/transport_security_state_static.json&l=2145).

austinpray:
Ah you know what? That was it. If I try it with austinpray.com it works fine.

builtbylane:
for those stumbling in late, roots trellis now supports letsencrypt : https://roots.io/trellis/docs/ssl/#lets-encrypt

PR #280, opened 2015-07-19
Validate that at least the minimum required version of ansible is used

louim:
This will validate the version of Ansible used, to prevent errors that could happen with older version than what we require.

Il will also present the user with an easy to understand message if they need to upgrade their version. I set the version to `1.9.2` in prevision of #274 being merged in.

### DIFF
--- a/roles/common/defaults/main.yml
+++ b/roles/common/defaults/main.yml
@@ -1,0 +1,1 @@
+minimum_ansible_version: 1.9.2
--- a/roles/common/tasks/main.yml
+++ b/roles/common/tasks/main.yml
@@ -1,4 +1,11 @@
 ---
+- name: Validate Ansible version
+  assert:
+    that:
+     - "{{ ansible_version is defined }}"
+     - "{{ ansible_version.full | version_compare(minimum_ansible_version, '>=') }}"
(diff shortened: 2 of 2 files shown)


swalkinshaw:
This is great :+1: 

It would also be nice to detect beta versions since we've had a few people get errors because homebrew (I think) installs the version 2 beta by default for some reason. But I'm not sure if there's a consistent way to detect that.

louim:
@swalkinshaw `ansible_version` looks like this:
```
"ansible_version": {
            "full": "1.9.2",
            "major": 1,
            "minor": 9,
            "revision": 2,
            "string": "1.9.2\n  configured module search path = /usr/share/ansible"
        }
```
We could search the `string` for things like "devel" or "beta", but that's sketchy at best. I don't think there is a surefire way of detecting beta currenty.

swalkinshaw:
@louim shouldn't this be the very first task to run?

louim:
@swalkinshaw you are right, I was thinking that those task were required to be run for windows users, but all that is required for Ansible to run is installed in the `windows.sh`.

swalkinshaw:
This PR is so good it prevents me from using Trellis:

> Your Ansible version is too old. Trellis require at least 1.9.2. Your version is 1.9.0.1

:joy:

austinpray:
:+1: This is great. Glad 1.9.2 is finally out as well.

Issue #188, opened 2015-04-26
Deploy.sh with HHVM Fatal Error


BrandonShutter:
When you run deploy.yml with hhvm enabled, you get an error running `post_finalize_commands.`  This is due to it trying to restart `php5-fpm` without it existing.
```
TASK: [deploy | Run post_finalize_commands] *********************************** 
failed: [ipaddress] => (item=sudo service php5-fpm reload) => {"changed": true, "cmd": "sudo service php5-fpm reload", "delta": "0:00:00.016006", "end": "2015-04-26 17:10:04.735646", "item": "sudo service php5-fpm reload", "rc": 1, "start": "2015-04-26 17:10:04.719640", "warnings": []}
stderr: php5-fpm: unrecognized service

FATAL: all hosts have already failed -- aborting
```

rstormsf:
confirmed - face the same issue

rstormsf:
FIX:

https://github.com/roots/bedrock-ansible/blob/master/roles/deploy/defaults/main.yml#L85-L86

handle `if/else` if `hhvm` is true 
run 
`sudo service hhvm restart`

rstormsf:
For some reason, it just hangs there and is not ran. 
I had to manually ssh into and run
`
sudo service hhvm restart
`

swalkinshaw:
@rstormsf It's failing because of a sudoers issue. You can customize this variable: https://github.com/roots/bedrock-ansible/blob/5f97284dff3079563f5dcd811b9f27611a627135/roles/deploy/defaults/main.yml#L85-L86

```yaml
project_post_finalize_commands:
  - sudo service hhvm restart
```

And this is needed too:

```yaml
web_sudoers:
  - "/usr/sbin/service php5-fpm *"
  - "/usr/sbin/service hhvm *"
```

(or just remove php5-fpm if you aren't using it).

https://github.com/roots/bedrock-ansible/blob/5f97284dff3079563f5dcd811b9f27611a627135/group_vars/all#L17-L18

CFXd:
This is still an issue (bump)

louim:
@CFXd would you be interested doing a pull request? This should be easy to fix.

CFXd:
@louim not at this time—I only know enough to be dangerous :-/

swalkinshaw:
Fixed by https://github.com/roots/trellis/pull/378

PR #282, opened 2015-07-26
Restart HHVM or PHP accordingly

CFXd:
See https://github.com/roots/trellis/issues/188

### DIFF
--- a/roles/deploy/defaults/main.yml
+++ b/roles/deploy/defaults/main.yml
@@ -85,7 +85,8 @@ project_post_build_commands:
 # These are meant primarily for service restarts such as php5-fpm or memcached.
 project_post_finalize_commands:
   - if wp core is-installed; then wp eval 'wp_clean_themes_cache(); switch_theme(get_stylesheet());'; fi
-  - sudo service php5-fpm reload
+  - if dpkg --list | grep -Fq 'php5-fpm'; then sudo service php5-fpm reload; fi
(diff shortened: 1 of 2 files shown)


roles/deploy/defaults/main.yml
```diff
@@ -85,7 +85,8 @@ project_post_build_commands:
 # These are meant primarily for service restarts such as php5-fpm or memcached.
 project_post_finalize_commands:
   - if wp core is-installed; then wp eval 'wp_clean_themes_cache(); switch_theme(get_stylesheet());'; fi
-  - sudo service php5-fpm reload
+  - if dpkg --list | grep -Fq 'php5-fpm'; then sudo service php5-fpm reload; fi
+  - if dpkg --list | grep -Fq 'hhvm'; then sudo service hhvm restart; fi
```
louim: You could also do the same with `initctl list | grep -Fq 'hhvm'` or `php5-fpm`, which will return the status of the service (and its state). I think its a little bit cleaner too.

CFXd: Tried that cmd directly on the box itself with poor results so will leave this PR as is for now.

swalkinshaw:
@CFXd I'm trying to think of another solution to this so just leaving this open for a bit.

swalkinshaw:
Closing since this will change base on #305.

Issue #266, opened 2015-07-08
Fix Multisite


austinpray:
1. clone trellis
2. clone blank bedrock
3. configure development for multisite as per wiki instructions. For posterity here are the current multisite wiki instructions: https://gist.github.com/austinpray/b800ed3ec89987e6d784
4. vagrant up

It will error out. There are three problems I can see:

- [ ] wiki needs updating
- [x] [wp installed?](https://github.com/roots/trellis/blob/70ebd4527978ee88bfddd2d1b240f7c01ed20158/roles/wordpress-install/tasks/main.yml#L23-L30) task will always fail
- [x] No wildcard in nginx `server_name`

### wiki needs updating
I'm pretty sure you cannot add this stuff:

```
define('WP_ALLOW_MULTISITE', true);
define('MULTISITE', true);
define('SUBDOMAIN_INSTALL', true); // Set to false if using subdirectories
define('DOMAIN_CURRENT_SITE', getenv('DOMAIN_CURRENT_SITE'));
define('PATH_CURRENT_SITE', '/');
define('SITE_ID_CURRENT_SITE', 1);
define('BLOG_ID_CURRENT_SITE', 1);
```

Before running the playbook or it will cause DB errors.

### [wp installed?](https://github.com/roots/trellis/blob/70ebd4527978ee88bfddd2d1b240f7c01ed20158/roles/wordpress-install/tasks/main.yml#L23-L30) task will always fail
WordPress multisite will ALWAYS spit out garbage when trying to install. Which makes this: https://github.com/roots/trellis/blob/70ebd4527978ee88bfddd2d1b240f7c01ed20158/roles/wordpress-install/tasks/main.yml#L30 kinda flimsy. Perhaps we should just look at exit codes?

### No wildcard in nginx `server_name` for subdomains

We need to do this for subdomains:

```
server_name  example.dev  *.example.dev
```

Here is my hacked together code:

```
server_name {% for host in item.value.site_hosts %} {{ host }} {% if item.value.multisite.subdomains %} *.{{ host }} {% endif %} {% endfor %}
```

Please tell me jinja2 has some sort of prefix filter I don't know about.

chriszarate:
I'm using Trellis for a production subdomain multisite site. I use it as a dependency (via `requirements.yml`) for our own Ansible project. 

[...]

PR #283, opened 2015-07-29
Fix multisite install

alan-c:
Regarding #266.

### wp installed?
WP_CLI outputs a variety of information to STDERR and/or STDOUT depending on the scenario that it is working through. That output, while possibly being helpful for debugging purposes, is not directly useful to the two install plays that follow. Therefore, after toying around with ideas for expanded conditionals or group_by usage I simplified and chose to set `failed_when: False`.

FWIW, this play appears redundant and may be able to be removed/replaced entirely. WP-CLI appears to call `is_blog_installed()` at the beginning of both its `install` function and `multisite_install` function - see lines 433-434, 502-503 both leading to lines 560-563 in [this file](https://github.com/wp-cli/wp-cli/blob/master/php/commands/core.php). As it is, keeping this play may open up some possibilities for custom modifications/expansion based on individual user needs.

I also addressed the HTTP_HOST issues which were cropping up in the process as well.

### wildcard handling for server_name
I used @austinpray's code. It works.

### how to/the wiki
The procedures in the wiki basically work. Specifically the steps are:
1. Manually add the multisite constants to the `config/application.php` file. If they are not added and Bedrock is in use, WP-CLI will stick the constants at the end of the `wp-config.php` file, which is unworkable.
2. Update the `multisite` settings in `group_vars`, including adding `domain_current_site: example.com` under `env`.
3. (For subdomain installs only) Make sure that the Landrush plugin is installed and then swap in the changes given in the wiki into the `Vagrantfile`.

### DIFF
--- a/roles/wordpress-setup/templates/wordpress-site.conf.j2
+++ b/roles/wordpress-setup/templates/wordpress-site.conf.j2
@@ -24,7 +24,7 @@ server {
   {% if item.value.multisite.enabled | default(false) -%}
     {% if item.value.multisite.subdomains | default(false) -%}
       rewrite ^/(wp-.*.php)$ /wp/$1 last;
-      rewrite ^/(wp-(content|admin|includes)/.*) /wp/$1 last;
+      rewrite ^/(wp-(content|admin|includes).*) /wp/$1 last;
     {%- else -%}
       include wordpress_multisite_subdirectories.conf;
     {%- endif %}
--- a/roles/wordpress-setup/templates/wordpress-site.conf.j2
+++ b/roles/wordpress-setup/templates/wordpress-site.conf.j2
@@ -7,7 +7,7 @@ server {
   listen 80;
   {% endif %}
 
-  server_name  {{ item.value.site_hosts | join(' ') }};
+  server_name  {% for host in item.value.site_hosts %} {{ host }} {% if item.value.multisite.subdomains %} *.{{ host }} {% endif %} {% endfor %};
   access_log   {{ www_root }}/{{ item.key }}/logs/access.log;
   error_log    {{ www_root }}/{{ item.key }}/logs/error.log;
 
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -6,7 +6,6 @@ ENV['VAGRANT_DEFAULT_PROVIDER'] = 'virtualbox'
 
 require 'yaml'
 
-PRIVATE_IP   = '192.168.50.5'
 ANSIBLE_PATH = __dir__ # path targeting Ansible directory (relative to Vagrantfile)
 
 # Set Ansible roles_path relative to Ansible directory
@@ -28,7 +27,7 @@ Vagrant.configure('2') do |config|
   config.ssh.forward_agent = true
 
   # Required for NFS to work, pick any local IP
-  config.vm.network :private_network, ip: PRIVATE_IP
+  config.vm.network :private_network, ip: '192.168.50.5'
 
   hostname, *aliases = wordpress_sites.flat_map { |(_name, site)| site['site_hosts'] }
   config.vm.hostname = hostname
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -50,7 +50,7 @@ Vagrant.configure('2') do |config|
   
   # Comment out this control block for subdomain multisite installs
   if Vagrant.has_plugin? 'vagrant-hostsupdater'
-    config.hostsupdater.aliases = aliases
+    config.hostsupdater.aliases = aliases + www_aliases
   else
     puts 'vagrant-hostsupdater missing, please install the plugin:'
     puts 'vagrant plugin install vagrant-hostsupdater'
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -34,21 +34,6 @@ Vagrant.configure('2') do |config|
   config.vm.hostname = hostname
   www_aliases = ["www.#{hostname}"] + aliases.map { |host| "www.#{host}" }
 
-  # Uncomment this control block for subdomain multisite installs
-  # See note on next control block
-  # if Vagrant.has_plugin? 'landrush'
-  #   config.landrush.enabled = true
-  #   config.landrush.tld = config.vm.hostname
-  # 
-  #   aliases.each do |host|
-  #     config.landrush.host host, PRIVATE_IP
-  #   end
-  # else
-  #   puts 'landrush missing, please install the plugin:'
-  #   puts 'vagrant plugin install landrush'
-  # end
-  
-  # Comment out this control block for subdomain multisite installs
   if Vagrant.has_plugin? 'vagrant-hostsupdater'
     config.hostsupdater.aliases = aliases + www_aliases
   else
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -6,6 +6,7 @@ ENV['VAGRANT_DEFAULT_PROVIDER'] = 'virtualbox'
 
 require 'yaml'
 
+PRIVATE_IP   = '192.168.50.5'
 ANSIBLE_PATH = __dir__ # path targeting Ansible directory (relative to Vagrantfile)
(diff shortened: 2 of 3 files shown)


austinpray:
Will test this soon.

Test cases I can think of off the top of my head:

- Fresh subdomain multisite
- Fresh subdirectory multisite
- Subdomain export database, vagrant destroy, vagrant up with imported database
- Subdirectory export database, vagrant destroy, vagrant up with imported database
- Deploy subdomain to remote DO
- Deploy subdirectory to remote DO

What other edgecases would there be?

roles/wordpress-setup/templates/wordpress-site.conf.j2
```diff
@@ -7,7 +7,7 @@ server {
   listen 80;
   {% endif %}
 
-  server_name  {{ item.value.site_hosts | join(' ') }};
+  server_name  {% for host in item.value.site_hosts %} {{ host }} {% if item.value.multisite.subdomains %} *.{{ host }} {% endif %} {% endfor %};
```
retlehs: nice

swalkinshaw:
@alan-c good catch on the `wp is-installed` task. I'm good with removing it after looking at the WP-CLI source :+1:. Do you want to do this?

alan-c:
Sure, I'll pull it out and tweak the references accordingly on the next two plays.

swalkinshaw:
This works but it results in the task always being reported as "Changed". Would be nice to tweak `change_when` to fix this. It shouldn't be changed if stdout contains "WordPress is already installed."

kalenjohnson:
- [x] Fresh subdomain multisite
- [x] Fresh subdirectory multisite
- [x] Subdomain export database, vagrant destroy, vagrant up with imported database
- [x] Subdirectory export database, vagrant destroy, vagrant up with imported database
- [x] Deploy subdomain to remote DO
- [x] Deploy subdirectory to remote DO

:100: Awesome work @alan-c, everything's working. I took the majority of the evening tonight testing each of these.

Pretty happy to see Multisite working again :joy:

kalenjohnson:
Just did a regular site install as well, no multisite, just to make sure that's working as well. Should be good!

alan-c:
Thanks, glad to help. This afternoon/evening I've got a tweak on the `changed_when` as suggested by @swalkinshaw which i'll add in.

ascottmccauley:
I incorporated the changes here, and am still getting the error: `Notice: Constant MULTISITE already defined in /srv/www/test.com/current/web/wp-config.php on line 11`, because of the `is_multisite` problem. Is there something else that I might be missing?

kalenjohnson:
@ascottmccauley there should be no constants defined in `wp-config.php`, they should all be in `config/application.php`

ascottmccauley:
@kalenjohnson, Right. Forgot that was a holdover from trying everything else under the sun to get multisite to work.

roles/wordpress-setup/templates/wordpress-site.conf.j2
austinpray: :sunglasses: @fullyint @louim do you see any better way to do this? I came up with this under pressure and obviously it looks kinda dirty.

fullyint: Without having followed the bigger picture of the multisite discussion (I might be missing something), I think what you have is the best way to handle it. 

One (untested) alternative that comes to mind is just to tell users that if they're using multisite, they should add their the subdomain wildcard version to the list of `site_hosts`:
```yaml
site_hosts:
  - site1.com
  - *.site1.com
  - site2.com
  - *.site2.com
```
Then you could just keep the original 
`server_name  {{ item.value.site_hosts | join(' ') }};`

The other idea is pretty bad and I haven't tested: 
`server_name  {{ item.value.site_hosts | join(' ') }} {% if item.value.multisite.subdomains %} *.{{ item.value.site_hosts | join(' *.') }} {% endif %};`

austinpray: >they should add their the subdomain wildcard version to the list of `site_hosts`

This causes catastrophic failure with landrush on development. Dunno why though.

louim: @austinpray this seems the most logical way to handle it. 

> I'm told this fails 

@fullyint  have you been told why it fails?
I'm not sure if I understand [this](https://github.com/phinze/landrush#wildcard-subdomains) correctly (haven't tested), but to me it looks like using:

```
site_hosts:
  - site1.com
```

`*.site.com` would automatically be routed correctly to the IP of the vagrant box?

fullyint: @louim I haven't been told why the wildcard in `site_hosts` fails, and I'm afraid I haven't looked into it. Here is an [example of it failing](https://discourse.roots.io/t/multisite-provisioning-failure/4328/5), but here's an example of a [trellis fork where it doesn't fail](https://github.com/roots/trellis/issues/266#issuecomment-120727239).

alan-c: @louim Yes, that IP routing behavior is what I have been seeing/using.  

@fullyint Landrush does not allow executing this with a wildcard host:

`config.landrush.host host, PRIVATE_IP`

Why? No idea. I didn't find anything in a cursory look through the Landrush docs or issues.

The fork where it doesn't fail that you linked configures Landrush/the Vagrant box in a different manner versus the method presented in the wiki. Among several things, it doesn't use `site_hosts` to generate the subdomains for Landrush/the Vagrant box - and therefore doesn't encounter the DNS armageddon that results. Instead it uses configuration code that seems more specifically adapted to its environments/needs.

When I was testing things, it seemed to me that the most basic choice was between removing the wildcard from `site_hosts` and leaving the wiki as is on one hand or using the wildcard and significantly reworking the Landrush configuration code on the other. As my Landrush experience is minimal, I went with the former - but perhaps another approach is better/more elegant?

fullyint: Looks good to me @alan-c. Thanks for all the work you've done on this.

austinpray:
Rebased and fixed up over here https://github.com/roots/trellis/pull/319

roles/wordpress-setup/templates/wordpress-site.conf.j2
```diff
@@ -24,7 +24,7 @@ server {
   {% if item.value.multisite.enabled | default(false) -%}
     {% if item.value.multisite.subdomains | default(false) -%}
       rewrite ^/(wp-.*.php)$ /wp/$1 last;
-      rewrite ^/(wp-(content|admin|includes)/.*) /wp/$1 last;
+      rewrite ^/(wp-(content|admin|includes).*) /wp/$1 last;
```
brocheafoin: I added the following line here:

`  rewrite ^/(xmlrpc.php)$ /wp/$1 last;`

or else XML-RPC-dependent stuff (Jetpack authentication, in my case) was not working. Can submit a PR if you want. Thanks for all your hard work!

retlehs: this seems like a core bug since xmlrpc breaks when WP is installed in a subdirectory. i'm not sure this is something that we should add to trellis, but it's something we could at least cover in the wiki/docs.

brocheafoin: Yeah, probably not exactly related to fixing multisite, then. Please disregard.

drumba:
@kalenjohnson I tested the multisite subdirectory deployment to remote DO and it is not working for me. I always get a Database Connection error. The tables are created. On the VM everything is working fine. I think there is still some strange thing in the configuration.

developdaly:
Same issue as @kalenjohnson -- local is fine, remote breaks. cc/ @bbttxu

austinpray:
Can you guys create an issue with some info on how to reproduce?

PR #287, opened 2015-08-03
Task for setting right php-fpm logfile permissions

ZloeSabo:
Need to be sure that php-fpm has access to write to it's log
before service starts to avoid failure.


austinpray:
Haven't run into this, on what kind of setup does this occur for you?

swalkinshaw:
I've never run into this either or had any other bug reports of it. If you have any more details we can re-visit this.

PR #291, opened 2015-08-05
Update main.yml

BrandonShutter:
Caching was breaking carts.  This prevents nginx from caching cart items.  

You can see the bug here: https://www.caltonnutrition.com/app/uploads/2015/08/CaltonBug.mp4

adding `cart` and `checkout` prevents `nginx` from breaking this again.

### DIFF
--- a/roles/nginx/defaults/main.yml
+++ b/roles/nginx/defaults/main.yml
@@ -9,5 +9,5 @@ nginx_cache_duration: 30s
 nginx_cache_key_storage_size: 10m
 nginx_cache_size: 250m
 nginx_cache_inactive: 1h
-nginx_skip_cache_uri: /wp-admin/|/xmlrpc.php|wp-.*.php|/feed/|index.php|sitemap(_index)?.xml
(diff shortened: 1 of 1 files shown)


swalkinshaw:
We're only going to support core/native WordPress functionality here. We made that a default variable so it could easily be overwritten.

Just re-define `nginx_skip_cache_uri` in `group_vars/all`.

BrandonShutter:
Thanks.  I need to learn more about `ansible` default variables obviously.

swalkinshaw:
See http://docs.ansible.com/ansible/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable

austinpray:
I don't have any problems with adding a section about application specific configs over on [the wiki](https://github.com/roots/trellis/wiki/FastCGI-caching).

kalenjohnson:
I've added this:

> If you have specific pages you don't want cached, feel free to override `skip_cache_uri` in `group_vars/all` or the specific environment, and add the page names:
```yml
skip_cache_uri: /wp-admin/|/xmlrpc.php|wp-.*.php|/feed/|index.php|sitemap(_index)?.xml|checkout|cart
```

PR #292, opened 2015-08-06
Fix comment in Vagrantfile: use absolute path for ANSIBLE_PATH

fullyint:
To be consistent with #257 and prevent issues like https://discourse.roots.io/t/vagrant-global-status-error/4417


swalkinshaw:
:+1:

PR #293, opened 2015-08-08
Comment out lookup of potentially nonexistent public keys.

chriszarate:
In our fork of Trellis, I like to keep `group_vars/all` unchanged from upstream and override values in separate vars files. This has been a good approach, especially since the values in `group_vars/all` are generally sensible defaults.

The `users` hash, though, provides potentially nonexistent public key paths (`~/.ssh/id_rsa.pub`) that can lead to an error during playbook execution. As a nested hash value, it's also difficult to override. Commenting them out, as is done with the example GitHub value, seems like a flexible and consistent approach.

### DIFF
--- a/group_vars/all
+++ b/group_vars/all
@@ -22,13 +22,13 @@ users:
     groups:
       - "{{ web_group }}"
     keys:
-      - "{{ lookup('file', '~/.ssh/id_rsa.pub') }}"
+      # - "{{ lookup('file', '~/.ssh/id_rsa.pub') }}"
       # - https://github.com/username.keys
   - name: "{{ admin_user }}"
     groups:
       - sudo
     keys:
-      - "{{ lookup('file', '~/.ssh/id_rsa.pub') }}"
+      # - "{{ lookup('file', '~/.ssh/id_rsa.pub') }}"
       # - https://github.com/username.keys
 
 admin_user: admin


louim:
I'm not quite sure how someone without a key would use trellis, but in my opinion, I think having the user key set by default is a good option. I'm also wondering why it would be harder to override?

I would rather see:
`"{{ lookup('file', '~/.ssh/id_rsa.pub') | default(omit) }}"` than having the key commented out.

chriszarate:
It's not we don't use public key auth, it's that we don't want public key auth for `web_user` (or even `admin_user` in some circumstances, if we're accessing via another account).

I have Ansible's `hash_behaviour` set to `merge` and was having trouble getting Trellis to not try copying a key from that default location (which doesn't exist on my system).

I'll test `default(omit)`.

fullyint:
I'm also inclined to keep the default key because I'm guessing that would help Trellis work out of the box for most users.

I like the `default(omit)` filter @louim. It looks like it would resolve the issue here, and more...

Perhaps Trellis users who will not use `deploy.yml` shouldn't have to supply a key for `web_user`. In this case, `dev.yml` and `server.yml` don't need to load the key on the server and `default(omit)` would prevent them from failing due to a missing key. 

`default(omit)` could apply to `admin_user` too, with caution. Suppose a user disables root login for a remote VPS and has no pub key file for `admin_user`. `server.yml` would finish without loading a pub key on the server for `admin_user`. Neither `root` nor `admin_user` could connect via SSH.

To deal with this, we could finish the `users` role with a `fail` task causing `server.yml` to fail when all these conditions are met:
* not vagrant vm
* `sshd_permit_root_login = "no"`
* server has no key for `admin_user`

chriszarate:
Piping to `default(omit)` doesn't work. The lookup throws an error (as opposed to just an undefined value) and stops playbook execution. I'll try some other strategies to override the opinionated default public key path and report back here.

chriszarate:
I take it back. I think piping to `default(omit)` should work. Further overriding `users` *should* be fairly easy (it's a list, not a hash, duh).

I was confused by some very strange behavior which appears to be caused by this Ansible bug:
https://github.com/ansible/ansible/issues/9944

Basically, even if I override the users variable entirely, Ansible is still trying to template the lookup in `group_vars/all` and throws an error when it can't find `~/.ssh/id_rsa.pub`—even though it does nothing with it.

Not sure what the best course of action here is. This seems like it should be affecting more people than just me. Or does everyone still have `~/.ssh/id_rsa.pub` so they never run into this?

austinpray:
>Not sure what the best course of action here is. This seems like it should be affecting more people than just me. Or does everyone still have ~/.ssh/id_rsa.pub so they never run into this?

Pretty much everyone I have ever worked with and set up on trellis followed [this guide](https://help.github.com/articles/generating-ssh-keys/) and copypasta'd the heck out of the commands. Also ssh-keygen defaults to id_rsa as well.

That said: I do agree this is an issue. Basically the only way the ssh keys will work in the current implementation in a large team, many maintainers situation:

1. All the maintainers have an id_rsa
2. All the maintainers have a github account with their keys configured
3. You add the list of public keys to a file in the repo
4. You maintain a list of public keys on your own server exposed via http

You should definitely technically be able to do:

```
- "{{ lookup('file', '~/.ssh/id_rsa.pub') | default(omit) }}"
- "{{ lookup('file', '~/.ssh/special_snowflake.pub') | default(omit) }}"
```

But won't this lead to a tug of war? Every time each individual maintainer provisions a server the other maintainers will lose access right?

For complicated situations like this adding a list of public keys to the repo sounds pretty ideal actually.

Related https://github.com/roots/trellis/issues/294

fullyint:
@chriszarate 
> Piping to `default(omit)` doesn't work. The lookup throws an error (as opposed to just an undefined value)

This led me to test. Despite using `default(omit)`, I still got the lookup error for missing local key files, causing the playbook to fail. Maybe you've found a way around it?

With just preliminary testing, the idea below
* successfully skips if  local key file is missing (see `when:` statement on task)
* spares Trellis users from having to see `lookup(stuff)` in `group_vars/all`

```yaml
# group_vars/all

users:
  ⋮
    keys:
      - path: ~/.ssh/id_rsa.pub
      - url: https://github.com/username.keys
```

```yaml
# roles/users/tasks/main.yml

- name: Add SSH keys
  authorized_key: user="{{ item.0.name }}"
                  key="{% if item.1.url is defined %}{{ item.1.url }}{% else %}{{ lookup('file', item.1.path) }}{% endif %}"
  with_subelements:
    - users | default([])
    - keys
  when: item.1.url is defined or lookup('pipe', '[ -f ' + item.1.path + ' ] && echo "file_exists" || :') == 'file_exists'
```
Not super pretty.

Still could use an additional task (or something) to check that the server has a pub key for `admin_user` if `root` login will be disabled.

@austinpray 
> But won't this lead to a tug of war? Every time each individual maintainer provisions a server the other maintainers will lose access right?

I'd have to double-check, but I think [`authorized_key`](http://docs.ansible.com/ansible/authorized_key_module.html) just concatenates keys into `~/.ssh/authorized_keys` on server, only removing a key if the parameter `state=absent`.

swalkinshaw:
I don't think `omit` works in this case. I believe the problem is that Ansible evaluates the `lookup` before it's actually used in the module argument.

We've bene making progress in making Trellis more user-friendly with better errors. So instead of commenting this out I'd rather we do a check if keys exist and `fail` with a helpful error message if not.

PR #301, opened 2015-08-15
Allow user to set the timezone of the server and php

Kimisyl:
Add a way to set server's timezone from a user value. The user value will also be used to set `date.timezone` for php-fpm.default

The timezone is compared with zone files before being set to ensure its validity.

### DIFF
--- a/roles/common/tasks/main.yml
+++ b/roles/common/tasks/main.yml
@@ -18,3 +18,21 @@
   - python-mysqldb
   - curl
   - git-core
+
+- name: Validate timezone variable
+  stat: path=/usr/share/zoneinfo/{{ default_timezone }}
+  register: timezone_path
+  changed_when: false
+
+- name: Explain timezone error
+  fail: msg="{{ default_timezone }} is not a valid timezone. For a list of valid timezones, check https://php.net/manual/en/timezones.php"
+  when: not timezone_path.stat.exists
+
+- name: Get current timezone
+  command: cat /etc/timezone
+  register: current_timezone
+  changed_when: false
+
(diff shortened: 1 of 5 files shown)


roles/common/tasks/main.yml
```diff
@@ -18,3 +18,22 @@
   - python-mysqldb
   - curl
   - git-core
+
+- name: Validate timezone variable
+  stat: path=/usr/share/zoneinfo/{{ default_timezone }}
+  register: valid_timezone
```
swalkinshaw: How about naming this `timezone_path` (since it might not be valid)?

roles/common/tasks/main.yml
```diff
@@ -18,3 +18,22 @@
...
+  changed_when: false
+
+- name: Explain timezone error
+  fail: msg="{{ default_timezone }} is not a valid timezone. For a list of valid timezones, check https://php.net/manual/en/timezones.php"
+  when: valid_timezone.stat.exists == False
```
swalkinshaw: `when: not valid_timezone.stat.exists` should work? Little nicer.

roles/common/tasks/main.yml
```diff
@@ -18,3 +18,22 @@
...
+
```
swalkinshaw: Extra blank line here

swalkinshaw:
This is great :+1:. Just a few minor suggestions. I really like the helpful error message as well.

Kimisyl:
@swalkinshaw Fixed and rebased.

swalkinshaw:
Tested and works great. Thanks!

Issue #298, opened 2015-08-11
Add option to block outgoing mail on development


retlehs:
there should be an option to block outgoing mail for development environments

right now it's possible to accidentally send out emails to users/customers/clients, which you probably don't want to do on dev

maybe have a whitelist of allowed domains?

louim:
Better option should be to use [Mailhog](https://github.com/mailhog/MailHog) in the development environment by default to catch all mails.

retlehs:
nice. @kalenjohnson also mentioned http://mailcatcher.me/

kalenjohnson:
Mailhog's catchline: "Inspired by MailCatcher, easier to install."

Sounds good

Built in Go. Sorry Mailhog

louim:
In fact Mailhog is directly inspired from mailcatcher and do the same work, but being in go, it is self contained, and doesn't need Ruby to be run. It provides binary for all platforms https://github.com/mailhog/MailHog/releases

kalenjohnson:
I'd be up to give it a try. Ruby isn't installed on a Trellis box by default, so if something works just as well, no need to have it installed.

retlehs:
https://github.com/geerlingguy/ansible-role-mailhog

retlehs:
closed by #304

PR #304, opened 2015-08-16
Capture and display all mail with MailHog in development

louim:
This is a preliminary PR covering issue #298. There is a couple of things I'd like to address / get feedback on before considering it ready to be merged, but it's working great in its current form. See comments inline on the code.

### DIFF
--- a/group_vars/development
+++ b/group_vars/development
@@ -1,4 +1,5 @@
 mysql_root_password: devpw
+mailhog_install_ssmtp: no
 
 web_user: vagrant
 
@@ -29,6 +30,15 @@ wordpress_sites:
       db_name: example_dev
       db_user: example_dbuser
       db_password: example_dbpassword
+
+mail_smtp_server: localhost:1025
+mail_admin: admin@example.dev
+mail_hostname: example.dev
+ssmtp_auth_method: ""
+ssmtp_start_tls: 'no'
+ssmtp_tls: 'no'
+
+ferm_enabled: false
 
 php_error_reporting: 'E_ALL'
 php_display_errors: 'On'
--- a/requirements.yml
+++ b/requirements.yml
@@ -7,3 +7,5 @@
   src: MaximeThoonsen.php5-xdebug
 - name: swapfile
   src: kamaln7.swapfile
+- name: mailhog
+  src: geerlingguy.mailhog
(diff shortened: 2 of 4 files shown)


group_vars/development
```diff
@@ -1,4 +1,6 @@
 mysql_root_password: devpw
+mailhog_binary_url: https://github.com/mailhog/MailHog/releases/download/v0.1.7/MailHog_linux_amd64
```
louim: This will not be needed once https://github.com/geerlingguy/ansible-role-mailhog/pull/6 is merged in.

group_vars/development
```diff
@@ -30,6 +32,24 @@ wordpress_sites:
...
+ferm_input_list:
+  - type: dport_accept
+    dport: [http, https]
+    filename: nginx_accept
+  - type: dport_accept
+    dport: [8025]
+  - type: dport_limit
+    dport: [ssh]
+    seconds: 300
+    hits: 20
```
louim: This block is mostly a copy of the one in the `all` settings. Because it is not possible to append to a list in Ansible its not possible to only add the rule to open port `8025`, I have to override the whole list.

another option would be to put `ferm_enabled: false` for development by default. what do you guys think?

requirements.yml
```diff
@@ -7,3 +7,6 @@
   src: MaximeThoonsen.php5-xdebug
 - name: swapfile
   src: kamaln7.swapfile
+- name: mailhog
+  src: https://github.com/Berdir/ansible-role-mailhog
```
louim: I'm using the fork of the pull request until https://github.com/geerlingguy/ansible-role-mailhog/pull/2 in merged, because the role install ssmtp by default and we do it too, so it would conflict.

group_vars/development
swalkinshaw: Let's just disable it in development. This goes a little beyond dev/prod parity.

requirements.yml
retlehs: looks like it got merged thanks to you! :+1:

louim:
Thanks to the speedy merge on the Mailhog Ansible role, the PR is now updated and ready to go. I added some information on the readme about how to access Mailhog in development. Feel free to find better wording to explain it! :+1:

swalkinshaw:
Tested and working great. Want to squash?

louim:
@swalkinshaw done!

Issue #303, opened 2015-08-15
Deploy local commands: rsync permissions


swalkinshaw:
See https://github.com/roots/roots-example-project.com/issues/22 for more details.

/cc @fullyint

fullyint:
@austinpray pointed out that "the [synchronize](https://docs.ansible.com/ansible/synchronize_module.html) module uses the `-a` flag by default."
"This is equivalent to **-rlptgoD**" ([man page](http://manpages.ubuntu.com/manpages/trusty/en/man1/rsync.1.html)).

**-r, --recursive**: (keep)
**-l, --links**: (keep) copy symlinks as symlinks
**-p, --perms**: (drop) this is the problem, so we'll set the parameter `perms=no`
**-t, --times**: (keep) preserve modification times
**-g, --group**: (drop) preserve group. We'll set the parameter `group=no`.
**-o, --owner**: (drop) preserve owner (super user only). We'll set the parameter `owner=no` just to be safe (?).
**-D** The -D option is equivalent to --devices --specials
**--devices**: (keep?) "transfer character and block device files to the remote system to recreate these devices"
**--specials**: (keep?) "transfer special files such as named sockets and fifos"

In my tests source permissions were synced over, but the owner:group seemed to always end up `web:www-data`. I tried adding `perms=no group=no owner=no` to [our synchronize task](https://github.com/roots/trellis/blob/f18d68fea6501b9e2180173845209b4489607a9b/roles/deploy/tasks/main.yml#L33) and the read and write permissions were no longer synced. I don't understand, however, why the executability was always synced, even if I ran the rsync command manually. Maybe it's just my machine or something I'm not understanding. 

I'm pretty inexperienced with permissions and I'm left wondering whether some people would actually want their local files' custom permissisons to sync over (?). 

If we want to explicitly set permissions, adding the following would set directories to 755 and files to 644:
`rsync_opts=--chmod=Du=rwx,--chmod=Dg=rx,--chmod=Do=rx,--chmod=Fu=rw,--chmod=Fg=r,--chmod=Fo=r`
(I didn't have to repeat the `--chmod=` on a manual rsync, but it seemed necessary with the Ansible synchronize module.)

Would we want to set permissions explicitly? Are those the permissions we'd want (D755,F644)?

swalkinshaw:
I don't think we should ever have local permissions being transferred over by default. I'd say it's just a bad practice since you wouldn't expect it and could lead to issues.

644 should be good.

[...]

PR #309, opened 2015-08-19
Set permisssions on files deployed via rsync

fullyint:
Followup to #303 and roots/roots-example-project.com#22

Unfortunately we can't just use
`rsync_opts=--chmod=Du=rwx,Dg=rx,Do=rx,Fu=rw,Fg=r,Fo=r`
because of ansible/ansible-modules-core#1163


swalkinshaw:
:+1:

Issue #288, opened 2015-08-03
Consolidate host files


fullyint:
Trellis could use a single host file instead of separate host files per environment. A single file may be cleaner conceptually and technically. It would also accommodate drop-in playbooks that need to access hosts from multiple environments at the same time (e.g., a playbook that syncs the db from a staging host to a production host).

### Example Host File

|                |      Web hosts     | DB hosts |
| --------------:|:------------------:|:--------:|
| **Staging**    | webhost1           | dbhost1  |
| **Production** | webhost2, webhost3 | dbhost2  |

The hosts above could translate to the host file below.
```
[staging]
webhost1
dbhost1

[production]
webhost2
webhost3
dbhost2

[web]
webhost1
webhost2
webhost3

[db]
dbhost1
dbhost2
```

### Revised Commands
* `ansible-playbook server.yml -e env=staging` (or `env=production`)
* `ansible-playbook dev.yml` (see note about "Development Host File" below)

### Changes that Users would Face
* Transfer hosts to new host file, a one-time task per Trellis install.
* Remember to use revised commands
* Can run `ansible-playbook dev.yml` if desired over `vagrant up` or `vagrant provision`. This makes it easier to add the `--tags` flag to run only portions of the playbook. (see note about "Development Host File" below)

### Edits to Trellis
**`ansible.cfg`**
* Add `inventory = hosts` to the `[defaults]`. This `hosts` is a directory like the current `hosts` directory.

**`hosts` directory**
* (**Edit:** This `hosts` directory idea is revised in later comments.)
* The `hosts` directory would contain an inventory file named `non-dev` (or something). The file content would be like the example host file above.
* The `hosts` directory could contain something to make Ansible aware of the development host defined at `.vagrant/provisioners/ansible/inventory/vagrant_ansible_inventory`. This "something" could be a relative symlink, a copy of the vagrant host file, etc. (see note about "Development Host File" below)

**`hosts` parameter in playbooks**
* `hosts: web:&development` in `dev.yml`
* `hosts: web:&{{ env }}` in `server.yml` and `deploy.yml`

If a user were to run the command `ansible-playbook server.yml -e env=staging`

> This would mean the hosts must be in the group webservers and the host must also be in the group staging
- http://docs.ansible.com/ansible/intro_patterns.html

**`deploy.sh`**
* adjust to accommodate `env` extra-var

### Development Host File
[...]

PR #313, opened 2015-08-27
Refactor hosts files

fullyint:
* Enables cross-environment playbooks (e.g., syncing files/database) by loading entire `hosts` directory
* Accommodates separate web vs. db servers by making `[web]` its own group instead of a child group of `<environment>`.
* Pairs nicely with #314 
* Initial discussion in #288

Commands
* Replace `-i hosts/<environment>` with `-e env=<environment>`
* Dev can just be `ansible-playbook dev.yml`. You'd need vagrant vm in ssh config till #314 is merged.
* unchanged: `./deploy.sh <environment> <site>`

I kept hosts files separate by environment. Alternatively, a single file could look like this: 

```
# Each host machine should appear in the "Environments" section and in the "Types" section.
# Each host must only be listed once per [group], even if it will host multiple sites.


# Environments

[development]
127.0.0.1

[staging]
192.168.50.6

[production]
192.168.50.7


# Types

[web]
127.0.0.1
192.168.50.6
192.168.50.7

[db]

```

### DIFF
--- a/server.yml
+++ b/server.yml
@@ -1,12 +1,16 @@
 ---
+- include: variable-check.yml
+  vars:
+    playbook: server.yml
+
 - name: Determine Remote User
-  hosts: web
+  hosts: web:&{{ env }}
   gather_facts: false
   roles:
     - { role: remote-user, tags: [remote-user, always] }
 
 - name: WordPress Server - Install LEMP Stack with PHP 5.6 and MariaDB MySQL
-  hosts: web
+  hosts: web:&{{ env }}
   sudo: yes
   vars_files:
     - vars/sudoer_passwords.yml
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Refactor hosts files ([#313](https://github.com/roots/trellis/pull/313))
 * Fixes #436 - Let WP handle 404s for PHP files ([#448](https://github.com/roots/trellis/pull/448))
 * Fixes #297 - Use `php_flag` vs `php_admin_flag` ([#447](https://github.com/roots/trellis/pull/447))
 * Fixes #316 - Set WP permalink structure during install ([#316](https://github.com/roots/trellis/pull/316))
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ For remote servers, you'll need to have a base Ubuntu 14.04 server already creat
 2. Add your server IP/hostnames to `hosts/<environment>`.
 3. Specify public SSH keys for `users` in `group_vars/all/users.yml`. See the [SSH Keys docs](https://roots.io/trellis/docs/ssh-keys/).
 4. Consider setting `sshd_permit_root_login: false` in `group_vars/all/security.yml`. See the [Security docs](https://roots.io/trellis/docs/security/).
-5. Run `ansible-playbook -i hosts/<environment> server.yml`.
+5. Run `ansible-playbook server.yml -e env=<environment>`
 
 ## Deploying to remote servers
 
@@ -84,7 +84,7 @@ Full documentation: https://roots.io/trellis/docs/deploys/
 1. Add the `repo` (Git URL) of your Bedrock WordPress project in the corresponding `group_vars/<environment>/wordpress_sites.yml` file.
 2. Set the `branch` you want to deploy (optional - defaults to `master`).
 3. Run `./deploy.sh <environment> <site name>`
-4. To rollback a deploy, run `ansible-playbook -i hosts/<environment> rollback.yml --extra-vars="site=<site name>"`
+4. To rollback a deploy, run `ansible-playbook rollback.yml -e "site=<site name> env=<environment>"`
 
 ## Configuration
 
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -1,6 +1,7 @@
 [defaults]
 roles_path = vendor/roles
 force_handlers = True
+inventory = hosts
 
 [ssh_connection]
 ssh_args = -o ForwardAgent=yes -o ControlMaster=auto -o ControlPersist=60s
--- a/deploy.sh
+++ b/deploy.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 shopt -s nullglob
 
-DEPLOY_CMD="ansible-playbook -i hosts/$1 deploy.yml --extra-vars="site=$2""
+DEPLOY_CMD="ansible-playbook deploy.yml -e env=$1 -e site=$2"
 ENVIRONMENTS=( hosts/* )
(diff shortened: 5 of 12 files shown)


austinpray:
Going to try this out tonight

swalkinshaw:
Think I'm onboard with this :+1:

swalkinshaw:
@fullyint can you do a PR for doc updates on https://github.com/roots/docs?

We should merge at the same time.

fullyint:
docs in roots/docs#4

After this PR, commands will use `-e env=<environment>` instead of `-i hosts/<environment>`. This PR initiates playbooks with a task to "Ensure environment is defined," giving a helpful message if not. Otherwise, omitting the new `-e env` would produce the message `skipping: no hosts matched`, which is less clear. We can remove this task in the future once most users have transitioned to the new command.

server.yml
```diff
@@ -1,12 +1,19 @@
 ---
+- hosts: localhost
+  tasks:
+    - name: Ensure environment is defined
+      fail:
+        msg: "Environment missing. Use `-e` to define `env`:\nansible-playbook server.yml -e env=<environment>"
+      when: env is not defined
+
```
swalkinshaw: could this be separated into its own playbook and included so it's DRY?

swalkinshaw:
@fullyint this will need an update to `windows.sh` unfortunately

server.yml
fullyint: Great idea.

I added a commit to include the new `variable-check.yml` playbook. It's DRY but also a bit more complex to look at. The messages needed to differ slightly per playbook.

You'll see that I went with yaml:
```
- include: variable-check.yml
  vars:
    playbook: server.yml
```
even though it could have been more succinct:
```
- include: variable-check.yml playbook=server.yml
```

fullyint:
@swalkinshaw Re: updating `windows.sh`, I figure the [`ansible-playbook` command](https://github.com/roots/trellis/blob/1ccb789fbce617b7497bfaff14f7ecd434f94baa/windows.sh#L45) is the only candidate for consideration.

**Inventory.** This PR removes the `-i` everywhere except in `windows.sh`. When ansible runs `dev.yml` within a Vagrant VM on Windows, I think it must use the `chmod -x ${TEMP_HOSTS}/*` copy made by `windows.sh`. So, I think `windows.sh` must retain the `-i` to override `ansible.cfg`'s new default `inventory = hosts`.

**`env` variable.** `windows.sh` runs the `dev.yml` playbook which is actually the only playbook that doesn't use the `env` variable introduced in this PR. Instead, `dev.yml` specifies the environment group directly: 
```
hosts: web:&development
```
So, I don't think `windows.sh` needs to add the `-e` option for defining `env`.

I don't notice anything else that might need adjustment in `windows.sh`. Let me know if I missed anything.

rposborne:
+1

swalkinshaw:
@fullyint squash + changelog entry? Ready to merge after that :+1:

fullyint:
Added changelog entry and squashed

scherii:
*Please feel free to tell that I should open a new issue if this doesn't belong here:*

With this change we now have to enter the domain name instead of the server's IP.

`./deploy.sh alpha alpha.domain.tld`
```
TASK: [Ensure site is valid] **************************************************
failed: [XXX.XXX.XXX.XXX] => {"failed": true}
msg: Site `alpha.domain.tld` is not valid. Available sites to deploy: beta.domain.tld
```

(Our staging stages are named `alpha` and `beta`.)
So our host files look like this:

```
[beta]
beta.domain.tld

[web]
beta.domain.tld
```

After replacing the IP with the domain name it works again.

fullyint:
@scherii 

**Setting the stage**

`./deploy.sh `[`<group>`](https://github.com/roots/trellis/blob/11516d12e42355d0304eb4fb8c15be7ca8fcf6d4/hosts/staging#L4)[`<site>`](https://github.com/roots/trellis/blob/11516d12e42355d0304eb4fb8c15be7ca8fcf6d4/group_vars/staging/wordpress_sites.yml#L3)

By specifying the `group` in your command, you select a subset of all the hosts in your [hosts directory](https://github.com/roots/trellis/blob/11516d12e42355d0304eb4fb8c15be7ca8fcf6d4/ansible.cfg#L4), specifically the [hosts subset](https://github.com/roots/trellis/blob/11516d12e42355d0304eb4fb8c15be7ca8fcf6d4/deploy.yml#L7) that are both in the `group` (or `env`) you specified _and_  the `web` group.

The `Ensure site is valid` task/error checks that the `site` in your command is listed in the `wordpress_sites` that are applicable to the `group` specified.

For example, the default command `./deploy staging example.com` checks that [`example.com`](https://github.com/roots/trellis/blob/11516d12e42355d0304eb4fb8c15be7ca8fcf6d4/group_vars/staging/wordpress_sites.yml#L3) is in the staging group's `wordpress_sites` (i.e., in `group_vars/staging/wordpress_sites.yml`). 

**The error you saw**

If you ran the command `./deploy.sh alpha alpha.domain.tld` and it gave the message `Site `\``alpha.domain.tld`\`` is not valid`, that indicates that the `alpha` group's `wordpress_sites` didn't include the site name `alpha.domain.tld`. To try to help, it reports the sites it can find for the `alpha` group: `Available sites to deploy: beta.domain.tld`.

(sounds like you've already solved things, but...)
To fix this, there are several ways you could associate a list of `wordpress_sites` to the alpha group (to be sure that `alpha.domain.tld` is in the list). You could define them in `group_vars/alpha/wordpress_sites.yml`. Alternatively, you could leave them defined in `group_vars/staging/wordpress_sites.yml` but adjust your `hosts/staging` like this:

```
[alpha]
111.xxx.xxx.xxx

[beta]
122.xxx.xxx.xxx

[web]
111.xxx.xxx.xxx
122.xxx.xxx.xxx

[staging]
alpha
beta
```
There are probably other possible approaches too. Check out [hosts and groups](http://docs.ansible.com/ansible/intro_inventory.html#hosts-and-groups) at Ansible docs.

**Using IP vs domain**

As for IP vs domain, I think you're saying that you used to be able to use IPs in `hosts/staging` but now it appears you must use domain names. 

I think you can still use IPs in your hosts file. Given the example hosts file above, if you were to issue the command `./deploy.sh alpha example.com` it will deploy the `example.com` site to the host `111.xxx.xxx.xxx`. And of course `./deploy.sh beta example.com` will deploy to `122.xxx.xxx.xxx`.

You might also be able to use those same commands with this simplified hosts file (untested):
```
alpha ansible_ssh_host=111.xxx.xxx.xxx
beta ansible_ssh_host=122.xxx.xxx.xxx

[staging]
alpha
beta

[web]
alpha
beta
```
----

I hope that addresses your point. If you feel there is a bug or fix that is needed, feel free to respond here. If you have questions or would like assistance troubleshooting, check out [Roots discourse](https://discourse.roots.io).

Issue #295, opened 2015-08-10
Dynamic development IP


evanfuture:
So, I have a bash script that gets the latest IP matching the roots default pattern (from the Vagrantfile), adds one to it and sed's it back into the Vagrantfile.  This being so I can have launch multiple boxes without having to launch my /etc/hosts file every time:  https://gist.github.com/evanfuture/baa53fced4b18783e498

Is there a way to integrate this into Ansible maybe, as I find it saves a lot of time?

Referencing: https://github.com/roots/trellis/blob/master/Vagrantfile

louim:
Seems like if would be feasible using `vagrant-hostmanager` instead of `vagrant-hostupdater`.

changing the line referencing the ip in the `Vagrantfile` to : 
```
config.vm.network "private_network", type: "dhcp"
```
and implementing this in the `Vagrantfile`:

```
cached_addresses = {}
config.hostmanager.ip_resolver = proc do |vm, resolving_vm|
  if cached_addresses[vm.name].nil?
    if hostname = (vm.ssh_info && vm.ssh_info[:host])
      vm.communicate.execute("/sbin/ifconfig eth0 | grep 'inet addr' | tail -n 1 | egrep -o '[0-9\.]+' | head -n 1 2>&1") do |type, contents|
        cached_addresses[vm.name] = contents.split("\n").first[/(\d+\.\d+\.\d+\.\d+)/, 1]
      end
    end
  end
  cached_addresses[vm.name]
end
```

This is stolen from: https://github.com/smdahlen/vagrant-hostmanager/issues/86. I haven't tester, so you may need to fiddle a bit with this to get it working. 

Pull request is appreciated if you get it working! :smile:

evanfuture:
I found that before, but I wasn't sure what would be lost by removing vagrant-hostupdater from trellis.

louim:
@evanfuture `vagrant-hostupdater` doesn't seems to be maintained anymore (last commit is from 31 Oct 2013). I think it's all for the best to switch to hostmanager.

nathanielks:
I use https://github.com/oscar-stack/vagrant-auto_network in my projects. I'm a fan.

evanfuture:
Ok, having looked at both plugins, I'm not sure its actually necessary to have a new dependency.  After all, the vagrantfile is in Ruby.  So, I did this:

    # next lines are commands to retrieve the latest ip address used of our prefered pattern from your etc/hosts file.
[...]

PR #314, opened 2015-08-27
Add ansible-ssh role to generate ssh config

fullyint:
* Run `ansible-playbook dev.yml` without vagrant commands and without manually adding dev vm to ssh config. Grabs latest `vagrant ssh-config`, sorting out port forwarding between multiple VMs.
* Pairs nicely with #313 to run cross-environment playbooks (syncing files/database), including dev.
* Now you can use ssh config "includes". Files listed in `ssh_config_files` are `cat`-ed into `roles/ansible-ssh/files/config`.
* Initial discussion in #288
* #295 "Dynamic development IP" would be a really nice accompaniment

Could someone test on Windows?	
* May have to adjust or comment out the `ssh_config_files` in `group_vars/all` (?)
* The `vagrant ssh-config` command that runs would typically not be available when Windows users run `dev.yml` inside the vm. This gives return code 127, which I catch with `vagrant_ssh_config.rc == 127`, then set the `host` and `connection`, like in [`windows.sh`](https://github.com/roots/trellis/blob/8a84639257bdeaceb3b440823e2ac78c5c17906a/windows.sh#L39).
* Are the parameters `--sudo --user=vagrant` necessary in `windows.sh`? They already exist in `dev.yml`.

**Caution.** This refers to the vagrant vm as `vagrant-vm` in the hosts file. For consistency, this PR renames vagrant's reference to the vm from `default` to `vagrant-vm`. This latter means the Vagrantfile will stop working for existing VMs named `default`. Users would have to comment out the line
```
config.vm.define 'vagrant-vm'
```
and destroy the vm, then uncomment the line and `vagrant up`.

We don't have to rename the VM but it might be nice to have the names be consistent. Changing the name is a breaking change of sorts, but it seems better to do it now, while pre v1.0.

### DIFF
--- a/roles/ansible-ssh/tasks/main.yml
+++ b/roles/ansible-ssh/tasks/main.yml
@@ -1,0 +1,27 @@
+---
+- name: Remove prior ssh config info
+  local_action: file path=/tmp/ansible-ssh-config state=absent
+  changed_when: false
+
+- name: Gather ssh config info for Vagrant VM
+  local_action: shell vagrant ssh-config --host {{ inventory_hostname }} >> /tmp/ansible-ssh-config
+  when: "'development' in group_names"
+  register: vagrant_ssh_config
+  failed_when: vagrant_ssh_config.rc == 1
+  changed_when: false
+
+- name: Gather user-specified ssh config files
+  local_action: shell echo "" && cat {{ item }} >> /tmp/ansible-ssh-config
+  with_items: ssh_config_files | default([])
+  changed_when: false
+
+- name: Update ssh config file
+  local_action: command rsync -acv /tmp/ansible-ssh-config roles/ansible-ssh/files/config
+  register: ansible_ssh_config
+  changed_when: "'ansible-ssh-config' in ansible_ssh_config.stdout"
+
+- name: Set connection facts if control machine appears to be a Vagrant VM
+  set_fact:
+    ansible_ssh_host: 127.0.0.1
+    ansible_connection: local
+  when: vagrant_ssh_config.rc is defined and vagrant_ssh_config.rc == 127
--- a/roles/ansible-ssh/tasks/main.yml
+++ b/roles/ansible-ssh/tasks/main.yml
@@ -15,6 +15,10 @@
   with_items: ssh_config_files | default([])
   changed_when: false
 
+- name: Ensure ansible-ssh/files directory exists
+  local_action: file path=roles/ansible-ssh/files state=directory
+  changed_when: false
+
 - name: Update ssh config file
   local_action: command rsync -acv /tmp/ansible-ssh-config roles/ansible-ssh/files/config
   register: ansible_ssh_config
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .vagrant
 vendor/roles
+roles/ansible-ssh/files/config
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Add ansible-ssh role to generate ssh config ([#314](https://github.com/roots/trellis/pull/314))
 * Update SSL cipher suite ([#386](https://github.com/roots/trellis/pull/386))
 * Support for other Vagrant providers (VirtualBox, VMWare, Parallels) ([#340](https://github.com/roots/trellis/pull/340))
 * Specify versions for Ansible Galaxy requirements ([#385](https://github.com/roots/trellis/pull/385))
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -28,6 +28,7 @@ end
 Vagrant.require_version '>= 1.5.1'
 
 Vagrant.configure('2') do |config|
+  config.vm.define 'vagrant-vm'
   config.vm.box = 'ubuntu/trusty64'
   config.ssh.forward_agent = true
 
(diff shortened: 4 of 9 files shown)


austinpray:
This fails for me:

```

PLAY [Prepare Ansible SSH Config] *********************************************

TASK: [ansible-ssh | Remove prior ssh config info] ****************************
ok: [vagrant-vm -> 127.0.0.1]

TASK: [ansible-ssh | Gather ssh config info for Vagrant VM] *******************
ok: [vagrant-vm -> 127.0.0.1]

TASK: [ansible-ssh | Gather user-specified ssh config files] ******************
ok: [vagrant-vm -> 127.0.0.1] => (item=~/.ssh/config)

TASK: [ansible-ssh | Update ssh config file] **********************************
failed: [vagrant-vm -> 127.0.0.1] => {"changed": false, "cmd": ["rsync", "-acv", "/tmp/ansible-ssh-config", "roles/ansible-ssh/files/config"], "delta": "0:00:00.031313", "end": "2015-08-27 20:57:59.052043", "rc": 12, "start": "2015-08-27 20:57:59.020730", "stdout_lines": ["building file list ... done"], "warnings": ["Consider using synchronize module rather than running rsync"]}
stderr: rsync: push_dir#3 "/Users/austinpray/Desktop/trellis/roles/ansible-ssh/files" failed: No such file or directory (2)
rsync error: errors selecting input/output files, dirs (code 3) at /SourceCache/rsync/rsync-45/rsync/main.c(580) [receiver=2.6.9]
rsync: connection unexpectedly closed (8 bytes received so far) [sender]
rsync error: error in rsync protocol data stream (code 12) at /SourceCache/rsync/rsync-45/rsync/io.c(453) [sender=2.6.9]
stdout: building file list ... done

PLAY [WordPress Server - Install LEMP Stack with PHP 5.6 and MariaDB MySQL] ***

GATHERING FACTS ***************************************************************
FATAL: no hosts matched or all hosts have already failed -- aborting


TASK: [common | Validate Ansible version] *************************************
FATAL: no hosts matched or all hosts have already failed -- aborting


PLAY RECAP ********************************************************************
           to retry, use: --limit @/Users/austinpray/dev.retry

vagrant-vm                 : ok=3    changed=0    unreachable=0    failed=1

Ansible failed to complete successfully. Any error output should be
visible above. Please fix these errors and try again.
```

austinpray:
```
diff --git a/roles/ansible-ssh/tasks/main.yml b/roles/ansible-ssh/tasks/main.yml
index bd5d04e..8768674 100644
--- a/roles/ansible-ssh/tasks/main.yml
+++ b/roles/ansible-ssh/tasks/main.yml
@@ -3,6 +3,10 @@
   local_action: file path=/tmp/ansible-ssh-config state=absent
   changed_when: false
 
+- name: Ensure ansible-ssh/files exists 
+  local_action: file path=roles/ansible-ssh/files/config state=directory
+  changed_when: false
+
 - name: Gather ssh config info for Vagrant VM
   local_action: shell vagrant ssh-config --host {{ inventory_hostname }} >> /tmp/ansible-ssh-config
   when: "'development' in group_names"
```

Works for me

austinpray:
>We don't have to rename the VM but it might be nice to have the names be consistent. Changing the name is a breaking change of sorts, but it seems better to do it now, while pre v1.0.

I'm good with this, `vagrant-vm` makes much more sense as to what the name is actually for. `default` is incredibly cryptic.

Also more errors:

image

This is after `vagrant destroy -f && vagrant up`. Investigating.

fullyint:
Thank you @austinpray. I've added your patch. On an earlier version I had committed a default config file to the repo, so its parent directory existed: `roles/ansible-ssh/files`. But I added the config file to the `.gitignore` so, then the `files` directory never made it into the repo. 

I've added the task you suggested, but immediately before the rsync task that would require the directory be present.

roles/ansible-ssh/tasks/main.yml
```diff
@@ -0,0 +1,31 @@
+---
+- name: Remove prior ssh config info
+  local_action: file path=/tmp/ansible-ssh-config state=absent
+  changed_when: false
+
+- name: Gather ssh config info for Vagrant VM
+  local_action: shell vagrant ssh-config --host {{ inventory_hostname }} >> /tmp/ansible-ssh-config
+  when: "'development' in group_names"
```
swalkinshaw: not needed?

fullyint: This task must be ready to skip for staging/production hosts for whom we don't want to run `vagrant ssh-config`. I had to run this ansible-ssh role run in `server.yml` too because once `ansible.cfg` specifies a ssh config file, the file must be present and only that config file will be used. This doesn't mean you'll be without info from your regular `~/.ssh/config`, however, because this new config file will suck in whatever ssh config "includes" files you specify in `ssh_config_files` (defined in `group_vars/all/main.yml`).

fullyint:
**Problem:** The remote-user manual ssh test could inappropriately show that `root` cannot connect under this combination of conditions: 
* a host appears in the ansible hosts file as a nickname only, with no ansible_ssh_host (no IP)
* the host's IP is in a HostName parameter one of the `ssh_config_files` "includes", but not in the default `~/.ssh/config`
* the local action ssh test runs `ssh root@nickname` using default `~/.ssh/config` (no entry for nickname)
* connection fails with stderr `ssh: Could not resolve hostname nickname`

**Resolution:** I added a commit with two changes to resolve this.
* I moved the ansible-ssh role before the remote-user role. This assembles the `ssh_config_files` into one file at `roles/ansible-ssh/files/config`. Now the nickname host's `HostName` (IP) is available. 
* I adjusted the manual ssh `root` connection test to use this custom ssh config file by adding `-F roles/ansible-ssh/files/config`. Instead of adding `-F`, I'd rather just rebase this on top of #321 (if/when it's merged), which seems superior to the manual ssh check.

fullyint:
Closing. The proposed role is cool if and when you need it. However, only a small minority of users would need it. Anyone is welcome to turn it into a galaxy role.

PR #317, opened 2015-08-28
Accommodate file encryption via ansible vault

fullyint:
Moves some variables to `group_vars/<environment>/vault.yml` files.

To start using [Ansible Vault](http://docs.ansible.com/ansible/playbooks_vault.html):
1) copy `.vault_pass.example` to `.vault_pass`, edit the password, and probably `chmod 600 .vault_pass`
2) edit `ansible.cfg`:
```diff
- vault_password_file = .vault_pass.example
+ vault_password_file = .vault_pass
```
3) encrypt `vault.yml` files with command below (doesn't support fileglobs -- ansible/ansible#6241):
```
ansible-vault encrypt group_vars/all/vault.yml group_vars/development/vault.yml group_vars/staging/vault.yml group_vars/production/vault.yml
```

Preliminary discussion in #308 where I mentioned these items:
* encryption is optional
* there are scripts to help one avoid committing unencrypted password files

Would we want `mysql_root_user: root` vaulted? It is currently in `group_vars/all/database.yml`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,14 +64,14 @@ example.com/    - Primary folder for the project
 
 ## Development setup
 
-1. Configure your [WordPress sites](#wordpress-sites) in `group_vars/development/wordpress_sites.yml`
+1. Configure your [WordPress sites](#wordpress-sites) in `group_vars/development/wordpress_sites.yml` and in `group_vars/development/vault.yml`.
 2. Run `vagrant up`
 
 ## Remote server setup (staging/production)
 
 For remote servers, you'll need to have a base Ubuntu 14.04 server already created.
 
-1. Configure your [WordPress sites](#wordpress-sites) in `group_vars/<environment>/wordpress_sites.yml`. Also see the [Passwords docs](https://roots.io/trellis/docs/passwords/).
+1. Configure your [WordPress sites](#wordpress-sites) in `group_vars/<environment>/wordpress_sites.yml` and in `group_vars/<environment>/vault.yml`. See the [Vault docs](https://roots.io/trellis/docs/vault/) for how to encrypt files containing passwords.
 2. Add your server IP/hostnames to `hosts/<environment>`.
 3. Specify public SSH keys for `users` in `group_vars/all/users.yml`. See the [SSH Keys docs](https://roots.io/trellis/docs/ssh-keys/).
 4. Consider setting `sshd_permit_root_login: false` in `group_vars/all/security.yml`. See the [Security docs](https://roots.io/trellis/docs/security/).
@@ -96,9 +96,9 @@ Full documentation: https://roots.io/trellis/docs/deploys/
 
 Before using Trellis, you must configure your WordPress sites.
 
-The `group_vars` directory contains directories for each environment (`development`, `staging`, and `production`). Each environment has its own `wordpress_sites.yml` variables file in [YAML](http://en.wikipedia.org/wiki/YAML) format.
+The `group_vars` directory contains directories for each environment (`development`, `staging`, and `production`). Each environment has its own `wordpress_sites.yml` variables file in [YAML](http://en.wikipedia.org/wiki/YAML) format, with an accompanying `vault.yml` file for sensitive information.
 
-For example: configure the sites on your Vagrant development VM by editing `group_vars/development/wordpress_sites.yml`.
+For example: configure the sites on your Vagrant development VM by editing `group_vars/development/wordpress_sites.yml` and `group_vars/development/vault.yml`.
 
 `wordpress_sites` is the top-level dictionary used to define the WordPress sites, databases, Nginx vhosts, etc that will be created. Each site's variables are nested under a site "key" (e.g., `example.com`). This key is just a descriptive name and serves as the default value for some variables. See our [example project](https://github.com/roots/roots-example-project.com/blob/master/ansible/group_vars/development/wordpress_sites.yml) for a complete working example.
 
@@ -120,7 +120,7 @@ For example: configure the sites on your Vagrant development VM by editing `grou
 * `db_import` - Path to local `sql` dump file which will be imported (optional)
 * `admin_user` - WP admin user name (*development* only, required)
 * `admin_email` - WP admin email address (*development* only, required)
-* `admin_password` - WP admin user password (*development* only, required)
+* `admin_password` - WP admin user password (*development* only, required, in `vault.yml`)
 * `multisite` - hash of multisite options. See the [Multisite docs](https://roots.io/trellis/docs/multisite/).
   * `enabled` - Multisite enabled flag (required, set to `false`)
   * `subdomains` - subdomains option
@@ -135,7 +135,7 @@ For example: configure the sites on your Vagrant development VM by editing `grou
   * `wp_env` - environment (required, matches group name: `development`, `staging`, `production`)
   * `db_name` - database name (required)
   * `db_user` - database username (required)
-  * `db_password` - database password (required)
+  * `db_password` - database password (required, in `vault.yml`)
   * `db_host` - database hostname (default: `localhost`)
   * `domain_current_site` (required if multisite.enabled is `true`)
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
+.vault_pass
 .vagrant
 vendor/roles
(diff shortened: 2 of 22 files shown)


swalkinshaw:
@fullyint no need for `mysql_root_user` to be vaulted.

ansible.cfg
```diff
@@ -1,6 +1,7 @@
 [defaults]
 roles_path = vendor/roles
 force_handlers = True
+vault_password_file = .vault_pass.example
```
swalkinshaw: Shouldn't this be set to `.vault_pass`? It's one less thing to change if users want it enabled.

swalkinshaw:
@fullyint let's move forward with this :+1:

Can you work on a new doc section?

ansible.cfg
fullyint: The playbook fails if `vault_password_file` refers to a missing file. The file `.vault_pass` will be missing by default because it is `.gitignore`d. Users who don't want ansible vault would have to remove this line or create the `.vault_pass` file. I explain below why I didn't just omit this line and require vault users to add it.

To enable vault, we could have users create new stuff: add `vault_password_file` line to `ansible.cfg` and create the `.vault_pass` file. I chose the alternative of having users edit example versions of each (steps 1 and 2 in this PR's first comment) because I speculate that users would be more likely to try vault if they could edit example stuff instead of create new stuff.

An added bonus of having the examples present in the repo is that one can actually skip steps 1 and 2, and do only step 3 (run the `ansible-vault encrypt` command). That means one can test how easy it is to use ansible vault by just running that single command.

swalkinshaw: I had naively hoped that Ansible would just skip using `vault_password_file` if nothing was encrypted so that this could be set to `.vault_pass`.

fullyint: I originally assumed it would skip. That's what I would have preferred.

mxxcon:
I'm sorry if this was discussed somewhere earlier, but why are you completely moving some variables into vaults?
I'm sure many of you already read https://www.reinteractive.net/posts/167-ansible-real-life-good-practices Why not use structure similar to shown there?
For example: in `group_vars/production/wordpress_sites.yml` you remove almost every single record related to database info.
Why not instead keep the current structure but assign variables to their vaulted counterparts:
```
      db_name: {{ vaulted_db_name }}
      db_user: {{ vaulted_db_user }}
      db_password: {{ vaulted_db_password }}
```
And then your `vault.yml` will contain
```
   vaulted_db_name: wp_myblog
   vaulted_db_user: wp_myblog_user
   vaulted_db_password: password123
```

This has the benefit of minimizing change to playbooks and keeps existence of variables more discoverable, while still getting the benefit of ansible-vault. And if a user decided not to use vault, they can just simply replace `{{ }}` values with plain text.

austinpray:
I agree with the approach @mxxcon suggests. You should not have to look at two files to get the full picture.

QWp6t:
:+1:

fullyint:
@mxxcon I like that approach but I couldn't implement it directly because `wordpress_sites` is a hash. 

I mentioned above that there was some preliminary discussion in #308. Here's the [excerpt](https://github.com/roots/trellis/pull/308#issuecomment-133958856) that addresses the specific article you mentioned, and its approach:
>I tried that but ran into trouble with the `wordpress_sites` dict (can't remember enough to explain). However, my "vaulted" variables are prefixed by `vault`. An example is `password="{{ vault.mysql_root_password }}"`. At least that prefix lets you know that the variable is defined in one of the encrypted `vault.yml` files.

#308 then continues with some discussion of hashes.

If I remember, the big challenges is how to loop over the `wordpress_sites` hash when some of the values are vaulted. You'll see how I handled this by searching this PR's diff for `wordpress_sites_vault[item.key]`.

I'd be delighted if anyone can offer a better working implementation.

mxxcon:
I see. I have not fully grokked trellis yet, so I'll defer to you guys since you are more familiar with it. :smiley:

fullyint:
Docs at roots/docs#5. I think this is ready. Thank you, @mxxcon, your comments inspired good revisions.

**A Few Updates**
I made two changes for better discoverability and to provide the full picture in `wordpress_sites.yml`.

1) I switched to using the structure @mxxcon suggested ([pseudo leaf](https://www.reinteractive.net/posts/167-ansible-real-life-good-practices), [recommended by ansible](http://docs.ansible.com/ansible/playbooks_best_practices.html#variables-and-vaults)) for scalar variables (not used in loops), e.g., `mail_password: "{{ vault_mail_password }}"`.

2) For the hash variables (used in loops), I added their names in comments where the variables would have appeared in the hash if left unencrypted. That way the unencrypted file shows the full picture and the variables still show up in searches.
```yaml
wordpress_sites:
  example.com:
    ⋮
    env:
      wp_home: http://example.dev
      wp_siteurl: http://example.dev/wp
      wp_env: development
      # Define the following variables in group_vars/development/vault.yml
      # db_name:
      # db_user:
      # db_password:
      ⋮
```

One other change worth noting was that I removed the `.vault_pass.example` file and the related example default in `ansible.cfg`, changing my position on [this decision](https://github.com/roots/trellis/pull/317#discussion_r42829186).

**Vault with Variables in a Hash**
I believe I've used a good strategy for handling the encrypted variables from the `wordpress_sites` hash. I don't believe the pseudo leaf structure can be applied effectively to a hash that will be used in tasks that loop. Here are a few examples of failed or suboptimal implementations of the pseudo leaf structure to the `wordpress_sites` hash.

```yaml
# 1 - Fail: If all sites use `vault_db_password`, they all have the same password
wordpress_sites:
  site-1.com:
    db_password: {{ vault_db_password }}
  site-2.com:
    db_password: {{ vault_db_password }}

# 2 - Suboptimal: Requiring users to edit variable names means more work and more risk for error
wordpress_sites:
  site-1.com:
    db_password: {{ vault_db_password_1 }}
  site-2.com:
    db_password: {{ vault_db_password_2 }}

# 3 - Fail: The `item.key` variable is only defined in the scope of the looping task and will be undefined in the more global scope of the example below
wordpress_sites:
  site-1.com:
    db_password: "{{ vault_wordpress_sites[item.key].env.db_password }}"
  site-2.com:
    db_password: "{{ vault_wordpress_sites[item.key].env.db_password }}"
```

README.md
```diff
@@ -117,9 +117,9 @@ For example: configure the sites on your Vagrant development VM by editing `grou
 * `site_title` - WP site title (*development* only, default: project name)
 * `db_create` - whether to auto create a database or not (default: `true`)
 * `db_import` - Path to local `sql` dump file which will be imported (optional)
-* `admin_user` - WP admin user name (*development* only, required)
-* `admin_email` - WP admin email address (*development* only, required)
-* `admin_password` - WP admin user password (*development* only, required)
+* `admin_user` - WP admin user name (*development* only, required, in `vault.yml`)
+* `admin_email` - WP admin email address (*development* only, required, in `vault.yml`)
```
swalkinshaw: `admin_email` shouldn't be in vault

README.md
```diff
@@ -132,9 +132,9 @@ For example: configure the sites on your Vagrant development VM by editing `grou
   * `wp_home` - `WP_HOME` constant (required)
   * `wp_siteurl` - `WP_SITEURL` constant (required)
   * `wp_env` - environment (required, matches group name: `development`, `staging`, `production`)
-  * `db_name` - database name (required)
-  * `db_user` - database username (required)
-  * `db_password` - database password (required)
+  * `db_name` - database name (environment variable, required, in `vault.yml`)
```
swalkinshaw: ditto for `db_name` and `db_user`?

README.md
swalkinshaw: `admin_user` as well?

mxxcon:
@swalkinshaw why not? Do you not consider these to be sensitive?

Speaking of sensitive info, unless I'm misunderstanding this, is it proper for `group_vars/all/vault.yml` to contain `vault_sudoer_passwords.admin` variable? Shouldn't it be unique per environment(dev/stage/prod), if not even unique per host?

swalkinshaw:
@mxxcon not really. Keep in mind all of this data should still be kept in private repos. I only consider "secrets" as applicable for vault.

I agree that `vault_sudoer_passwords` should probably be env dependent.

fullyint:
As @swalkinshaw suggested, I removed a few vars from the vault: 
* WP's `admin_user` and `admin_email`
* env vars: `db_user` and `db_name`

As @mxxcon suggested, I split out `vault_sudoer_passwords` into the separate `group_vars/<environment>/vault.yml` files.

I updated the [docs PR](https://github.com/roots/docs/pull/5) to reflect the above changes.

Issue #305, opened 2015-08-16
Deploy command hooks


swalkinshaw:
Right now our deploy command "hooks" are limited to strings that get passed into the `command` or `shell` module. This means we can't really take advantage of Ansible modules at all.

This leads to issues like https://github.com/roots/trellis/pull/282 with non-ideal solutions.

Instead we could include task files for each hook:

```yaml
# roles/deploy/tasks/main.yml
- name: Run post_finalize_tasks
  include: hooks/post-deploy.yml
```

```yaml
# roles/deploy/tasks/hooks/post-deploy.yml
- name: Restart php-fpm
  shell: sudo php5-fpm reload
  when: not hhvm

- name: Restart hhvm
  shell: sudo hhvm restart
  when: hhvm
```

/cc @CFXd @louim @fullyint

louim:
That is an excellent Idea. Moreover, `project_deploy_module` got a refactor and is got renamed to `project_deploy`: https://github.com/f500/ansible-project_deploy. They now include hook like you propose. With a little refactor we could use theses. 

Quick example from the docs: `project_deploy_hook_on_perform_build: "{{ playbook_dir }}/deploy_hooks/perform-build.yml"`

swalkinshaw:
Wow that's hilarious. They even named the dir `hooks`. I was just thinking earlier I should see if that project had been updated since I hadn't looked in a long time.

swalkinshaw:
Fixed by https://github.com/roots/trellis/pull/378

PR #320, opened 2015-08-30
Deploy Hooks: use task files

swalkinshaw:
This implements proper hook files with Ansible tasks as discussed in https://github.com/roots/trellis/issues/305.

This feature is heavily borrowed/stolen from https://github.com/f500/ansible-project_deploy with a few differences.

This PR separates outs the `main.yml` tasks file into multiple files. Each file represents a "step" in the deploy flow. And for each step we have `before` and `after` hooks available.

The steps are:

1. `initialize`
2. `update`
3. `prepare`
4. `build`
5. `share`
6. `finalize`

Also note that technically the `after` hook of one step and the `before` hook of the following step are the same. However I wanted to include the `before`/`after` nomenclature for semantic reasons mostly. Also so it's more obvious *when* the hook is run. The f500 project uses hooks like `project_deploy_hook_on_initialize` which, in my opinion, could be a little confusing as to where/when exactly it's happening.

Hooks available:

* `deploy_initialize_before`
* `deploy_initialize_after`
* `deploy_update_before`
* `deploy_update_after`
* `deploy_prepare_before`
* `deploy_prepare_after`
* `deploy_build_before`
* `deploy_build_after`
* `deploy_share_before`
* `deploy_share_after`
* `deploy_finalize_before`
* `deploy_finalize_after`

Those are all variables which can be defined to override the default. Right now only `deploy_build_after` and `deploy_finalize_after` have corresponding hook files to replicate functionality we previously had in the commands.

### Todo

- [x] better naming
- [x] testing
- [x] docs (see https://github.com/roots/docs/pull/1)

### DIFF
--- a/group_vars/all/logrotate.yml
+++ b/group_vars/all/logrotate.yml
@@ -17,4 +17,4 @@ logrotate_scripts:
         if [ -d /etc/logrotate.d/httpd-prerotate ]; then \
               run-parts /etc/logrotate.d/httpd-prerotate; \
             fi \
-      postrotate: invoke-rc.d nginx rotate >/dev/null 2>&1
+      postrotate: service nginx rotate


swalkinshaw:
@louim any thoughts on this?

louim:
@swalkinshaw, looks great to me, it will really help making the deploy process more customizable (and easier to maintain). :+1:

Issue #285, opened 2015-08-02
Non Root User on AWS Errors on Provisioning server.yml


BrandonShutter:
I have an Ubuntu box on AWS.  Setting `admin_user: ubuntu` in `group_vars/all` results in an error when the `Determine remote user task` runs.

Rather than Trellis trying to use `ubuntu` over `root` it errors out with ssh garbage.

Here's the output of `-vvvv`:

```
Brandons-MacBook-Pro:ansible shutter$ ansible-playbook -i hosts/production server.yml -vvvv
```

    PLAY [Determine Remote User] ************************************************** 
    
    TASK: [remote-user | Determine whether to connect as root or admin_user] ****** 
    <127.0.0.1> REMOTE_MODULE command ssh -o PasswordAuthentication=no  "echo root" || echo ubuntu #USE_SHELL
    <127.0.0.1> EXEC ['/bin/sh', '-c', 'mkdir -p $HOME/.ansible/tmp/ansible-tmp-1438381915.22-77758919300933 && chmod a+rx $HOME/.ansible/tmp/ansible-tmp-1438381915.22-77758919300933 && echo $HOME/.ansible/tmp/ansible-tmp-1438381915.22-77758919300933']
    <127.0.0.1> PUT /var/folders/lf/878x4xqx5q71w08fpj496mnh0000gn/T/tmpnwxIKk TO /Users/shutter/.ansible/tmp/ansible-tmp-1438381915.22-77758919300933/command
    <127.0.0.1> EXEC ['/bin/sh', '-c', u'LANG=en_US.UTF-8 LC_CTYPE=en_US.UTF-8 /usr/bin/python /Users/shutter/.ansible/tmp/ansible-tmp-1438381915.22-77758919300933/command; rm -rf /Users/shutter/.ansible/tmp/ansible-tmp-1438381915.22-77758919300933/ >/dev/null 2>&1']
    ok: [52.3.121.253 -> 127.0.0.1] => {"changed": false, "cmd": "ssh -o PasswordAuthentication=no root@52.3.121.253 \"echo root\" || echo ubuntu", "delta": "0:00:11.073787", "end": "2015-07-31 18:32:06.360790", "rc": 0, "start": "2015-07-31 18:31:55.287003", "stderr": "", "stdout": "Please login as the user \"ubuntu\" rather than the user \"root\".", "stdout_lines": ["Please login as the user \"ubuntu\" rather than the user \"root\"."], "warnings": []}
    
[...]

PR #321, opened 2015-08-30
Use ansible ping module to determine remote user

fullyint:
Thanks @louim for [pointing out](https://github.com/roots/trellis/issues/288#issuecomment-127487221) that if a user's hosts file were to use a nickname like this... 
```
[web]
nickname ansible_ssh_host=192.168.50.5
```
... the remote-user task would try `ssh root@nickname` when it should really use the `ansible_ssh_host` value if it exists, with the `ansible_hostname` as the fallback default.

Thinking about this, I realized the hosts file could have other important parameters to use too, such as `ansible_ssh_port`, `ansible_ssh_private_key_file`, etc. 

Rather than try to add in all these parameters to a local action ssh command, I realized we could just use the [ping module](http://docs.ansible.com/ansible/ping_module.html) which tests whether Ansible can connect to each host in the exact manner used for `ansible-playbook` commands (I think).

This should be a more reliable approach. It should also help avoid problems like [this](https://discourse.roots.io/t/fatal-192-168-50-5-ssh-error-permission-denied-publickey/4566/9), which resulted from the task testing a manual ssh connection that is slightly different from the connection Ansible would use on its own.

Successfully handles the AWS issue raised in #285

### DIFF
--- a/roles/remote-user/tasks/main.yml
+++ b/roles/remote-user/tasks/main.yml
@@ -1,10 +1,14 @@
 ---
 - name: Determine whether to connect as root or admin_user
-  local_action: shell ssh -o PasswordAuthentication=no root@{{ inventory_hostname }} "echo can_connect"
+  local_action: command ansible {{ inventory_hostname }} -m ping -i {{ inventory_file }} -u root
   failed_when: false
   changed_when: false
   register: root_status
 
 - name: Set remote user for each host
   set_fact:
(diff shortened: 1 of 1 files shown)


swalkinshaw:
Good idea :+1:

austinpray:
Tested and working for provisioning a remote server and for Vagrant

PR #322, opened 2015-08-30
Add more HTTP security headers

swalkinshaw:
* `X-Xss-Protection` (https://scotthelme.co.uk/hardening-your-http-response-headers/#x-xss-protection)
* `X-Content-Type-Options` (https://scotthelme.co.uk/hardening-your-http-response-headers/#x-content-type-options)

The value here for `X-Xss-Protection` has some issues in IE8 but who cares?

We also need to look into CSP headers but that's more complicated.

### DIFF
--- a/roles/wordpress-setup/templates/wordpress-site.conf.j2
+++ b/roles/wordpress-setup/templates/wordpress-site.conf.j2
@@ -35,7 +35,9 @@ server {
   include ssl-stapling.conf;
 
   add_header Strict-Transport-Security "max-age=31536000; includeSubdomains; preload";
+  add_header X-Content-Type-Options "nosniff" always;
   add_header X-Frame-Options SAMEORIGIN;
+  add_header X-Xss-Protection "1; mode=block" always;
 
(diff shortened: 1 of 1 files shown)


louim:
I like this!

austinpray:
https://www.owasp.org/index.php/List_of_useful_HTTP_headers and http://blogs.msdn.com/b/ie/archive/2008/07/02/ie8-security-part-iv-the-xss-filter.aspx

>This header enables the Cross-site scripting (XSS) filter built into most recent web browsers. It's usually enabled by default anyway, so the role of this header is to re-enable the filter for this particular website if it was disabled by the user. This header is supported in IE 8+, and in Chrome (not sure which versions). The anti-XSS filter was added in Chrome 4. Its unknown if that version honored this header.

So we don't need to specify `X-Xss-Protection`

swalkinshaw:
@austinpray I can't find too much information on this beyond that Chrome and Safari enable it by default. But I'm not sure if their default has "block".

austinpray:
I'll look into it

PR #327, opened 2015-09-01
hhvm restart IO redirection typo should be 2>&1

jeffbski:
For the hhvm restart the IO redirection should be 2>&1

Current command raises error: /bin/sh: 1: 1: not found


swalkinshaw:
Thanks that was a silly error.

PR #329, opened 2015-09-06
Improve subtree variable name and documentation

swalkinshaw:
Renamed `subtree` to `subtree_path` to better reflect what it actually is. `subtree` also wasn't even documented in the `README`.


austinpray:
git-subtree uses "prefix" instead of "path".

What about dropping "subtree" entirely and making it something more intuitive?

PR #332, opened 2015-09-06
Update wordpress_sites.yml

BrandonShutter:
Same order as https://api.wordpress.org/secret-key/1.1/salt/


swalkinshaw:
Want to update the staging file as well?

swalkinshaw:
Oh I see

PR #333, opened 2015-09-06
Update Keys to match https://api.wordpress.org/secret-key/1.1/salt/

BrandonShutter:
Update order of keys to match https://api.wordpress.org/secret-key/1.1/salt/


swalkinshaw:
Thanks!

PR #334, opened 2015-09-06
Validate subtree_path during deploy

swalkinshaw:
Setting a proper `subtree_path` seems to be a common problem. Previously
the error was a fairly cryptic one during the `mv` command.

Now we'll check if the `subtree_path` actually exists in the cloned repo
and provide a useful fail msg if it does not.


fullyint:
This works in my tests. That custom error message will be helpful.

PR #339, opened 2015-09-07
Remove HHVM MySQL socket workaround

swalkinshaw:
Originally added in #145 to fix an HHVM issue which has been fixed as of 3.6.1.


swalkinshaw:
See https://github.com/roots/trellis/pull/343

PR #340, opened 2015-09-07
Vagrant provider agnostic setup

ckovey:
This allows Trellis to be provider agnostic without losing any functionality.  I've tested on Windows 8.1 with VMWare Workstation 11 using Vagrant 1.7.4

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,9 +1,6 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
-# Specify Vagrant provider as Virtualbox
-ENV['VAGRANT_DEFAULT_PROVIDER'] = 'virtualbox'
-
 require 'yaml'
 
 ANSIBLE_PATH = __dir__ # absolute path to Ansible directory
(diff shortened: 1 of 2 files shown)


Vagrantfile
```diff
@@ -75,13 +71,17 @@ Vagrant.configure('2') do |config|
     end
   end
 
+  # Give VM access to all cpu cores on the host
+  cpus = case RbConfig::CONFIG['host_os']
+           when /darwin/ then `sysctl -n hw.ncpu`.to_i
+           when /linux/ then `nproc`.to_i
+           else 2
+         end
```
swalkinshaw: can you restore this formatting/indentation to what it was?

Vagrantfile
```diff
@@ -91,6 +91,27 @@ Vagrant.configure('2') do |config|
...
+    # Specify provider box
+    config.vm.box = 'puppetlabs/ubuntu-14.04-64-nocm'
+
+    # Customize memory in MB
+    vmw.vmx["memsize"] = 1024
+    vmw.vmx["numvcpus"] = cpus
+  end
+
+  # Parallels settings
+  config.vm.provider "parallels" do |prl|
```
swalkinshaw: single quotes here and above :)

swalkinshaw:
@ckovey thanks for this. Just had a couple minor style issues. Would you be able to squash commits once it's ready?

ckovey:
Fixed style issues and squashed everything up, sorry about that @swalkinshaw !

austinpray:
Wanna add VMware fusion while we are at it? I have a license I can test with.

ckovey:
@austinpray would you mind trying it as is?  Supposedly the different VMware provider names were unified into `vmware_desktop`

austinpray:
@ckovey I see! I'll test it tomorrow then

austinpray:
Something isn't right:

image

The virtualbox provider should be using `ubuntu/trusty64` right?

ckovey:
I changed the box url to use the override way of doing things in their example: https://docs.vagrantup.com/v2/providers/configuration.html.  

Then it seemed like `vmware_desktop` only applies to boxes and not Vagrant configs as my windows machine would try and load the virtualbox `ubuntu/trusty64` using the override method.  I guess somehow before it was simply taking the last `config.vm.box` defined as that would explain how mac was still trying to load the puppetlabs box.  I added `vmware_workstation` and `vmware_fusion` blocks.

Finally DRY'd the memory setting out to be with CPU setting.

Tested this on my Win 8 VMware and MBP with Virtualbox.  I don't have fusion if you want to try again

swalkinshaw:
@ckovey looks great. We'll eventually need to add some documentation that we support other providers now but this is still good for now.

swalkinshaw:
@ckovey could you rebase this? should be good to go after that.

ckovey:
@swalkinshaw rebased, thanks.  Looks like someone figured out how to get Windows CPU count, nice!

merchantguru:
Does this mean Trellis can work with Parallels?

swalkinshaw:
@merchantguru it should.

merchantguru:
Parallels works with a few tweaks. Errors I got:  

1) "The box you're attempting to add doesn't support the provider you requested. Please find an alternate box...."

Using `parallels/ubuntu-14.04` instead of `ubuntu/trusty64` worked.

2) "The specified host network collides with a non-hostonly network! This will cause your specified IP to be inaccessible. Please change the IP or name of your host only network so that it no longer matches that of a bridged or non-hostonly network."

I changed 
`  config.vm.network :private_network, ip: '192.168.50.5'` to 
`  config.vm.network :private_network, ip: '192.168.1.200'`

Not sure if you'd want to make a change like that to the default Trellis Vagrantfile or what.

ckovey:
Thanks @merchantguru, I did not have Parallels to test so I just made an assumption with the box.  I've updated the commit with the `parallels/ubuntu-14.04` base you found if you want to try again fresh.

re the IP error.  This I believe is because you have multiple Trellis installations going or some other device/VM on that IP in your network.  The master version of `Vagrantfile` specified the IP https://github.com/roots/trellis/blob/master/Vagrantfile and I have not changed that here.  You should get the same error if you closed the master repo twice and tried to `vagrant up` on each.  So I believe this part would be "out of scope" for this PR

swalkinshaw:
@merchantguru awesome, thanks for testing.

I agree about the networking error. Unless that error would happen on a standard first Trellis install then we shouldn't worry about it.

merchantguru:
OK @ckovey, @swalkinshaw, I've retested and Parallels works flawlessly. Thanks for making this provider agnostic!

swalkinshaw:
@ckovey thanks for this!

PR #344, opened 2015-09-09
Use proper booleans for sshd role

swalkinshaw:
The template should handle this conversion to make it easier on users.


swalkinshaw:
Docs: https://github.com/roots/docs/pull/2

louim:
:tada: :+1: 

This will be much cleaner!

PR #345, opened 2015-09-09
Ensure admin_user can connect before disabling root

### LINKED ISSUE #294: Users role should fail if server has no keys
Born out of the discussion from https://github.com/roots/trellis/pull/293

>
we could finish the users role with a fail task causing server.yml to fail when all these conditions are met:
>
- not vagrant vm
- sshd_permit_root_login = "no"
- server has no key for admin_user

If for instance the user forgets to configure their SSH keys or their SSH keys do not resolve they should not be locked out of the server.

This is a low-priority enhancement.

fullyint:
Resolves #294 

Only merge after #344 which converts `sshd_permit_root_login` to boolean.

### DIFF
--- a/roles/users/tasks/main.yml
+++ b/roles/users/tasks/main.yml
@@ -9,6 +9,15 @@
               line="%sudo ALL=(ALL:ALL) ALL"
               validate="/usr/sbin/visudo -cf %s"
 
+- name: Fail if root login will be disabled but admin_user will not be a sudoer
+  assert:
+    that:
+      - "{{ admin_user in (users | map(attribute='name') | list) }}"
+      - "{% for item in users if item.name == admin_user %}{{ 'sudo' in item.groups }}{% endfor %}"
+      - "{{ admin_user in sudoer_passwords.keys() }}"
+    msg: "When `sshd_permit_root_login: false`, you must add `sudo` to the `groups` for admin_user (in `users` hash), and set a password for admin_user in `sudoer_passwords`. Otherwise Ansible could lose the ability to run the necessary sudo commands."
+  when: not sshd_permit_root_login
+
 - name: Setup users
   user: name="{{ item.name }}"
         group="{{ item.groups[0] }}"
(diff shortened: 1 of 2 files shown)


roles/users/tasks/main.yml
```diff
@@ -34,3 +34,24 @@
...
+  failed_when: false
+  changed_when: false
+  sudo: no
+  register: admin_user_status
+  when: not sshd_permit_root_login
+
+- name: Fail if root login will be disabled but admin_user cannot connect
+  fail:
+    msg: 'Halting playbook before disabling root login (`sshd_permit_root_login: false`) because the admin_user appears unable to connect. Ensure that the admin_user appears in your `users` hash with a valid entry for `keys`.'
+  when: not sshd_permit_root_login and admin_user_status.rc != 0
```
swalkinshaw:
```yaml
when: not sshd_permit_root_login and admin_user_status|failed
```

Think that's the same

fullyint: Nice. Thanks. I'll use that instead.

fullyint:
I added a commit with adjustments to handle this scenario:

Imagine root login was disabled on a previous run of `server.yml`. Now someone removes `sudo` from their admin_user's `groups` and reprovisions. The "Setup users" task would remove the admin_user from the `sudo` group and now Ansible has no sudoers. To prevent that, we need to verify that `sudo` is in `groups` (and potentially fail) before running "Setup users". 

However, the check for whether admin_user can connect to the server must be run later, after the "Add SSH keys" task.

So, our checks for `sudo` and general connectivity are necessarily separated in the users role.

Note that it might be more convenient to users if the `sudo` group check were performed sooner, perhaps in the `common` role, instead of unnecessarily waiting so far into the playbook to check and fail. Conceptually, however, the check fits in the `users` role, so that's where I left it.

swalkinshaw:
@fullyint squash + CHANGELOG update?

Looks good :+1:

fullyint:
updated CHANGELOG and squashed

PR #347, opened 2015-09-10
Update README.md

dagobertrenouf:
Seems to me this was forgotten when changing the name of the projet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ You should now have the following directories at the same level somewhere:
 
 ```
 example.com/    - Primary folder for the project
-├── ansible/    - Your version of this repo (renamed to just `ansible`)
+├── trellis/    - Your clone of this repo (renamed to just `trellis`)
 └── site/       - A Bedrock-based site (suggested to name this the generic `site` since your project name is already at the top level)
 ```
 


swalkinshaw:
It wasn't named `ansible` because the project was `bedrock-ansible`.

But this does bring up a good point of whether this should be renamed to `trellis` anyway.

dagobertrenouf:
I'm coming from using bedrock-ansible previously, it jumped on me like as something that had been forgotten while changing names.

dagobertrenouf:
I remember in an earlier version of the repo the instructions said:

You should now have the following directories at the same level somewhere:
`- bedrock-ansible/
- example.dev/`

I think that's why I associate the folder name with the name of the repo.

retlehs:
if it's changed to `trellis` then it seems like `site` should be renamed to `bedrock`

i say keep it as is right now

mAAdhaTTah:
@retlehs If that's the case, maybe changing the name to `server` (so it's `server` & `site`) might make more sense?

README.md
```diff
@@ -49,7 +49,7 @@ You should now have the following directories at the same level somewhere:
 
 ```
 example.com/    - Primary folder for the project
-├── ansible/    - Your version of this repo (renamed to just `ansible`)
+├── trellis/    - Your version of this repo (renamed to just `trellis`)
```
austinpray: While we are at it: 

"Your version of this repo (renamed to just `trellis`)" :arrow_right: "Your clone of this repo"

swalkinshaw: @dagobertrenouf are you able to update this?

dagobertrenouf: Yep, will do over the weekend

dagobertrenouf: Will only be able to do this week after all.

dagobertrenouf:
@swalkinshaw added @austinpray suggestion. Let me know if I need to do something more (like squish all commits in one? I'm not sure).

swalkinshaw:
@dagobertrenouf thanks. Squash would be great :+1:

dagobertrenouf:
@swalkinshaw done ;)

swalkinshaw:
Thanks!

Issue #349, opened 2015-09-11
WP-CLI Error during "Install WP" step of Wordpress Install role


redconfetti:
Receiving this error when building my local VM using Trellis + Bedrock:
```
Fatal error: File not found: ../vendor/autoload.php in phar://wp-cli.phar/php/wp-cli.php(23)(3415a90e53dd43920967a079a985ac64) : eval()'d code on line 7
```

I'd get this issue even when I log into the VM and try to run `wp --help` from the 'current' directory.

```
vagrant@guestvm:/srv/www/mysite.dev/current$ wp --help

Fatal error: File not found: ../vendor/autoload.php in phar://wp-cli.phar/php/wp-cli.php(23)(3415a90e53dd43920967a079a985ac64) : eval()'d code on line 7
```

This doesn't occur from the `current/web` directory.
```
vagrant@guestvm:/srv/www/mysite.dev/current$ cd web
vagrant@guestvm:/srv/www/mysite.dev/current/web$ wp --help
Error: The site you have requested is not installed.
Run `wp core install`.
```

I'll file a pull request to fix [this directory path](https://github.com/roots/trellis/blob/master/roles/wordpress-install/tasks/main.yml#L34).

retlehs:
something failed during your provision - this shouldn't be necessary. unable to reproduce on latest trellis (master)

```
vagrant@example:/srv/www/example.com/current$ wp option get home
http://example.dev
```

please open a thread on [our discourse](https://discourse.roots.io/) if you still have problems figuring this out

redconfetti:
@retlehs - You're right. I was using an [outdated version](https://github.com/roots/bedrock/blob/f5a85d6d/web/wp-config.php) of [bedrock/web/wp-config](https://github.com/roots/bedrock/blob/125d03af/web/wp-config.php). Thanks for pointing that out.

```
vagrant@guesthost:/srv/www/mysite.dev/current$ wp option get home
http://enneagraminstitute.dev
```

I'd recently copied over changes by hand for the files, must have missed that.

PR #350, opened 2015-09-11
Updated chdir path for WP-CLI to conform to new Bedrock web root

redconfetti:
See #349


retlehs:
something failed during your provision - this shouldn't be necessary. unable to reproduce on latest trellis (master)

```
vagrant@example:/srv/www/example.com/current$ wp option get home
http://example.dev
```

please open a thread on [our discourse](https://discourse.roots.io/) if you still have problems figuring this out

PR #356, opened 2015-09-17
Fix #352 - Block PHP scripts in Nginx for uploads

### LINKED ISSUE #352: wordpress uploads folder capable of executing php scripts
if a php script (or some other script) makes it into the uploads folder, the script can be run.

example.com/app/uploads/2015/09/test.php can be run

i am capable of fixing this and submitting a pull request, i am just not sure which file to do this in. my gut says in the wordpress.conf file, but i was trying to find where the 403 is returned when trying to access the uploads folder directly e.g. example.com/app/uploads/2015/09/ OR example.com/app/uploads returns a 403, but im not sure where that is set. i can do it either place!

erikbelusic:
Prevent php scripts from being accessed or executed inside the app/uploads folder and any subdirectories by adding an nginx directive to the wordpress.conf.

This has been tested to return a 403 error for any file with a ```.php``` extension in app/uploads or any subdirectory of uploads. Assets with any other extension are not affected and can be accessed as normal.

### DIFF
--- a/roles/nginx/templates/wordpress.conf.j2
+++ b/roles/nginx/templates/wordpress.conf.j2
@@ -1,5 +1,10 @@
 # {{ ansible_managed }}
 
+# Prevent PHP scripts from being executed inside the uploads folder.
+location ~* /app/uploads/.*\.php$ {
+  deny all;
+}
+
 location / {
   try_files $uri $uri/ /index.php?$args;
 }


roles/nginx/templates/wordpress.conf.j2
```diff
@@ -1,5 +1,10 @@
 # {{ ansible_managed }}
 
+# Prevent php scripts from being executed inside the uploads folder.
+location ~* /app/uploads/.*\.php$ {
+    deny all;
```
retlehs: use 2 spaces instead of 4

roles/nginx/templates/wordpress.conf.j2
```diff
@@ -1,5 +1,10 @@
 # {{ ansible_managed }}
 
+# Prevent php scripts from being executed inside the uploads folder.
```
retlehs: tiny nitpick, but while you're at it please change PHP to uppercase

erikbelusic:
done and done! i shouldn't have did this on my master branch though....right? i should have made a feature branch?

swalkinshaw:
@erikbelusic yeah you probably have created a branch but it's not a big deal. You can always re-create your fork.

One more thing: we usually like to keep PRs to a single commit (unless there's a reason for more) to keep our Git history clean. Would you be able to squash these two commits into one? We can provide instructions if you haven't done it before.

Although I just realized that's the best reason to do work on a feature branch since you need to force push after a rebase/squash and obviously you should avoid that on `master` :)

erikbelusic:
@swalkinshaw i have not done it before. please provide guidance =]

erikbelusic:
@swalkinshaw i think i got it. let me know if it turned out correctly.

swalkinshaw:
@erikbelusic perfect! Thanks.

erikbelusic:
@swalkinshaw  now that this has been merged, can i delete and refork to make any additional contributions on feature branches without messing this up?

swalkinshaw:
@erikbelusic yep I'd definitely do that.

louim:
@erikbelusic No need to delete.

Assuming you named the upstream (this remote) `upstream`:

```
git fetch --all
git reset --hard upstream/master
git push -f
```

PR #357, opened 2015-09-18
Add sudoer_passwordless option.

chriszarate:
Not sure if there's a philosophical objection to this, but I need it for my environment.

### DIFF
--- a/roles/users/tasks/main.yml
+++ b/roles/users/tasks/main.yml
@@ -6,7 +6,7 @@
   lineinfile: dest=/etc/sudoers
               state=present
               regexp="^%sudo"
-              line="%sudo ALL=(ALL:ALL) ALL"
+              line="%sudo ALL={{ sudoer_passwordless | ternary('(ALL) NOPASSWD:', '(ALL:ALL)') }} ALL"
               validate="/usr/sbin/visudo -cf %s"
 
 - name: Fail if root login will be disabled but admin_user will not be a sudoer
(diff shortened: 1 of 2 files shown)


BrandonShutter:
I would use this.

swalkinshaw:
@chriszarate yeah I'm not sure about this. I might have a philosophical objection but also concerned about adding options like this vs a single way of doing it.

chriszarate:
Fair enough. Feel free to close.

BrandonShutter:
Humor me, please.  Why is this a bad idea?  I personally have been using this.

swalkinshaw:
@BrandonShutter I personally find typing the password a strong reminder for what is actually going on.

Might not actually be a valid reason though.

roles/users/tasks/main.yml
```diff
@@ -6,7 +6,7 @@
   lineinfile: dest=/etc/sudoers
               state=present
               regexp="^%sudo"
-              line="%sudo ALL=(ALL:ALL) ALL"
+              line="%sudo ALL={{ sudoer_passwordless | ternary('(ALL) NOPASSWD:', '(ALL:ALL)') }} ALL"
```
swalkinshaw: @chriszarate just realized an alternative solution I think might be better. Just have a variable called `sudo_group_permissions` which defaults to `(ALL:ALL)` and then you could customize it to `(ALL) NOPASSWD:`

chriszarate:
Nice! That's much more flexible than the Boolean and ternary. I'll revise.

swalkinshaw:
Closing this for now. @chriszarate feel free to re-open whenever you get a chance to work on this again.

PR #358, opened 2015-09-18
Skip connection check if playbooks are being run locally.

chriszarate:
The major use case for running playbooks locally is Packer builds. For a variety of reasons, Packer's Ansible provisioner uploads playbooks to the VM and runs them with `--connection=local`. So the recent addition of `local_action`s to check the user's connection causes the playbooks to fail.

There are other `local_action`s in the `deploy` and `remote_user` roles, but they are only used in the `deploy.yml` and `server.yml` playbooks, so less likely to cause an issue.

### DIFF
--- a/roles/users/tasks/main.yml
+++ b/roles/users/tasks/main.yml
@@ -50,7 +50,7 @@
   changed_when: false
   sudo: no
   register: admin_user_status
-  when: not sshd_permit_root_login
+  when: not sshd_permit_root_login and ansible_connection | default('ssh') != 'local'
 
 - name: Fail if root login will be disabled but admin_user cannot connect
   fail:


roles/users/tasks/main.yml
```diff
@@ -50,7 +50,7 @@
   changed_when: false
   sudo: no
   register: admin_user_status
-  when: not sshd_permit_root_login
+  when: not sshd_permit_root_login and inventory_hostname != '127.0.0.1'
```
fullyint: Consider an Ansible inventory like this:
```
[web]
nickname ansible_ssh_host=127.0.0.1
``` 
The variable `inventory_hostname` would be `nickname` and wouldn't be captured by your addition to the `when:` conditional. You could check `ansible_ssh_host` first, and fallback to `inventory_hostname`. [Credit here](https://github.com/louim/bedrock-ansible/commit/4b4ff3365b254eaa941c37febf3fbc673f8b5c51#diff-db0d022f251967eca56175a8e2023698R3).

chriszarate: Thanks! Checking whether `ansible_ssh_host` is defined should be enough—it isn't when playbook are running locally.

chriszarate:
Thanks to an assist from @fullyint I think this is good. Checking for `ansible_ssh_host` to be defined is pretty innocuous. If it's not defined, it makes sense to ignore a variable about Ansible's SSH connectivity.

This will help anyone else who might be combining Trellis with Packer—for, say, AMIs for dev instances or immutable deployments.

roles/users/tasks/main.yml
```diff
@@ -50,7 +50,7 @@
   changed_when: false
   sudo: no
   register: admin_user_status
-  when: not sshd_permit_root_login
+  when: not sshd_permit_root_login and ansible_ssh_host is defined
```
fullyint: If staging or production hosts files are like the Trellis default files and list only IP and do not explicitly define `ansible_ssh_host`, I'm pretty sure `ansible_ssh_host is defined` will return `false`. If so, this task will sometimes not run for staging/production but should. Does that sound right?

Anyone aware of an ansible variable to check whether `--connection=local` was passed? That strikes me as a more direct test of the issue at hand. @chriszarate What specifically fails when Packer runs this? Is it that the ping module run on a local connection always fails? Just fails for admin user? Can we test for that specific problem instead of just testing local connection (but perhaps local connection is the exact problem)?

chriszarate: For clarity (this was confusing to me initially): Ansible's `--connection=local` flag means that the playbook is run on the local machine (i.e., on your Mac, were you so inclined). There is no SSH involved at all. In the context of Packer, this means that the playbooks, roles, group_vars, host_vars, etc., are uploaded to the guest machine (whether local VM or cloud) and run on the guest with `--connection=local`. This task fails on Packer because we are not connecting via SSH.

Huge thanks for pointing out that defining hosts by IP means that `ansible_ssh_host` will be undefined. That was unexpected! And, _of course_, a different search term of the docs turns up `ansible_connection`. It is defined when the `--connection` flag is passed. Sheesh. I'll push an update.

fullyint: When I first saw your PR a while back I placed my hope in being able to test the `ansible_connection` variable, but in my tests it never seemed defined unless the hosts file (inventory file) explicitly declared it for a host (e.g., `127.0.0.1 ansible_connection=local`). I don't think it will be defined when `--connection=local` is passed, nor if `connection: local` is a playbook parameter. It is maddening.

So, I fear the best option still is something like (untested):
```
when: not sshd_permit_root_login and '127.0.0.1' != ansible_ssh_host | default(inventory_hostname)
```
It's hard to believe there isn't some ansible fact to access that indicates that the connection is local.

chriszarate:
Shame upon me and glory _again_ to @fullyint. Now using `ansible_connection`, which is undefined unless the `--connection` flag is passed to Ansible. We should only ever skip this task if `--connection=local` is passed. Tested with both named hosts and IP addresses in Ansible `hosts` file.

roles/users/tasks/main.yml
chriszarate: I've tested it a few different ways, and `ansible_connection` is defined and set to `local` when I (or Packer) pass `--connection=local`. Ansible 1.9.3.

chriszarate: It does seem that _actually connecting_ over SSH to a host overrides `ansible_connection`. Ansible seems to prefer connecting over SSH if it can, even when `--connection=local` is set. So if Ansible knows how to SSH to `127.0.0.1`, it will, and will "undefine" `ansible_connection` accordingly. That seems like good behavior, even if Ansible perhaps should alert you that it's not using your preferred connection. Not sure if that's what is happening for you.

fullyint: The results you are seeing seem like what is expected. My results seem unexpected (Ansible 1.9.2 & 1.9.3). So, I'd trust yours. I'm guessing my problem is that I'm not testing via Packer, but on my local mac. Or I simply may not be understanding how things work. 

I create an entry in my `hosts/staging` file that shouldn't be able to connect via SSH because there's no IP to help the connection succeed. So, `ansible_connection` should remain `local`. It shouldn't be set back to undefined due to a successful SSH connection.
 ```
[web]
fake
```
Then I run `ansible-playbook server.yml -i hosts/staging --connection=local`. I've made the first couple tasks check the value of the variable: 
```
- debug: var=ansible_connection
- local_action: debug msg="{{ ansible_connection }}"
```
They both come out undefined. I'm probably missing something.

chriszarate: Very strange. It seems like an entry in your Ansible `hosts` file implies "connect to this host via SSH," which Ansible does no matter what you pass with `--connection`. And it sets `ansible_connection` accordingly, whether or not it is actually able to connect. It does this even for inventories passed as string with `-i` (e.g., `-i "[web] 127.0.0.1,"`—that comma is important). This is especially strange because this is how Packer does it.

If, however, you set the host to `127.0.0.1` in the playbook via `hosts: 127.0.0.1`, it works. I'll take another look this weekend.

chriszarate: @fullyint, I'm able to replicate your issue now. Thanks again. This is a bit of a bummer. I don't see any clear way forward except (1) trying to make intelligent guesses based on related connection variables or (2) requiring `ansible_connection` to be set via inventory or extra vars. I suppose the latter is best and can be documented.

swalkinshaw:
@chriszarate good to go or not yet?

chriszarate:
@swalkinshaw, Not yet. I'll get back to it this weekend.

swalkinshaw:
Closing this for now. @chriszarate feel free to re-open whenever you get a chance to work on this again.

PR #361, opened 2015-09-21
Move client_max_body_size outside of fastcgi block.

chriszarate:
Many (plugin-supplied) upload URLs are rewritten, so putting `client_max_body_size` in a block that only matches literal PHP files means users posting to a path like `/my-custom-form/upload/` will get a `HTTP 413` error if they exceed Nginx's default `client_max_body_size` of 1 MB.

### DIFF
--- a/roles/nginx/templates/wordpress.conf.j2
+++ b/roles/nginx/templates/wordpress.conf.j2
@@ -9,6 +9,9 @@ location / {
   try_files $uri $uri/ /index.php?$args;
 }
 
+# Set the max body size equal to PHP's max POST size.
+client_max_body_size {{ php_post_max_size | default('25m') | lower }};
+
 include h5bp/directive-only/x-ua-compatible.conf;
 include h5bp/location/cross-domain-fonts.conf;
 include h5bp/location/protect-system-files.conf;
--- a/roles/wordpress-setup/templates/wordpress-site.conf.j2
+++ b/roles/wordpress-setup/templates/wordpress-site.conf.j2
@@ -87,7 +87,6 @@ server {
     {% else -%}
     fastcgi_pass unix:/var/run/php5-fpm-wordpress.sock;
     {% endif -%}
-    client_max_body_size 0;
   }
 }
 


austinpray:
Couple things here:

1. I don't like that we are disabling the max request size. Perhaps a more seasoned sysadmin can chime in, but as far as I know this opens you up to low effort DDOS attacks. 
2. Maybe you could accomplish this for your site by using https://github.com/roots/trellis/blob/57e9d3902ffeb708a7264df228d21dfa140becba/roles/wordpress-setup/tasks/nginx.yml#L19-L24 and create a site-specific conf? That would be the "correct" solution. 

I would definitely recommend making the client_max_body_size configurable by the user via a variable such as `nginx_client_max_body_size`. For those php files it should equal the [PHP max upload size](https://github.com/roots/trellis/blob/7f78498a099784816d7ee25987a57f6791c71bd9/roles/php/defaults/main.yml#L12-L15).

chriszarate:
Good point. Disabling Nginx's max body size isn't a good approach. I've updated this PR to raise it to PHP's mast POST size. However, I do think that only raising it for URIs that end in `.php` will run counter to quite a few users' expectations.

For example, a default BuddyPress install has at least one rewritten upload URLs (for avatars) that will get hit with Nginx's default 1 MB limit. Many stock form plugins will have the same issue. 

I think what's happening is that Nginx applies the default `client_max_body_size` whenever rewritten URLs hit `try_files`. It returns `413` and never hits the higher one in the fastcgi block.

I moved the modified `client_max_body_size` to `wordpress.conf.j2`, since it's really a WordPress-specific change. As a WordPress-centric project, there's not much use case for increasing the limit for non-PHP files, so I didn't add a `nginx_client_max_body_size` to apply more broadly.

austinpray:
Yeah I'm fine with this compromise. This should actually be better than what we had before where we totally disabled the check. @louim @fullyint any objections?

louim:
Good for me.  The `lower` filter work correctly even when the are numbers?

austinpray:
@louim yep it appears to be working. Here is a DO instance that I just provisioned with this PR to test: `ssh root@159.203.71.235`

fullyint:
Afraid I don't know enough about it to be helpful. Except I did notice that although supposedly you can set `client_max_body_size` in the `http`, `server`, or `location` blocks, [some](http://stackoverflow.com/a/14358928) say you must set it in all three (doubtful). I bring it up just to comment that I believe this switches `client_max_body_size` from a `location` block to a `server` block. If you've tested that this works and solves the problem, @chriszarate, then sounds good to me. Thanks for your work on it!

austinpray:
Out of the box 

```
# Ansible managed: /Users/austinpray/Desktop/trellis/roles/nginx/templates/wordpress.conf.j2 modified on 2015-09-24 19:05:56 by austinpray on AustinMacbookPro

# Prevent PHP scripts from being executed inside the uploads folder.
location ~* /app/uploads/.*\.php$ {
  deny all;
}

location / {
  try_files $uri $uri/ /index.php?$args;
}

# Set the max body size equal to PHP's max POST size.
client_max_body_size 25m;

include h5bp/directive-only/x-ua-compatible.conf;
include h5bp/location/cross-domain-fonts.conf;
include h5bp/location/protect-system-files.conf;
```

Bumping up to 100M

```
# Ansible managed: /Users/austinpray/Desktop/trellis/roles/nginx/templates/wordpress.conf.j2 modified on 2015-09-24 19:05:56 by austinpray on AustinMacbookPro

# Prevent PHP scripts from being executed inside the uploads folder.
location ~* /app/uploads/.*\.php$ {
  deny all;
}

location / {
  try_files $uri $uri/ /index.php?$args;
}

# Set the max body size equal to PHP's max POST size.
client_max_body_size 100m;

include h5bp/directive-only/x-ua-compatible.conf;
include h5bp/location/cross-domain-fonts.conf;
include h5bp/location/protect-system-files.conf;
```

looks like it's working!

chriszarate:
Thanks, y'all!

PR #362, opened 2015-09-25
Add wildcard to site_host for port 80 if multsite subdomains is enabled (mirror port 443).

chriszarate:
Without this, requests to http://subdomain.example.com hit `no-default.conf` and are served `HTTP 444`.

### DIFF
--- a/roles/wordpress-setup/templates/wordpress-site.conf.j2
+++ b/roles/wordpress-setup/templates/wordpress-site.conf.j2
@@ -93,7 +93,7 @@ server {
 {% if item.value.ssl is defined and item.value.ssl.enabled | default(False) %}
 server {
   listen 80;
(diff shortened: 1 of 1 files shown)


austinpray:
Haven't run into this but maybe I didn't test rigorously enough. I'll check this out.

chriszarate:
Automatically adding the wildcard in the main `server` block was relatively recent (8e3ae15). Before that, I explicitly listed the wildcard subdomain in `site_hosts`. Perhaps you did, too. (It causes no harm if you still do that, just a little redundant.)

chriszarate:
To be clear, the `444` behavior only happens if you use SSL and multisite subdomains.

swalkinshaw:
I'm thinking we should keep this explicit for now. Maybe if we get better multisite handling in the future we can revisit it but I'd rather not silently set up hosts which could be confusing.

PR #366, opened 2015-09-29
Use Windows envar to determine number of CPUs

QWp6t:
Modern versions of Windows have an environment variable `NUMBER_OF_PROCESSORS` that can be used as a quick and dirty way to get the number of CPUs.

Older versions of Windows, where this envar isn't present, will still fall back to a value of `2`.


QWp6t:
I modified this slightly from the initial commit. I moved the line to the top so that the environment variable is checked first. This is to allow users on any platform to set that variable as an easy means of automatically setting the number of CPUs used in their VMs.

swalkinshaw:
:+1:

Issue #315, opened 2015-08-27
Inconsistencies with logfiles


retlehs:
one server:

```
-rw-r----- 1 web  www-data         0 Aug 23 06:45 access.log
-rw-r--r-- 1 root root     121922353 Aug 26 04:31 access.log.1
-rw-r----- 1 web  www-data         0 Aug 23 06:45 error.log
-rw-r--r-- 1 root root         83120 Aug 26 01:41 error.log.1
```

another server:

```
-rw-r--r-- 1 root root     2276589 Aug 26 04:31 access.log
-rw-r--r-- 1 root root       73289 Aug 26 03:23 error.log
```

i don't think either of these are working as intended. 

ref https://github.com/roots/trellis/blob/0a7bdef92097bf875ca8eb1f68f4111549b29fa6/group_vars/all#L60

louim:
> `-rw-r----- 1 web  www-data         0 Aug 23 06:45 error.log`

This one is matching. When the rotation occur, the new log are set correctly. Seem's like the initial log doesn't have the right owner / permissions

kalenjohnson:
I'm also seeing this:
```
-rw-r----- 1 deploy www-data       0 Aug 26 06:35 access.log
-rw-r----- 1 deploy www-data 2651252 Aug 27 23:03 access.log.1
-rw-r----- 1 deploy www-data  213008 Aug 25 10:42 access.log.2.gz
-rw-r----- 1 deploy www-data  240942 Aug 19 09:40 access.log.3.gz
-rw-r----- 1 deploy www-data  288319 Aug 11 13:31 access.log.4.gz
-rw-r----- 1 deploy www-data  262321 Aug  5 12:27 access.log.5.gz
-rw-r----- 1 deploy www-data  803570 Jul 30 20:15 access.log.6.gz
-rw-r--r-- 1 root   root      283252 Jul 15 19:10 access.log.7.gz
```

austinpray:
![image](https://cloud.githubusercontent.com/assets/2192970/9538286/4604883c-4d08-11e5-886e-6b5b840944db.png)

pls fix asap tyvm

kalenjohnson:
actually looks like @austinpray is user error

fullyint:
[...]

PR #377, opened 2015-10-07
Fix Nginx postrotate log script

alan-c:
Addresses one of the two issues raised here: https://github.com/roots/trellis/issues/315

Per `man service`:

```
The existence of an upstart  job  of  the  same name 
as  a script in /etc/init.d will cause the upstart job
to take precedence over the init.d script.
```

So, `invoke-rc.d nginx rotate >/dev/null 2>&1` fails because the `rotate` function has been removed since the switch of nginx to Upstart. This is seen when running `sudo logrotate -vf /etc/logrotate.d/wordpress-sites` on a vanilla bedrock/trellis/sage install, which will result in:

`error: error running shared postrotate script for '"/srv/www/**/logs/*.log" '`

Of note, the script should send the -USR1 signal to the master process to avoid the old worker processes which are shutting down from still using the old log file(s).

### DIFF
--- a/group_vars/all/logrotate.yml
+++ b/group_vars/all/logrotate.yml
@@ -17,4 +17,4 @@ logrotate_scripts:
         if [ -d /etc/logrotate.d/httpd-prerotate ]; then \
               run-parts /etc/logrotate.d/httpd-prerotate; \
             fi \
-      postrotate: invoke-rc.d nginx rotate >/dev/null 2>&1
+      postrotate: service nginx rotate


group_vars/all/logrotate.yml
```diff
@@ -17,4 +17,4 @@ logrotate_scripts:
         if [ -d /etc/logrotate.d/httpd-prerotate ]; then \
               run-parts /etc/logrotate.d/httpd-prerotate; \
             fi \
-      postrotate: invoke-rc.d nginx rotate >/dev/null 2>&1
+      postrotate: "[ -f /var/run/nginx.pid ] && kill -USR1 `cat /var/run/nginx.pid`"
```
swalkinshaw: There might be a really obvious reason why I'm wrong, but can't this just be `service nginx reload`?

alan-c: No, if you use `service nginx reload` won't work completely. The new worker processes will log to the correct file(s), but the old worker processes won't. They'll continue to use the old log file(s) as long as they hang around.

It won't be much of a discrepancy, but it is there. The busier the site, the bigger the discrepancy.

louim: @alan-c is right. This is the recommendend way for Nginx to handle log file reopening. See: http://nginx.org/en/docs/control.html

Alan, you should do `$(<"/var/run/nginx.pid")`. See : http://stackoverflow.com/questions/21432883/pid-cat-pidfile-or-read-pid-pidfile

alan-c: Thanks for the reference @louim. I'll clean up the syntax.

alan-c: Or..... I can ask for input. `strace` does indicate a performance benefit to `read pid < file` versus `cat file`. However, after trying everything I can think of to get the command working inside `/etc/logrotate.d/wordpress-sites`, I'm stuck with variations along the lines of this...

```
    PIDfile="/var/run/nginx.pid"
    if test ! -f $PIDfile
    then
      read nginxPID < $PIDfile
      kill -USR1 $nginxPID
    fi
```
or other similar. All attempts I've tried to get `$(<"/var/run/nginx.pid")` working with the `kill` command haven't found success. Not sure if it's an issue with syntax or what, but only works if I split it into two lines (`read` then `kill`) and explicitly use `read`. Depending on what I've tried I get either nothing at all (no error/no successful kill command) or permission denied errors.

Any thoughts?

louim: @alan-c I just tested `kill -USR1 $(<"/var/run/nginx.pid")` outside the script, and it seems to work correctly (with root privilege as expected). Is it working for you outside the script? I didn't want to make it more complicated for you. If it doesn't work inside the script, we could use `cat` as in your initial commit. the performance difference is minimal anyway.

alan-c: @louim Thanks for looking at it. I can get it to work outside ok, but for some reason that is eluding me it's failing inside, including on fresh installs. FWIW, a simple `echo` as opposed to `kill` fails inside as well - no output.

louim: I just looked a little bit more into it. `nginx` service actually have a `rotate` which seems to be exactly what we want. @alan-c want to try it?

alan-c: I'll test it out. I had seen a minor reference to it at [Stack Exchange](http://unix.stackexchange.com/questions/209585/nginx-logrotate-error-on-cron-job) when I initially looked at the issue, but I stuck with the `kill` command based-solution I'd used before.

According to [Nginx Essentials](http://shop.oreilly.com/product/9781785289538.do) [ that's a clean link, not an affiliate one ] on p.40, `rotate` sends a USR1 signal, so it should be good.

louim: Yep, I checked what the `rotate` command does:

```
rotate_logs() {
	# Rotate log files
	start-stop-daemon --stop --signal USR1 --quiet --pidfile $PID --name $NAME
	return 0
}
```

alan-c: Tests work nicely. Good catch, @louim. Command even adds brings a nice message to `logrotate -d`: ` * Re-opening nginx log files nginx`.

alan-c:
UUOC award presented to the first commit. I humbly accept on it's behalf. :smirk:

austinpray:
This is sick

swalkinshaw:
@alan-c possible to get a squash? Looks good :+1:

alan-c:
Squashed it.

swalkinshaw:
Thanks!

robrecord:
I am getting the error on my cron emails:

```
/etc/cron.daily/logrotate:
Usage: /etc/init.d/nginx {start|stop|status|restart|reload|force-reload|upgrade|configtest|check-reload}
error: error running shared postrotate script for '"/srv/www/**/logs/*.log"
'
run-parts: /etc/cron.daily/logrotate exited with return code 1
```

I dug around, and found that the command `rotate` was removed from the nginx init.d script, as per the following discussion:
https://bugs.launchpad.net/nginx/+bug/1450770/comments/7

This change appears to have been implemented incorrectly: https://github.com/roots/trellis/commit/e8d1b4a19cc10e2a8e41b806f7c5c95dcc080f59#diff-cbcc749dd2e98d8f319879eef36d48723e2c1ccf9d93b3d6ae4c3dd39ff3a7fe

I have added a fix in the comments there. Not sure how to invoke to test completely other than waiting to see if it works.

PR #381, opened 2015-10-13
Add dev's IP to ferm whitelist

QWp6t:
poached from [phil's method in fail2ban](https://github.com/roots/trellis/blob/master/roles/fail2ban/defaults/main.yml#L6).

### DIFF
--- a/group_vars/all/security.yml
+++ b/group_vars/all/security.yml
@@ -2,6 +2,9 @@ ferm_input_list:
   - type: dport_accept
     dport: [http, https]
     filename: nginx_accept
+  - type: dport_accept
+    dport: [ssh]
+    saddr: ["{{ lookup('pipe', 'dig +short myip.opendns.com @resolver1.opendns.com') }}"]
   - type: dport_limit
     dport: [ssh]
     seconds: 300


austinpray:
If you have multiple people provisioning a box or you provision a box from two different locations, does it remove the old entries? I will test but if anyone can get to it faster than me go ahead.

louim:
Yeah one problem with this is that it will only use the most recent deployer's ip. (Each deployment would overwrite the ip. Not ideal for organization I think.

austinpray:
I'm okay with each deployment overwriting the IP addresses as long as it doesn't leak ips such that every single place you have provisioned from (or every person who has ever provisioned) is whitelisted.

mAAdhaTTah:
Wouldn't that be preferable, actually? Presumably, only developers or people associated with the project are deploying, so wouldn't you want all of their IPs whitelisted?

louim:
@mAAdhaTTah it's in the config files. There nothing preventing your to edit the file to make the rule like so:

```
  - type: dport_accept
    dport: [ssh]
    saddr: ["my.first.dev.ip", "my.second.dev.ip", "my.third.dev.ip"]
```

it wouldn't be easily feasible to generate this list dynamically, because of the way the role is made.

mAAdhaTTah:
@louim Right, that looks like the behavior @austinpray prefers. I was just suggesting the other way might preferable, though if it's not feasible, it doesn't really matter :).

swalkinshaw:
@QWp6t pls add changelog entry.

PR #388, opened 2015-10-19
Adds HSTS configuration variables

austinpray:
Adds the following WordPress site variables

ssl.nginx_hsts_max_age
ssl.nginx_hsts_include_subdomains
ssl.nginx_hsts_preload

Adds the following nginx global defaults

nginx_hsts_max_age: 31536000
nginx_hsts_include_subdomains: true
nginx_hsts_preload: true

### DIFF
--- a/roles/wordpress-setup/templates/wordpress-site.conf.j2
+++ b/roles/wordpress-setup/templates/wordpress-site.conf.j2
@@ -38,7 +38,10 @@ server {
   include ssl.conf;
   include ssl-stapling.conf;
 
-  add_header Strict-Transport-Security "max-age=31536000; includeSubdomains; preload";
+  {% set hsts_max_age = item.value.ssl.hsts_max_age | default(nginx_hsts_max_age) %}
+  {% set hsts_include_subdomains = item.value.ssl.hsts_include_subdomains | default(nginx_hsts_include_subdomains) | ternary('includeSubdomains', None) %}
+  {% set hsts_preload = item.value.ssl.hsts_preload | default(nginx_hsts_preload) | ternary('preload', None) %}
+  add_header Strict-Transport-Security "max-age={{ [hsts_max_age, hsts_include_subdomains, hsts_preload] | reject('none') | join('; ') }}";
 
   {% if item.value.ssl.cert is defined and item.value.ssl.key is defined %}
   ssl_certificate         /etc/nginx/ssl/{{ item.value.ssl.cert | basename }};
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 * Update SSL cipher suite ([#386](https://github.com/roots/trellis/pull/386))
 * Support for other Vagrant providers (VirtualBox, VMWare, Parallels) ([#340](https://github.com/roots/trellis/pull/340))
 * Specify versions for Ansible Galaxy requirements ([#385](https://github.com/roots/trellis/pull/385))
+* Adds ability to configure [HSTS headers](https://developer.mozilla.org/en-US/docs/Web/Security/HTTP_strict_transport_security) with site variables. ([#388](https://github.com/roots/trellis/pull/388))
 
 ### 0.9.2: October 15th, 2015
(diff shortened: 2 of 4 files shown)


louim:
Any particular reason why we would want these options to be configurable vs enforcing them as sane defaults?

retlehs:
i ran into an issue today where `includeSubDomains` affected subdomains and made them inaccessible, had to remove it and set the `max-age` to 0

https://shrikantadhikarla.wordpress.com/2013/02/18/anatomy-of-includesubdomains-directive-in-http-strict-transport-security-policy/

austinpray:
@louim if you need to refresh your settings you might need to set max_age to 0 temporarily

>Whenever the Strict-Transport-Security header is delivered to the browser, it will update the expiration time for that site, so sites can refresh this information and prevent the timeout from expiring. Should it be necessary to disable Strict Transport Security, setting the max-age to 0 (over a https connection) will immediately expire the Strict-Transport-Security header, allowing access via http.

https://developer.mozilla.org/en-US/docs/Web/Security/HTTP_strict_transport_security

Also, mind as well keep them configurable since these are potentially site-breaking options

roles/wordpress-setup/templates/wordpress-site.conf.j2
```diff
@@ -38,7 +38,11 @@ server {
   include ssl.conf;
   include ssl-stapling.conf;
 
-  add_header Strict-Transport-Security "max-age=31536000; includeSubdomains; preload";
+  {% set hsts_max_age = item.value.ssl.hsts_max_age | default(nginx_hsts_max_age) %}
+  {% set hsts_include_subdomains = item.value.ssl.hsts_include_subdomains | default(nginx_hsts_include_subdomains) %}
+  {% set hsts_preload = item.value.ssl.hsts_preload | default(nginx_hsts_preload) %}
+
+  add_header Strict-Transport-Security "max-age={{ hsts_max_age }}{{ hsts_include_subdomains | ternary('; includeSubdomains', '') }}{{ hsts_preload | ternary('; preload', '') }}";
```
swalkinshaw: do we care how messy this is? If the variables were strings it could be:

```
add_header Strict-Transport-Security "max-age={{ [hsts_max_age, hsts_include_subdomains, hsts_preload].join(';') }}";
```

austinpray: The reason why I did the manual "; " prefix is because I was trying to follow the format:

image

https://developer.mozilla.org/en-US/docs/Web/Security/HTTP_strict_transport_security#Enabling_Strict_Transport_Security

Other formatting examples

https://tools.ietf.org/html/draft-hodges-strict-transport-sec-02#page-17

https://www.owasp.org/index.php/HTTP_Strict_Transport_Security

roles/wordpress-setup/templates/wordpress-site.conf.j2
```diff
@@ -38,7 +38,10 @@ server {
   include ssl.conf;
   include ssl-stapling.conf;
 
-  add_header Strict-Transport-Security "max-age=31536000; includeSubdomains; preload";
+  {% set hsts_max_age = item.value.ssl.hsts_max_age | default(nginx_hsts_max_age) %}
+  {% set hsts_include_subdomains = item.value.ssl.hsts_include_subdomains | default(nginx_hsts_include_subdomains) | ternary('includeSubdomains', None) %}
+  {% set hsts_preload = item.value.ssl.hsts_preload | default(nginx_hsts_preload) | ternary('preload', None) %}
+  add_header Strict-Transport-Security "max-age={{ [hsts_max_age, hsts_include_subdomains, hsts_preload] | reject('none') | join('; ') }}";
```
austinpray: results in `add_header Strict-Transport-Security "max-age=31536000; includeSubdomains; preload";`

swalkinshaw:
@austinpray changelog pls :watermelon:

austinpray:
@roots/trellis-contributors gimme a couple extra go-aheads and I'll merge.

swalkinshaw:
:+1:

louim:
:+1:

PR #390, opened 2015-10-21
Updates variables to use curl instead of dig

mxxcon:
The present `dig` command will not work on networks that restrict outgoing UDP traffic only to the network's own DNS servers.
This change uses `curl` to query OpenDNS's endpoint (with https to ensure response integrity).

### DIFF
--- a/group_vars/all/security.yml
+++ b/group_vars/all/security.yml
@@ -4,7 +4,7 @@ ferm_input_list:
     filename: nginx_accept
   - type: dport_accept
     dport: [ssh]
-    saddr: ["{{ lookup('pipe', 'dig +short myip.opendns.com @resolver1.opendns.com') }}"]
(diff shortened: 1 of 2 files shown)


louim:
Probably should be done there too? https://github.com/roots/trellis/blob/b3f322b5a22735f0c654c9ff5e7b7ae0f0f478b9/group_vars/all/security.yml#L7

mxxcon:
Thank you @louim. Missed that one.

QWp6t:
This is good for Windows users who use Ansible with Babun/Cygwin. By default, Cygwin includes curl but not dig. So yay!

swalkinshaw:
@mxxcon thanks. Possible to get the commits squashed into 1 if you're familiar with the process?

/cc @fullyint as well

mxxcon:
@swalkinshaw squished as requested. :dancer:

swalkinshaw:
Thanks!

Issue #393, opened 2015-10-23
Update cron setting + task


swalkinshaw:
The `system_cron` variable will be replaced by a `disable_wp_cron` env var. See https://github.com/roots/bedrock/pull/206

PR #394, opened 2015-10-24
Replace system_cron with disable_wp_cron

swalkinshaw:
Update to match Bedrock's new setting.

Ref https://github.com/roots/trellis/issues/393 and https://github.com/roots/bedrock/pull/206


retlehs:
shouldn't `disable_db_cron` be replaced with `disable_wp_cron`?

swalkinshaw:
:sob: yes

swalkinshaw:
@retlehs good now??

retlehs:
what about not doing this for multisite?

PR #397, opened 2015-10-26
Add deploy_build_before example hook for theme assets

retlehs:
not everyone using trellis is going to have a theme that needs this, but i would still like this to exist as a placeholder.

see https://discourse.roots.io/t/deployment-with-the-new-trellis-hooks/5016/2?u=benword

### DIFF
--- a/deploy-hooks/build-before.yml
+++ b/deploy-hooks/build-before.yml
@@ -1,0 +1,29 @@
+# Placeholder `deploy_build_before` hook for building theme assets locally
+# and then copying the files to the remote server
+#
+# Uncomment the lines below and replace `sage` with your theme folder
+#
+# ---
+# - name: Run npm install
+#   command: npm install
+#   connection: local
+#   args:
+#     chdir: "{{ project.local_path }}/web/app/themes/sage"
+# - name: Run bower install
+#   command: bower install
+#   connection: local
+#   args:
+#     chdir: "{{ project.local_path }}/web/app/themes/sage"
+# - name: Run gulp
+#   command: gulp --production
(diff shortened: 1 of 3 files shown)


retlehs:
another benefit is that this also provides an example of how to use a deploy hook along with where to put it

deploy-hooks/build-before.yml
```diff
@@ -0,0 +1,30 @@
...
+#   args:
+#     chdir: "{{ project.local_path }}/web/app/themes/sage"
+# - name: Run gulp
+#   command: gulp --production
+#   connection: local
+#   args:
+#     chdir: "{{ project.local_path }}/web/app/themes/sage"
+# - name: Copy project local files
+#   synchronize: src="{{ item.src }}"
+#                dest="{{ deploy_helper.new_release_path }}/{{ item.dest }}"
```
swalkinshaw: need to use new syntax here

fullyint:
Could this PR also remove the the existing "[Copy project local files](https://github.com/roots/trellis/blob/6e277a4d4106a3dd66ee19cce48c4f9cef32b92c/roles/deploy/tasks/build.yml#L4-L11)" task and the corresponding [`project_local_files`](https://github.com/roots/trellis/blob/6e277a4d4106a3dd66ee19cce48c4f9cef32b92c/roles/deploy/defaults/main.yml#L28-L37) default variable? They seem obsolete now that this PR's "Copy project local files" task demonstrates the new recommended approach of using deploy hooks.

swalkinshaw:
@fullyint good catch! Definitely need to remove those.

PR #399, opened 2015-10-28
Remove unused deploy defaults

fullyint:


swalkinshaw:
What about `project_local_files` in here as well?

fullyint:
Good idea. I removed `project_local_files` and the task that used it. #397 will demonstrate syncing files using the recommended approach of deploy hooks.

retlehs:
also https://github.com/roots/trellis/blob/de656ec28b51c78fd49f37d1f1187a131eca5802/roles/deploy/defaults/main.yml#L22

PR #400, opened 2015-10-29
Enable sorted iteration over env vars

nathanielks:
Python dictionaries aren't sorted so this enables sorting. Use case: `vlucas/phpdotenv` >= 2.0 enables nested variables referencing other envvars. If the envvar hasn't been declared yet, it'll leave it alone (which isn't desired). By sorting the envvars and iterating over them in the order defined, it enables us to better utilize this functionality.

### DIFF
--- a/roles/wordpress-install/templates/env.j2
+++ b/roles/wordpress-install/templates/env.j2
@@ -1,3 +1,4 @@
-{% for key, value in item.value.env.iteritems() %}
-{{ key | upper }}="{{ value }}"
+{% set env = item.value.env %}
+{% for item in env | dictsort %}
+{{ item.0 | upper }}={{ item.1 }}
 {% endfor %}


louim:
http://jinja.pocoo.org/docs/dev/templates/#dictsort

It wont keep the the original declaration order, but it will sort them alphabetically by key. Unless there is something that I'm missing, this wouldn't really help your case (except if your variables are named alphabetically).

roles/wordpress-install/templates/env.j2
```diff
@@ -1,3 +1,4 @@
-{% for key, value in item.value.env.iteritems() %}
-{{ key | upper }}="{{ value }}"
+{% set env = item.value.env %}
+{% for item in env | dictsort %}
```
swalkinshaw: no way to preserve `key, value` with this?

nathanielks:
@louim herp derp. You're totally right. Back to the drawing board...

roles/wordpress-install/templates/env.j2
nathanielks: unfortunately not as dictsort converts it to 0, 1

PR #392, opened 2015-10-23
WP-CLI role improvements - fixes #391

swalkinshaw:
* Switch to better syntax
* Improve variable names with `wp_cli_` prefix
* Use WP-CLI nightly

PR #407, opened 2015-11-11
Added WP-CLI bash completion script.

primozcigler:
This change automatically installs the tab completion script for wp-cli, making it more fun and less error-prone to use.

### DIFF
--- a/roles/wp-cli/defaults/main.yml
+++ b/roles/wp-cli/defaults/main.yml
@@ -1,2 +1,4 @@
 phar_url: https://raw.githubusercontent.com/wp-cli/builds/gh-pages/phar/wp-cli.phar
 bin_path: /usr/bin/wp
+completion_url: https://raw.githubusercontent.com/wp-cli/wp-cli/master/utils/wp-completion.bash
+completion_path: /etc/bash_completion.d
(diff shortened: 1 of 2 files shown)


roles/wp-cli/defaults/main.yml
```diff
@@ -1,2 +1,4 @@
 phar_url: https://raw.githubusercontent.com/wp-cli/builds/gh-pages/phar/wp-cli.phar
 bin_path: /usr/bin/wp
+wp_cli_completion_url: https://raw.githubusercontent.com/wp-cli/wp-cli/master/utils/wp-completion.bash
```
swalkinshaw: I think this should wait for #392 if we do it since there could be a disconnect between the completion commands.

primozcigler: Agreed. That's why I splitted it into 2 commits. I don't know why #392 is not merged yet though?

swalkinshaw: We internally debated about whether we should be using a nightly version at all. Because then there's a difference between dev/prod.

primozcigler: I see. But bash completion is the same for both AFAIK (I think is the the same script since I started using WP-CLI some 2 years ago or so).

swalkinshaw:
I really should have look at the script. Of course it just determines the completions from wp-cli so it will be in sync.

Thanks a lot :)

primozcigler:
I'm glad to contribute to this project, sir! :smiley:

PR #411, opened 2015-11-13
Add `wp core update-db` to finalize hook

### LINKED ISSUE #398: Add `wp core upgrade-db` to deploy task
Got the idea here: https://github.com/wp-cli/wp-cli/issues/2171 Thought it might be a useful addition. Would trellis be interested in adding this by default or would this be better as a per-user thing?

mAAdhaTTah:
Fixes #398.

~~Couple things: first, I think this is where this is supposed to go, but if I'm reading this right, it doesn't look like this file is actually run during the deployment process and works more like an example? And second, the reason I don't know the answer to that is I haven't fully tested it because I'm running an old version of Trellis that I need to upgrade to the latest variable configuration (you'll notice my fork is still named `bedrock-ansible`).~~

~~That said,~~ I'm opening this up for comments, since this is my first Ansible task (which, as a side note, is a really neat benefit of the deployment hooks: it allows you to dip your toes into writing Ansible tasks without needing to understand everything at once).

Thoughts?

### DIFF
--- a/roles/deploy/hooks/finalize-after.yml
+++ b/roles/deploy/hooks/finalize-after.yml
@@ -11,7 +11,24 @@
   command: wp eval 'wp_clean_themes_cache(); switch_theme(get_stylesheet());'
   args:
     chdir: "{{ deploy_helper.new_release_path }}"
-  when: wp_installed.rc == 0
+  when: wp_installed | success
+
+- name: Update WP database
+  command: wp core update-db
+  args:
+    chdir: "{{ deploy_helper.new_release_path }}"
+  when: wp_installed | success and not project.multisite.enabled | default(false)
+
+- name: Warn about updating network database.
+  debug:
+    msg: "Updating the network database could take a long time with a large number of sites."
+  when: wp_installed | success and project.multisite.enabled | default(false)
+
+- name: Update WP network database
+  command: wp core update-db --network
+  args:
+    chdir: "{{ deploy_helper.new_release_path }}"
+  when: wp_installed | success and project.multisite.enabled | default(false)
 
 - name: Reload php5-fpm
   shell: sudo service php5-fpm reload


mAAdhaTTah:
Tested and working now.

roles/deploy/hooks/finalize-after.yml
```diff
@@ -13,6 +13,18 @@
...
   when: wp_installed.rc == 0
 
+- name: Update WP database
+  command: wp core update-db
+  args:
+    chdir: "{{ deploy_helper.new_release_path }}"
+  when: wp_installed.rc == 0 and not (project.multisite.enabled | default(False) == True)
+
+- name: Update WP network database
+  command: wp core update-db
```
retlehs: missing `--network` here?

should we note that this could take a really long time for larger networks?

mAAdhaTTah: Musta gotten lost in a copy/paste.

Added a message. Let me know if you have a preference on the copy.

roles/deploy/hooks/finalize-after.yml
```diff
@@ -13,6 +13,22 @@
     chdir: "{{ deploy_helper.new_release_path }}"
   when: wp_installed.rc == 0
 
+- name: Update WP database
+  command: wp core update-db
+  args:
+    chdir: "{{ deploy_helper.new_release_path }}"
+  when: wp_installed.rc == 0 and not (project.multisite.enabled | default(False) == True)
```
swalkinshaw: These can all use `when: wp_installed | success`

Also shouldn't last `==` so:

```
wp_installed | success and not (project.multisite.enabled | default(false))
```

mAAdhaTTah: Will do. I pulled `wp_installed.rc == 0` from line 14; should I fix that too?

swalkinshaw: That would be great :)

swalkinshaw: oh and the parentheses aren't need either but it might be a subject thing:

`wp_installed | success and not project.multisite.enabled | default(false)`

mAAdhaTTah: Done and done.

roles/deploy/hooks/finalize-after.yml
```diff
@@ -11,7 +11,23 @@
...
+  when: wp_installed | success
+
+- name: Update WP database
+  command: wp core update-db
+  args:
+    chdir: "{{ deploy_helper.new_release_path }}"
+  when: wp_installed | success and not project.multisite.enabled | default(false)
+
+- name: Warn about updating network database.
+  debug: msg="Updating the network database could take a long time with a large number of sites."
```
swalkinshaw:
```yaml
debug:
  msg:
```

:)

swalkinshaw:
This is looking good :+1: Probably will want to do some tests ourselves but it's pretty straightforward.

roles/deploy/hooks/finalize-after.yml
mAAdhaTTah: Done!

swalkinshaw:
@mAAdhaTTah this is good to go. Can you squash commits?

mAAdhaTTah:
@swalkinshaw done!

swalkinshaw:
Thanks!

PR #413, opened 2015-11-15
Set vbox name, closes #408

### LINKED ISSUE #408: Vagrant default name
Since the last release, the default name of the virtual machine is 'trellis' and no more the parent directory's name. Now I can't no more quickly differentiate the machines from each other.

Any ideas how to fix this? Should I manually define a machine name in Vagrantfile or move this one to parent dir as we used to do before?

QWp6t:


QWp6t:
Let me know if this is an issue with other providers too. I don't have any others to test.

But it's basically the same process for all of them.

VMWare Workstation/Fusion
`vmw.vmx['displayName'] = config.vm.hostname`

Parallels
`prl.name = config.vm.hostname`

Vagrantfile
```diff
@@ -98,6 +98,9 @@ Vagrant.configure('2') do |config|
     # Fix for slow external network connections
     vb.customize ['modifyvm', :id, '--natdnshostresolver1', 'on']
     vb.customize ['modifyvm', :id, '--natdnsproxy1', 'on']
+    
+    # Set VM name
+    vb.customize ['modifyvm', :id, '--name', config.vm.hostname]
```
swalkinshaw: `vb.name = config.vm.hostname`

swalkinshaw: @QWp6t update

Vagrantfile
```diff
@@ -98,26 +98,24 @@ Vagrant.configure('2') do |config|
...
+    # Set VM name
+    vb.name = config.vm.hostname
   end
 
-  # VMware Fusion settings
-  config.vm.provider 'vmware_fusion' do |vmf, override|
-    # Override provider box
-    override.vm.box = 'puppetlabs/ubuntu-14.04-64-nocm'
+  # VMware Workstation/Fusion settings
+  ["vmware_fusion", "vmware_workstation"].each do |p|
```
swalkinshaw: single quotes

Vagrantfile
```diff
@@ -98,26 +98,24 @@ Vagrant.configure('2') do |config|
...
+    config.vm.provider "p" do |vmw, override|
```
swalkinshaw: ditto

Vagrantfile
swalkinshaw: use `provider` as var

swalkinshaw: actually don't care much about the var name

Issue #415, opened 2015-11-18
Config File and auto install Sage & Bedrock on Vagrant Up


nbyloff:
I have been playing around with a forked instance of Trellis, and I borrowed some code from another project on Github for testing. I have tested this on Windows, and while it says it works on Mac & Linux, I have not tested it. So far with that projects help I have:

* Changed the Vagrantfile to use a config file for specifying options
* Enabled SSH forwarding on Windows so I can deploy & provision to remote machines using my personal Public/Private keys

Because the SSH forwarding should work in all environments, next I hope to:

* Configure the Vagrantfile to optionally auto clone Bedrock & Sage and setup the project for me on first `vagrant up`
* Add config variables to allow me to auto install `nvm` and `bower` as well (optionally) on the VM.
* Allow user to optionally configure private repos they may have of their own existing projects

While this is pretty limited so far, this is my `Trellis.yaml` config file

```javascript
vm:
  ip: 192.168.50.5
  memory: 1024
  cpus: 1
  box: ubuntu/trusty64

trellis:
  version: master
  debug: false
  https: false
  ansible_path: ansible

# Github account info
git:
  username: myuser
  email: myemail@example.com

# Leave blank if using only local vagrant insecure key
# or you do not need to SSH forward to other boxes
# with the same key
security:
  box_public_key: ~/.ssh/id_rsa.pub
  box_private_key: ~/.ssh/id_rsa
  forward_agent: true
```

This is the project & Vagrantfile I based my changes on: https://github.com/alt3/cakebox
Here's the commit that accomplishes the top two bullets:
f56716779700d66fbdacefc2da4546d13c500286

Is this something worth doing or anyone is interested in? This whole project makes things really easy in a couple steps. Thought might be nice to have it all done in one step.

swalkinshaw:
Thanks for sharing this. It's pretty cool and definitely simpler which is something we're trying to get better at.

We've just started on a roots-cli project at https://github.com/roots/roots-cli which will go a long way towards that goal. It will automated a lot of the boilerplate setup.

I'm going to close this issue because of that, but feel free to continue some discussion in here if you want. There might be something in there that would work separate like Trellis. Maybe SSH keys/forwarding for WIndows? I don't know much about that process.

nbyloff:
Fair enough. I saw that project but posted here for now for visibility. Majority of my team is Mac, but I stress environment agnostic code before making it in the `master`. So I test everything in Windows.

[...]

PR #417, opened 2015-11-20
SSH Forwarding From Host to Guest

nbyloff:
References #415 discussion.

Added two scripts and tweaked the Vagrantfile to check for host machines public and private key. If it's found, it will copy the public key to the guest so the user can login with this key. Since SSH forwarding is enabled, this should allow the user to connect to external machines/account that are aware of this public key.

It works on Mac, Linux, and Windows. Windows users must also:

* "Allow agent forwarding" in PuTTY under Connection-->SSH-->Auth
* Point to their id_rsa.ppk in the same Auth window
* Have Pageant running with their id_rsa.ppk key added before logging into the guest.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -59,6 +59,48 @@ Vagrant.configure('2') do |config|
     end
   end
 
+  # Prevent annoying "stdin: not a tty" errors
+  config.ssh.shell = "bash -c 'BASH_ENV=/etc/profile exec bash'"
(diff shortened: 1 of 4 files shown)


nbyloff:
I created this PR on the wrong commit; it removes the cron email. Should we delete this one and I recommit so it doesn't affect the cron email PR?

Sorry :\

nbyloff:
I will just close it, and do it right.

PR #418, opened 2015-11-20
SSH forwarding host to guest for all environments

nbyloff:
References #415 discussion.

Added two scripts and tweaked the Vagrantfile to check for host machines public and private key. If it's found, it will copy the public key to the guest so the user can login with this key. Since SSH forwarding is enabled, this should allow the user to connect to external machines/accounts that are aware of this same public key from within the guest.

It works on Mac, Linux, and Windows. Windows users must also:

* "Allow agent forwarding" in PuTTY under Connection-->SSH-->Auth
* Point to their id_rsa.ppk in the same Auth window
* Have Pageant running with their id_rsa.ppk key added before logging into the guest.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -59,6 +59,48 @@ Vagrant.configure('2') do |config|
     end
   end
 
+  # Prevent annoying "stdin: not a tty" errors
+  config.ssh.shell = "bash -c 'BASH_ENV=/etc/profile exec bash'"
+  config.ssh.username = "vagrant"
+
+  # Replace insecure Vagrant ssh public key with user generated private/public keys
+  public_key = File.join(Dir.home, ".ssh", "id_rsa.pub")
+  private_key = File.join(Dir.home, ".ssh", "id_rsa")
+
+  if File.exists?(private_key)
+    # A public key MUST be accompanied by a private key
+    if File.exists?(public_key)
+      # Copy user's public key to the vm so it can be validated and applied
+      config.vm.provision "file", source: public_key, destination: "/home/vagrant/.ssh/" + File.basename(public_key)
+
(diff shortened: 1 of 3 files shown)


nbyloff:
If this makes it into master, when you write docs, you might also include steps on how to auto load pageant and the user keys on Windows at boot up so it's not a manual process. I could write them up, but this is essentially it:

http://blog.shvetsov.com/2010/03/making-pageant-automatically-load-keys.html

It also expects keys to be in `.ssh` folder in the user directory, so on Windows, the user will need to put their keys in a folder called `C:\Users\username\.ssh`. By default the names need to be id_rsa.pub and id_rsa.ppk.

QWp6t:
I got a chance to review this, and I don't really understand why the keys are being copied to the vagrant box. The whole point of SSH forwarding is to avoid having to copy keys to intermediate servers. The keys are made available throughout the SSH connection chain when forwarding is enabled.

I put the following in my `%USERPROFILE%\.ssh\config` file and SSH forwarding works fine for me...
```
Host *
        ForwardAgent yes
```
If you want to be more explicit, you can set it on a per-host basis. Either way, this has worked for me; SSH forwarding works on my end, which brings me back to my first sentence: I don't really understand why the keys are being copied to the vagrant box.

I'll leave this open for now in case there's something I'm missing.

QWp6t:
I've added some docs for SSH forwarding to assist Windows users.

https://roots.io/trellis/docs/windows/#ssh-forwarding

I've tested all of the methods noted in the docs. They all work without having to copy any keys to the Vagrant guest.

This isn't as much of an issue for non-Windows users since the Ansible controller is the host machine, not the guest.

I'm going to go ahead and close this for now. You're still free to let me know if there's something I'm missing and if we should revisit this PR.

PR #414, opened 2015-11-16
Stop cron from emailing local server account

nbyloff:
This resolves issue #412

PR #421, opened 2015-11-21
Fixes #412 - Stop cron job emails

### LINKED ISSUE #412: Stop Cron Jobs from Sending Emails
By default, cron is trying to send emails after the http://sitename.com/wp/wp-cron.php job is run. I do this manually every time I checkout Trellis, but thought it might be useful for all installs, or at least a configurable option that allows the user to turn off email responses after a cron job is run.

Done simply by adding ``` > /dev/null 2>&1```

```javascript
- name: Setup WP system cron
  cron:
    name: "{{ item.key }} WordPress cron"
    minute: "*/15"
    user: "{{ web_user }}"
    job: "curl -s {{ item.value.env.wp_siteurl }}/wp-cron.php > /dev/null 2>&1"
    cron_file: "wordpress-{{ item.key | replace('.', '_') }}"
  with_dict: wordpress_sites
  when: item.value.env.disable_wp_cron | default(false) and not item.value.multisite.enabled | default(false)
```

nbyloff:
Supersedes #414 

Resolves issue #412

### DIFF
--- a/roles/wordpress-setup/tasks/main.yml
+++ b/roles/wordpress-setup/tasks/main.yml
@@ -26,7 +26,7 @@
     name: "{{ item.key }} WordPress cron"
     minute: "*/15"
     user: "{{ web_user }}"
-    job: "curl -s {{ item.value.env.wp_siteurl }}/wp-cron.php"
+    job: "curl -s {{ item.value.env.wp_siteurl }}/wp-cron.php > /dev/null 2>&1"
     cron_file: "wordpress-{{ item.key | replace('.', '_') }}"
   with_dict: wordpress_sites
   when: item.value.env.disable_wp_cron | default(false) and not item.value.multisite.enabled | default(false)


retlehs:
anyone know if we need to use/should use `wp-cron.php?doing_wp_cron` instead of just `wp-cron.php`?

swalkinshaw:
Good catch. It definitely should. WP uses it in a few places to alter behaviour: https://github.com/WordPress/WordPress/search?utf8=%E2%9C%93&q=doing_wp_cron

nbyloff:
Don't they automatically set it if it's not provided? 

[doing_cron section](https://github.com/WordPress/WordPress/blob/77e365efbf2e499e2ed11d29c101ea466cf1ceed/wp-cron.php#L71-L82)

Especially since this is set: `define('DISABLE_WP_CRON', true);`

From what I understand, that makes it so the transients aren't even used?

swalkinshaw:
@nbyloff good catch, think you're correct it's not needed.

PR #422, opened 2015-11-25
Added append_privs to WP setup - databases.

primozcigler:
This change allows that multiple WP installations share the same db
user.

Check the docs here: http://docs.ansible.com/ansible/mysql_user_module.html

For example I might have multiple WP installations, but a common scenario is to have a 1 DB user which will have access to multiple WP databases.

### DIFF
--- a/roles/wordpress-setup/tasks/database.yml
+++ b/roles/wordpress-setup/tasks/database.yml
@@ -11,6 +11,7 @@
 - name: Create/assign database user to db and grant permissions
   mysql_user: name="{{ item.value.env.db_user }}"
               password="{{ item.value.env.db_password }}"
+              append_privs=yes
               priv="{{ item.value.env.db_name | default(item.key) }}.*:ALL"
               state=present
               login_host="{{ item.value.env.db_host | default('localhost') }}"


swalkinshaw:
I would say you should have separate DB users :)

But regardless, this seems like a good default to have anyway. Thanks!

roles/wordpress-setup/tasks/database.yml
```diff
@@ -11,6 +11,7 @@
 - name: Create/assign database user to db and grant permissions
   mysql_user: name="{{ item.value.env.db_user }}"
               password="{{ item.value.env.db_password }}"
+              append_privs=true
```
swalkinshaw: Can you use `yes` instead of `true`? Matches up with Ansible docs.

primozcigler: New commit coming within minutes. I was not sure, as if you see the same page I pasted above, the example has `true`. :confused:

swalkinshaw: Yeah that's inconsistent on their part :( But they always use `yes`/`no` in the options section of their docs.

swalkinshaw:
@primozcigler would you be able to squash these commits?

primozcigler:
@swalkinshaw indeed, my bad :+1:

PR #433, opened 2015-12-07
Switch to mainline Nginx, replaces SPDY with HTTP2

QWp6t:
> ## Which Version Should I Use?
> 
> We recommend that in general you deploy the NGINX mainline branch at all times.

https://www.nginx.com/blog/nginx-1-6-1-7-released/

Because we're using mainline PPA, this also means that SPDY support is being dropped in favor of HTTP2.

#### HTTP2 vs SPDY browser compatibility
* http://caniuse.com/#feat=http2
* http://caniuse.com/#feat=spdy


swalkinshaw:
@QWp6t want to add a note about HTTP2 support in README in the "Whats Included" list?

Issue #434, opened 2015-12-13
Can't connect to diagnostic.opendns.com


joshrickert:
I haven't managed to get the provisioner to run today. It keeps failing with this error:

```
curl: (35) Server aborted the SSL handshake
fatal: [default] => Failed to template {{ lookup('pipe', 'curl -4 https://diagnostic.opendns.com/myip') }}: lookup_plugin.pipe(curl -4 https://diagnostic.opendns.com/myip) returned 35
```

I thought it might be an SSL issue, but switching to HTTP just gives me this instead:

```
curl: (52) Empty reply from server
fatal: [default] => Failed to template {{ lookup('pipe', 'curl -4 http://diagnostic.opendns.com/myip') }}: lookup_plugin.pipe(curl -4 http://diagnostic.opendns.com/myip) returned 52
```

These errors seem to inject themselves randomly throughout the provisioning script.

Should I assume it's just a connectivity issue to diagnostic.opendns.com? Any way we can work in a [fallback](http://stackoverflow.com/questions/3097589/getting-my-public-ip-via-api)?

QWp6t:
Does this work?: `{{ lookup('pipe', 'dig +short myip.opendns.com @resolver1.opendns.com') }}`

That's what we used to have, but some networks disallow UDP connections to external DNS servers, which caused it to fail.

Give it a shot to see if it works for you.

Otherwise, you can certainly use any of the others mentioned on the site you linked; this worked for me, for example: `{{ lookup('pipe', 'curl -4 https://ipinfo.io/ip') }}`.

Instead of coding in a fallback, we could probably give users the ability specify their own external IP lookup mechanism. @fullyint @swalkinshaw thoughts?

swalkinshaw:
@QWp6t I'm not opposed to having it customizable as long as it's simple.

joshrickert:
I gave it another shot today (with the default config) and Ansible ran successfully. It's probably an intermittent connectivity issue, either on the provider's end or mine.

I couldn't get any of the alternative services I mentioned working consistently either yesterday. If I spammed the `curl` command into my console I saw failures in maybe 1/10 cases.

I'll close this for now, but at least it's documented in case anyone else has issues in the future.

fullyint:
#435 offers one option for making the ip lookup customizable.

PR #435, opened 2015-12-14
Add variable for list of whitelisted IPs

fullyint:
Trellis does an IP lookup for the Ansible control machine, whitelisting it in the `ferm` and `fail2ban` roles.

This PR moves the ip lookup to a default `ip_whitelist` variable (DRY). This exposes a config for users to adjust the lookup if they have issues with network configuration (#390) or connectivity (#434). Users could also manually list IPs (vs. use lookups).

Note: To handle a temporary connectivity issue, users could define `ip_whitelist` [as quoted JSON](http://docs.ansible.com/ansible/playbooks_variables.html#passing-variables-on-the-command-line) in extra-vars on the CLI, instead of having to edit `group_vars/all/main.yml`, like this...

```
ansible-playbook server.yml -i hosts/staging -e '{"ip_whitelist":["xx.xxx.xx.xxx"]}'
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### 0.9.3: November 29th, 2015
+* Add variable for whitelisted IPs ([#435](https://github.com/roots/trellis/pull/435))
 * Nginx role improvements: use more h5bp configs ([#428](https://github.com/roots/trellis/pull/428))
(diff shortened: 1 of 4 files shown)


swalkinshaw:
:tada:

joshrickert:
Nice work on the quick fix, gentlemen.

PR #440, opened 2015-12-22
Use ipinfo.io instead of opendns for IP lookup

swalkinshaw:


austinpray:
Using this exact same config in other projects with success :+1: 

As long as @coderholic says he's not gonna pull the plug on free HTTPS I'm good to go with this

coderholic:
Yeah that's right, `/ip` isn't limited to https, and we've got no plans to change that in the future. 

If you do go head and use us we'd definitely appreciate the mention in the readme! :)

swalkinshaw:
@coderholic it actually looks like we might use https://api.ipify.org instead. I appreciate you taking the time to reply on this repo but in Ansible 2.0 there's going to be a module for IP lookup and I figure it makes more sense to use the service they default to. Unless you know a reason why we shouldn't use that site.

See http://docs.ansible.com/ansible/ipify_facts_module.html

swalkinshaw:
Closed in favour of https://github.com/roots/trellis/pull/444

PR #441, opened 2015-12-23
Fix bug in Vagrantfile where VMware provider wasn't detected properly

dblencowe:
Having ' around the VMware config in the Vagrantfile was causing it to not get picked up as a provider. I've tested this on OS X and don't think it's OS specific.


QWp6t:
Oops. My bad. I did that. :flushed: 

I'll go ahead and merge this. Thanks for the PR.

PR #442, opened 2015-12-23
Added logic to Vagrantfile to use vagrant-hostmanager plugin

dblencowe:
The vagrant-hostsupdater plugin seems to not be working with VMware Fusion
configurations. I've submitted an issue to the plugin but in the meantime
have added vagrant-hostmanager as an alternative

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -44,8 +44,14 @@ Vagrant.configure('2') do |config|
 
   if Vagrant.has_plugin? 'vagrant-hostsupdater'
     config.hostsupdater.aliases = aliases + www_aliases
+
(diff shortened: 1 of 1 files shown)


QWp6t:
:+1:

I'd be in favor of completely switching to vagrant-hostmanger. It seems like vagrant-hostsupdater is dying a slow death.

dblencowe:
@QWp6t Agreed, it's definitely not as active as hostmanager, just didn't want to step on anyones toes!

QWp6t:
Looks like we're gonna go ahead and switch off of `vagrant-hostsupdater`. See #443

PR #451, opened 2015-12-29
Improve Git deployment strategy with git archive

swalkinshaw:
Currently the deployment process looks like this:

1. Clone the repo into the `source` dir
2. Copy the `source` dir into a new release dir
3. Optionally move the `repo_subtree_path` directory to the release root
4. Optionally remove unwanted dirs in the new release

This process leads to some complications and bugs. By copying the entire
repo into the release dir and then moving the subtree, other directories
are left behind that you manually need to clean up.

There can also be conflicts if your subtree directory contains
directories with the same name as directories in the repo's root.

The new process looks like this:

1. Clone the repo into the `source` dir
2. Run `git archive` to populate the new release dir

`git archive` already supports subtrees and through `.gitattributes`, other files/folders can be excluded.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Improve Git deploy implementation via `git archive` ([#451](https://github.com/roots/trellis/pull/451))
 * Replace strip_www with optional redirect to www/non-www ([#452](https://github.com/roots/trellis/pull/452))
 * Accommodate file encryption via ansible vault ([#317](https://github.com/roots/trellis/pull/317))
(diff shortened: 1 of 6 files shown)


swalkinshaw:
Docs: https://github.com/roots/docs/pull/13

jawngee:
This doesn't copy git submodules in your repo.

retlehs:
@jawngee well, you shouldn't be using git submodules. not sure why you're commenting on this considering you already opened a thread on discourse and haven't replied to my question on there.

https://discourse.roots.io/t/submodules-weirdness/5981

Issue #312, opened 2015-08-24
Inconsistencies with add www / strip www 


retlehs:
via @fullyint:

```
(untested)
               DOES THE FORWARDING SUCCEED?

-- START URL --  ------------- GOAL URL ----------------------
                     A         B           C            D
                  http://   https://   http://www   https://www
1   http://          -         Y           N            N
2  https://          N         -           N            N
3   http://www       Y         Y           -            Y
4  https://www       N         Y           N            -

Existing Features       Missing Features
  - add_s                 - strip_s
  - strip_www             - add_www
```

kalenjohnson:
fullyint++ for nice graph

austinpray:
![image](https://cloud.githubusercontent.com/assets/2192970/9448771/0829f304-4a64-11e5-881d-223ac7207d55.png)

louim:
I think we want the `strip_www` to act like a toggle, either **forcing no** *www* or **forcing** the *www*. I don't think we want to redirect from *https* to *http*. The **bold N** are the case we would like to work.


|                       | http://             | https:// | http://www | https://www |
|-------------------|-------------------|-----------|-----------------------|--------|
| http://             |      -                 | Y       |       **N**                   |      **N** |
| https://           | N - by design |    -     |      N   - by design |       **N** |
| http://www     | Y                    |     Y    |       -                     |        Y |
| https://www   | N - by design |     Y     |      N - by design    |        - |

fullyint:
I hadn't thought we'd want any `N - by design` because it seems like a site would simply fail to load under conditions such as these:
* we use the `N - by design` for https://  -->  http://
* someone has their site listening only on port 80 at http://example.com
* a user erroneously navigates to https://example.com but we're not listening on port 443

I like the idea of thinking of it like toggles.
`www = true/false` forces www or non-www
`wordpress_sites[site].ssl.enabled = true/false` forces https or http

louim:
I still think we should be pushing our users towards HTTPS, and not allow the opposite. With https://letsencrypt.org/ around the corner, I think offering to downgrade connexions is not a good idea. 

Also, you would still need a Certificate set on the HTTPS port, because the redirect kick in only after the connexion is established. So it's a bad idea IMHO.

swalkinshaw:
Yeah they key to redirecting from `https` -> `http` is that a cert is needed anyway as @louim said. So we don't care about that.

The `www` variable (with a better name) should be moved into a WP site as well.

[...]

PR #452, opened 2015-12-29
Replace strip_www with optional redirect to www/non-www

fullyint:
Discussion in #312.

This `www_redirect` is on a per-site basis whereas `strip_www` was for all sites or none.

This includes extra logic to avoid redirecting subdomains, based on comments in #312. However, I wonder if some people would actually want `www.sub.example.com` to redirect to `sub.example.com`. If we can allow the redirect for subdomains, we could drop this portion of the conditional logic:
```
... and host.split('.') | difference('www') | count <= 2
```
(refs: [split](http://stackoverflow.com/a/30517735), [difference](http://docs.ansible.com/ansible/playbooks_filters.html#set-theory-filters), [count/length](http://jinja.pocoo.org/docs/dev/templates/#length))

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Replace strip_www with optional redirect to www/non-www ([#452](https://github.com/roots/trellis/pull/452))
 * Accommodate file encryption via ansible vault ([#317](https://github.com/roots/trellis/pull/317))
 * Fixes #353 - Allow insecure curl reqs for cron ([#450](https://github.com/roots/trellis/pull/450))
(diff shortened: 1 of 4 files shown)


retlehs:
> However, I wonder if some people would actually want www.sub.example.com to redirect to sub.example.com

:+1: ([ignore my previous comment](https://github.com/roots/trellis/issues/312#issuecomment-143929678))

fullyint:
Thanks @retlehs. I adjusted to allow redirects for subdomains:
```diff
- {% for host in item.value.site_hosts if item.value.www_redirect | default(true) and host.split('.') | difference('www') | count <= 2 %}
+ {% for host in item.value.site_hosts if item.value.www_redirect | default(true) %}
```

swalkinshaw:
Tested and works great :+1: 

Needs a docs update as well.

PR #458, opened 2016-01-06
Suppress vagrant-hostsupdater

QWp6t:
Since we're using `vagrant-hostmanager`, we can safely suppress `vagrant-hostsupdater` to prevent duplicate hosts entries when users have both plugins installed.

This is untested, but [it's consistent with their documentation](https://github.com/cogitatio/vagrant-hostsupdater#skipping-hostupdater).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Prevent duplicate hosts entries made by `vagrant-hostsupdater` ([#458](https://github.com/roots/trellis/pull/458))
 * Fix README's `ansible-playbook` command for server.yml ([#456](https://github.com/roots/trellis/pull/456)) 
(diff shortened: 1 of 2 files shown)


QWp6t:
Just tested it. It works.
```
==> default: Checking for host entries
==> default: Skipping adding host entries (config.vm.network hostsupdater: "skip" is set)
```

mAAdhaTTah:
So this is to prevent hostupdater from doing anything if it's already installed?

QWp6t:
Yeah, exactly. `vagrant-hostsupdater` automatically updates hosts files regardless of whether your `Vagrantfile` explicitly tells it to. This PR opts out of it.

swalkinshaw:
@QWp6t :+1: CHANGELOG pls

PR #460, opened 2016-01-12
Windows: Sync hosts dir with proper permissions

fullyint:
To run Ansible commands directly, Windows users have had to adjust permissions for files in the synced `hosts` directory. This PR implements @QWp6t's approach of [syncing `hosts` with the proper permissions](https://discourse.roots.io/t/remote-server-setup-on-windows/5594/2).

I bumped the Vagrant requirement to 1.8 to take advantage of the [shell provisioner](https://docs.vagrantup.com/v2/provisioning/shell.html)'s new `env` option which enables us to pass the `ANSIBLE_PATH_ON_VM` to `windows.sh`.

Could a Windows user test and confirm that this works?

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -54,6 +54,7 @@ Vagrant.configure('2') do |config|
     wordpress_sites.each_pair do |name, site|
       config.vm.synced_folder local_site_path(site), remote_site_path(name), owner: 'vagrant', group: 'www-data', mount_options: ['dmode=776', 'fmode=775']
     end
+    config.vm.synced_folder File.join(ANSIBLE_PATH, 'hosts'), File.join(ANSIBLE_PATH.sub(__dir__, '/vagrant'), 'hosts'), mount_options: ['dmode=755', 'fmode=644']
   else
     if !Vagrant.has_plugin? 'vagrant-bindfs'
       fail_with_message "vagrant-bindfs missing, please install the plugin with this command:\nvagrant plugin install vagrant-bindfs"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Windows: Sync `hosts` dir with proper permissions ([#460](https://github.com/roots/trellis/pull/460))
 * Fix `inventory_file` variable in connection tests ([#470](https://github.com/roots/trellis/pull/470))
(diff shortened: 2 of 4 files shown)


fullyint:
This PR simplifies the command at the end of `windows.sh`:
```diff
- ansible-playbook ${ANSIBLE_PATH}/dev.yml -i ${TEMP_HOSTS}/development --sudo --user=vagrant --connection=local
+ ansible-playbook dev.yml
```
- can use relative path to `dev.yml` because script already changed to that directory
- `-i` - can use [default hosts file](https://github.com/roots/trellis/blob/bc808d1f6b054167e6ca874d5948afd446c42924/ansible.cfg#L4) now that `hosts` are synced into place with the right permissions
- `--sudo` - already indicated as [play parameter](https://github.com/roots/trellis/blob/bc808d1f6b054167e6ca874d5948afd446c42924/dev.yml#L4) in `dev.yml`
- `--user=vagrant` - already indicated as [play parameter](https://github.com/roots/trellis/blob/bc808d1f6b054167e6ca874d5948afd446c42924/dev.yml#L5) in `dev.yml`
- `--connection=local` - I moved this to `hosts/development` which is only used for Windows anyway

fullyint:
This PR originally synced the `hosts` directory over the top of the vm's `hosts` directory, just so it would have specific `mount_options` (i.e., no execute bit set for `hosts` files). The latest commit removes that sync folder and instead applies these `mount_options` to the primary vagrant sync folder that is already there.

Vagrantfile
```diff
@@ -51,6 +51,7 @@ Vagrant.configure('2') do |config|
   end
 
   if Vagrant::Util::Platform.windows?
+    config.vm.synced_folder '.', '/vagrant', mount_options: ['dmode=776', 'fmode=664']
```
QWp6t: Doesn't this mean that `deploy.sh` won't execute now?

fullyint:
@QWp6t correctly pointed out that `mount_options: ['dmode=776', 'fmode=664']` on the primary vagrant sync folder would cause `deploy.sh` to not execute. I added a commit switching back to the approach of syncing the `hosts` directory with the needed permissions.

QWp6t:
Just tested this. It works. :+1:

PR #466, opened 2016-01-14
Add check before setting up permalink structure. Fixes #464

### LINKED ISSUE #464: Provisioning fails when site_install is set to false
When I set `site_install` to `false` in wordpress_sites.yml I get the following error:

screen shot 2016-01-14 at 13 59 42

This also seems to happen when multisite is enabled. From what I can gather it seems to occur because the task it tries to read the output from is skipped. 

I think I've found an easy fix for this. I'm testing it some more at the moment, if all goes well I'll submit a pull request for it.

miljan19:
This fixes issue #464. I've tested it with `site_install` set to `true` and `false`, with all other configuration set to their defaults.

Hope everything went well, this is my first pull request.

### DIFF
--- a/roles/wordpress-install/tasks/main.yml
+++ b/roles/wordpress-install/tasks/main.yml
@@ -57,7 +57,7 @@
     chdir: "{{ www_root }}/{{ item.key }}/current/"
   register: wp_permalink_results
   with_dict: wordpress_sites
-  when: "'No rewrite rules' in wp_permalink_check_results.results.0.stderr"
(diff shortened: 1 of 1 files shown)


QWp6t:
Thanks for this. A few people have had this issue with multisite and I think we were working on figuring out a way to resolve it. I'm kind of an idiot when it comes to Ansible, so I'll defer to @swalkinshaw to merge this PR. Just wanted to say thanks. :+1:

miljan19:
No problem! I'm no Ansible expert myself, so it would be good if someone more experienced could glance over it before it's merged.

swalkinshaw:
Looks good to me :+1:

PR #467, opened 2016-01-14
Fix conditional logic for permalink setup task

fullyint:
The "Setup Permalink Structure" task currently runs `when: "wp_permalink_check_results.changed`. However, `wp_permalink_check_results` will never be `changed` because of the `changed_when` rule:
```yaml
# Check Existing Permalink Structure
register: wp_permalink_check_results
changed_when: False
```
This PR updates the conditional logic.

Same as original: Sets permalink structure only if `site_install: true` and non-multisite.

Different from original: Runs only in conjunction with new install. Will never run when WP is already installed (even if permalink structure is not set). The rationale is that if WP is already installed, we don't want to potentially override configs users may have set via other means (WP GUI, via wp-cli). As an example, if a user has flushed their rewrite rules, we don't want to reset the rules.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Fix conditional logic for permalink setup task ([#467](https://github.com/roots/trellis/pull/467))
 * Fix permalink setup during WordPress Install ([#466](https://github.com/roots/trellis/pull/466))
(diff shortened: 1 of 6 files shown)


retlehs:
:+1: should also add to https://roots.io/trellis/docs/local-development-setup/

fullyint:
Good idea! Docs in roots/docs#14

PR #473, opened 2016-01-28
Fix Nginx includes for Ansible 2.0

fullyint:
1) In Ansible 1.9 the template task's `cwd` was the directory of the task file. In Ansible 2.0, it is the directory of the role. This PR enables the template task to accommodate either by using the `role_path` variable—an absolute path—instead of a relative path.

2) Ansible 2.0 changes [backreference escaping](https://github.com/ansible/ansible/commit/7f5080f64ab4a82648cb746990587c1aaff3f61d) for the [`regex_replace` filter](http://docs.ansible.com/ansible/playbooks_filters.html#other-useful-filters), needing only [2 backslashes instead of 4](https://docs.ansible.com/ansible/porting_guide_2.0.html#playbook), and in fact has a problem with 4. (Discussion at [ansible/ansible#13426](https://github.com/ansible/ansible/issues/13426#issuecomment-162249992).) ~~This PR uses a `ternary` filter to give two vs. four backslashes depending on Ansible version.~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Fix Nginx includes for Ansible 2.0 ([#473](https://github.com/roots/trellis/pull/473))
 * Use `ondrej/php` PPA since `ondrej/php-7.0` is deprecated ([#479](https://github.com/roots/trellis/pull/479))
 * Fix Ansible 2.x deploys and require version 2.x ([#478](https://github.com/roots/trellis/pull/478))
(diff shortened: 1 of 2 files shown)


remailednet:
Since ansible now requires 2.0.0.2 could the extra logic for deciding on the backslashes be dropped?

fullyint:
right on @remailednet 
I've removed the ternary that accommodated Ansible 1.9. This is thankfully much simpler now.

PR #475, opened 2016-01-30
Add option to add known_hosts on remote

fullyint:
Trellis automatically adds the key for the host contacted during `git clone` ([`accept_hostkey: yes`](https://github.com/roots/trellis/blob/11516d12e42355d0304eb4fb8c15be7ca8fcf6d4/roles/deploy/tasks/update.yml#L29)) but doesn't add keys for hosts contacted during [`composer install`](https://github.com/roots/trellis/blob/11516d12e42355d0304eb4fb8c15be7ca8fcf6d4/roles/deploy/hooks/build-after.yml#L13), so Ansible could hang/fail while waiting on this hidden message ([example](https://discourse.roots.io/t/provision-fails-on-install-dependencies-with-composer-task/5741/3)):

```
The authenticity of host ... can't be established.
RSA key fingerprint is [xyz]
Are you sure you want to continue connecting (yes/no)?
``` 

This PR extracts ssh hosts from composer.json and adds their host keys to `known_hosts`.

The `ssh-keyscan`—run by the git module and by tasks in this PR—is vulnerable to a man-in-the-middle attack. To reduce this vulnerability, the security docs will encourage users to set `accept_hostkeys: no` (a config added by this PR). Users can manually list trusted host keys in the new `group_vars/all/known_hosts.yml`.

Docs in roots/docs#15 

The ternary in the `with_items` is needed because ansible 1.9.4 doesn't have `stdout_lines` defined for vars registered in skipped tasks. Ansible will fail with error `with_items expects a list or a set`. I wasn't able to avoid the error by adding ` | default([])` at the end nor by adding `when: false`. Only the ternary seemed to work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Add option to add `known_hosts` on remote ([#475](https://github.com/roots/trellis/pull/475))
 * Update to PHP 7.0 and remove HHVM ([#432](https://github.com/roots/trellis/pull/432))
 * Windows: Sync `hosts` dir with proper permissions ([#460](https://github.com/roots/trellis/pull/460))
 * Fix `inventory_file` variable in connection tests ([#470](https://github.com/roots/trellis/pull/470))
--- a/group_vars/all/known_hosts.yml
+++ b/group_vars/all/known_hosts.yml
@@ -1,0 +1,10 @@
+# Documentation: https://roots.io/trellis/docs/security/#known-hosts
+
+# Host keys to add to known_hosts, e.g.,
+#   - git host for Bedrock-based project (`repo` variable in `wordpress_sites`)
+#   - git hosts in Bedrock project's composer.json
+known_hosts:
+  - name: github.com
+    key: github.com ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEAq2A7hRGmdnm9tUDbO9IDSwBK6TbQa+PXYPCPy6rbTrTtw7PHkccKrpp0yVhp5HdEIcKr6pLlVDBfOLX9QUsyCOV0wzfjIJNlGEYsdlLJizHhbn2mUjvSAHQqZETYP81eFzLQNnPHt4EVVUh7VfDESU84KezmD5QlWpXLmvU31/yMf+Se8xhHTvKSCZIFImWwoG6mbUoWf9nzpIoaSjB+weqqUUmpaaasXVal72J+UX2B+2RPW3RcT0eOzQgqlJL3RKrTJvdsjE3JEAvGq3lGHSZXy28G3skua2SmVi/w4yCE6gbODqnTWlg7+wC604ydGXA8VJiS5ap43JXiUFFAaQ==
+  - name: bitbucket.org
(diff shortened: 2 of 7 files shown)


swalkinshaw:
Wondering if the retrieve and add known hosts tasks for Composer would be better suited for a simple module? Those piped shell commands get a little hairy for Ansible and it's also duplicated in two places.

fullyint:
Closing. I will resubmit a revision that omits the option of auto-detecting and adding hostkeys for hosts in composer.json. It seems like a bad idea do a lot of extra work to create a feature that makes people vulnerable to MITM attacks (while running `ssh-keyscan -t` to collect hostkeys during provision). Better to just stay secure. The revision will allow users to manually list hostkeys in a new `known_hosts` list variable, then Trellis will load them.

chetzof:
Hey, was that implemented?

swalkinshaw:
@chetzof no

fullyint:
I no longer plan to revise and resubmit. Instead, users can implement this using Trellis [deploy hooks](https://roots.io/trellis/docs/deploys/#hooks). 

**1 - Define known_hosts in `group_vars/all/main.yml`** (or maybe `deploy.yml`), e.g., 
```
# Host keys to add to known_hosts, e.g.,
#   - git host for Bedrock-based project (`repo` variable in `wordpress_sites`)
#   - git hosts in Bedrock project's composer.json
known_hosts:
  - name: github.com
    key: github.com ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEAq2A7hRGmdnm9tUDbO9IDSwBK6TbQa+PXYPCPy6rbTrTtw7PHkccKrpp0yVhp5HdEIcKr6pLlVDBfOLX9QUsyCOV0wzfjIJNlGEYsdlLJizHhbn2mUjvSAHQqZETYP81eFzLQNnPHt4EVVUh7VfDESU84KezmD5QlWpXLmvU31/yMf+Se8xhHTvKSCZIFImWwoG6mbUoWf9nzpIoaSjB+weqqUUmpaaasXVal72J+UX2B+2RPW3RcT0eOzQgqlJL3RKrTJvdsjE3JEAvGq3lGHSZXy28G3skua2SmVi/w4yCE6gbODqnTWlg7+wC604ydGXA8VJiS5ap43JXiUFFAaQ==
  - name: bitbucket.org
    key: bitbucket.org ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEAubiN81eDcafrgMeLzaFPsw2kNvEcqTKl/VqLat/MaB33pZy0y3rJZtnqwR2qOOvbwKZYKiEO1O6VqNEBxKvJJelCq0dTXWT5pbO2gDXC6h6QDXCaHo6pOHGPUy+YBaGQRGuSusMEASYiWunYN0vCAI8QaXnWMXNMdFP3jHAJH0eDsoiGnLPBlBp4TNm6rYI74nMzgz3B9IikW4WVK+dc8KZJZWYjAuORU3jc1c/NPskD2ASinf8v3xnfXeukU0sJ5N6m5E8VLjObPEO+mN2t/FZTMZLiFqPWc/ALSqnMnnhwrNi2rbfg/rd/IpL8Le3pSBne8+seeFVBoGqzHM9yXw==
```

**2 - Define the deploy hook variable under `vars:` in `deploy.yml`:**
```
deploy_update_before: "{{ playbook_dir }}/deploy-hooks/update-before.yml"
```

**3 - Create a new file: `deploy-hooks/update-before.yml`** with these contents:
```
---
- name: Add known_hosts
  known_hosts:
    name: "{{ item.name }}"
    key: "{{ item.key | default(omit) }}"
    path: "{{ item.path | default('/home/' + web_user + '/.ssh/known_hosts') }}"
    state: "{{ item.state | default('present') }}"
  with_items: "{{ known_hosts | default([]) }}"
```

Notes:
- This deploy hook approach will only affect remote servers, not the `known_hosts` on the local dev VM. 
- This demos `name` and `path` options but will accommodate [other options](http://docs.ansible.com/ansible/known_hosts_module.html) if defined.
- You could set [`accept_hostkey: no`](https://github.com/roots/trellis/blob/30ec2d9f3917bde14cd2098a7e96a27163ba08f8/roles/deploy/tasks/update.yml#L29) so that the clone task won't run a `ssh-keyscan` which could be vulnerable to a MITM attack.

PR #479, opened 2016-02-05
updates ppa:ondrej/php-7.0 to ppa:ondrej/php

cam5:
Ondřej Surý's PPA now reads: "(deprecated use pap:ondrej/php)" in the description, and I can see that he modified the php7.0 packages a couple of hours ago. Apt had trouble resolving them for me when I provisioned a machine this morning but didn't have the issue when I tried again a little later. Regardless, we ought to update a deprecated PPA, no?

### DIFF
--- a/roles/php/tasks/main.yml
+++ b/roles/php/tasks/main.yml
@@ -1,7 +1,7 @@
 ---
 - name: Add PHP 7.0 PPA
   apt_repository:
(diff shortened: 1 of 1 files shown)


swalkinshaw:
Thanks! Merging this should I agree we should definitely switch.

PR #484, opened 2016-02-10
Add note about another bad characters in salts

dmgawel:
Characters `{{` causes fail2ban error: `fatal: [XXX.XXX.XXX.XXX] => Failed to template`. I've added note about it.

### DIFF
--- a/group_vars/production/vault.yml
+++ b/group_vars/production/vault.yml
@@ -11,7 +11,7 @@ vault_wordpress_sites:
     env:
       db_password: example_dbpassword
       # Generate your keys here: https://api.wordpress.org/secret-key/1.1/salt/
-      # These CANNOT contain the characters "{%" in succession
+      # These CANNOT contain the characters "{%" or "{{" in succession
       auth_key: "generateme"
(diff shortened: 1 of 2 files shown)


Foxaii:
Thanks for highlighting this as it can definitely be clarified, but I don't think that your wording is any more correct.

`These CANNOT contain the characters "{%" in succession`
I believe this means that `{` and `%` cannot be doubled up i.e. `{{` or `%%` will be parsed incorrectly.

Whereas your wording could be interpreted as you cannot have `{%` or `{{` in succession i.e. `{%{%` or  `{{{{` or `{%{{` or `{{{%` will be parsed incorrectly.

Whatever the note gets changed to should include examples, that way you can leave out the semantics.

swalkinshaw:
@Foxaii 

> I believe this means that { and % cannot be doubled up i.e. {{ or %% will be parsed incorrectly.

It's supposed to mean that `{%` can't appear exactly like that as a combination/group. `%%` shouldn't be a problem. `{{` obviously would though.

swalkinshaw:
@dmgawel thanks!

PR #486, opened 2016-02-12
Add reverse_www filter to fix www_redirect

fullyint:
@QWp6t pointed out that **when SSL is enabled**, `wordpress-site.conf.j2` will fail to forward http://www.example.com => https://example.com. The current template produces the conf below. 

```
server {
  listen 80;
  server_name example.com;
  return 301 https://$host$request_uri;
}

server {
  listen 443 ssl http2;
  server_name www.example.com;
  return 301 $scheme://example.com$request_uri;
}
```
- The first server block doesn't match http://www.example.com because it doesn't include a match for www. 
- The second server block doesn't match http://www.example.com because it listens only on 443.

This PR enables the first server block to also match www, taking the first step of forwarding http://www.example.com => https://www.example.com. Then the second server block will pick it up and strip the www.

I created and implemented a custom `reverse_www` filter to make the templating simpler.

### DIFF
--- a/filter_plugins/trellis_filters.py
+++ b/filter_plugins/trellis_filters.py
@@ -1,0 +1,38 @@
+# Make coding more python3-ish
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+import types
+
+from ansible import errors
+from ansible.compat.six import string_types
+
+def reverse_www(value):
+    ''' Add or remove www subdomain '''
+
+    # Check if value is a list and parse each item
+    if isinstance(value, (list, tuple, types.GeneratorType)):
+        values = []
+        for item in value:
+            values.append(reverse_www(item))
+        return values
+
+    # Add or remove www
+    elif isinstance(value, string_types):
+        if value.startswith('www.'):
+            return value[4:]
+        else:
+            return 'www.{0}'.format(value)
+
+    # Handle invalid input type
+    else:
+        raise errors.AnsibleFilterError('The reverse_www filter expects a string or list of strings, got ' + repr(value))
+
+
+class FilterModule(object):
+    ''' Trellis jinja2 filters '''
+
+    def filters(self):
+        return {
+            'reverse_www': reverse_www,
+        }
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,3 +9,6 @@ end_of_line = lf
 charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
+
+[*.py]
+indent_size = 4
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .vault_pass
 .vagrant
 vendor/roles
+*.py[co]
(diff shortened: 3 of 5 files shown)


filter_plugins/trellis_filters.py
```diff
@@ -0,0 +1,20 @@
...
+__metaclass__ = type
+
+import re
+
+def reverse_www(host=None):
+    ''' Add or remove the www subdomain '''
+    match = re.match(r'www\.(.+)', host)
+    if match:
+        return match.group(1)
+    return 'www.' + host
```
swalkinshaw: Not sure what idiomatic Python is but seems like this should be under the `else` condition.

fullyint: Excellent, thank you. I see this in [PEP 08](https://www.python.org/dev/peps/pep-0008/#programming-recommendations):
```python
# No
def bar(x):
    if x < 0:
        return
    return math.sqrt(x)

# Yes
def foo(x):
    if x >= 0:
        return math.sqrt(x)
    else:
        return None
```

Another improvement is that although [re.match()](https://docs.python.org/2/library/re.html#re.match) will...
> Return `None` if the string does not match the pattern

I should probably change the evaluation to `if match is not None:` because [PEP 8](https://www.python.org/dev/peps/pep-0008/#programming-recommendations) also says this:
> beware of writing if `x` when you really mean `if x is not None` -- e.g. when testing whether a variable or argument that defaults to None was set to some other value. The other value might have a type (such as a container) that could be false in a boolean context!

fullyint:
Additional commit
 - code style better resembles python and Ansible conventions
 - filter now accommodates one or more host names
 - filter provides informative error message if input value is not string or list of strings

Pros of new filter
 - makes the `wordpress-site.conf.j2` template much more readable, sparing most users from seeing elaborate jinja filter sequences
 - initiates the use of custom filters, bringing their power to Trellis

Cons of new filter
- adds far more lines of code overall than if we stick to existing filters
- probably wouldn't be used anywhere but in this instance, so it is not a benefit to DRY code 

Below is an alternative to this PR's `reverse_www` filter. It may be the better solution for the www redirects issue at hand.
```diff
 {% if item.value.ssl is defined and item.value.ssl.enabled | default(False) %}
 server {
   listen 80;
-  server_name {{ item.value.site_hosts | join(' ') }};
+  server_name {{ item.value.site_hosts | join(' ') }}{% for host in item.value.site_hosts %} {{ (host[:3] == 'www.') | ternary(host[4:], 'www.' + host) }}{% endfor %};
   return 301 https://$host$request_uri;
 }
 {% endif %}
 
 {% for host in item.value.site_hosts if item.value.www_redirect | default(true) %}
 server {
   {% if item.value.ssl is defined and item.value.ssl.enabled | default(false) -%}
     listen 443 ssl http2;
   {% else -%}
     listen 80;
   {% endif -%}
 
-  server_name {{ host | match('^www\\.(.*)') | ternary(host | regex_replace('^www\\.(.*)', '\\1'), 'www.' + host ) }};
+  server_name {{ (host[:3] == 'www.') | ternary(host[4:], 'www.' + host) }};
   return 301 $scheme://{{ host }}$request_uri;
 }
 {% endfor %}
```

PR #493, opened 2016-02-17
Keep Composer updated

retlehs:
accidentally removed composer auto-updating with #491, this re-enables it with the new role we're using


swalkinshaw:
:shipit:

PR #503, opened 2016-02-24
Added necessary php modules

BrandonShutter:
Fixing composer issues, as well as adding support for xml.

https://discourse.roots.io/t/call-to-undefined-function-xml-parser-create/6013
https://discourse.roots.io/t/composer-wp-packagist-failure-to-download-on-trellis-deploy/6018/


swalkinshaw:
Thanks!

PR #504, opened 2016-02-25
Add php7.0-mbstring

BrandonShutter:
https://discourse.roots.io/t/uncaught-error-call-to-undefined-function-mb-detect-encoding/6034


swalkinshaw:
Thanks!

PR #512, opened 2016-03-06
fixes add-apt-repository installation for VMWare's box

ckovey:
apt-get needs updating prior to trying to install add-apt-repository otherwise it fails finding it, which in turn causes ansible not be installed to the correct version, which results in the error 'provided hosts list is empty'

See https://discourse.roots.io/t/error-provided-hosts-list-is-empty/6127/4

### DIFF
--- a/windows.sh
+++ b/windows.sh
@@ -23,6 +23,7 @@ printf "\033[0m\n\n"
 
 # Check that add-apt-repository is installed for non-standard Vagrant boxes
 if [ ! -f /usr/bin/add-apt-repository ]; then
+  sudo apt-get -y update
   echo "Adding add-apt-repository..."
   sudo apt-get -y install software-properties-common
 fi


windows.sh
```diff
@@ -23,6 +23,7 @@ printf "\033[0m\n\n"
 
 # Check that add-apt-repository is installed for non-standard Vagrant boxes
 if [ ! -f /usr/bin/add-apt-repository ]; then
+  sudo apt-get update
   echo "Adding add-apt-repository..."
   sudo apt-get -y install software-properties-common
 fi
```
fullyint: There's another `sudo apt-get -y update` in the following conditional that installs Ansible. I suspect that both conditionals would typically run on the same occasion.

What do you think about combining the two conditionals with something like this?
```
if [ ! -f /usr/bin/add-apt-repository ] || [ ! -f /usr/bin/ansible ]; then
```
Then you could have the `sudo apt-get -y update` just once at the top of the combined conditional.

ckovey: I had thought about that, because they both do indeed run as you suspect on first `vagrant up` and adds to your initial provisioning time.

What I couldn't figure out was how to do both in one shot avoiding two `update`'s.  I needed the first one so I could first install `software-properties-common` (this problem is what started the linked discoure thread).  After adding the ansible repository you have to run `update` again to then install ansible itself.

I will push a change adding `-y` to the command to match the below one

fullyint: Ah, I see. Excellent. Thanks!

swalkinshaw:
Thanks!

PR #513, opened 2016-03-09
Add ansible playbook for windows-specific vagrant development

henkler:
When doing development with Sage on Windows using Trellis, we need at a bare minimum the following installed on the development Vagrant instance:

Node
Gulp (global)
Bower (global)

I've created a simple playbook and modifications to windows.sh to automatically pull these in on vagrant up.

### DIFF
--- a/roles/windows-dev/tasks/main.yml
+++ b/roles/windows-dev/tasks/main.yml
@@ -1,0 +1,11 @@
+---
+- name: Install "gulp" node.js package
+  npm:
+    name: gulp
+    global: yes
+
+- name: Install "bower" node.js package
+  npm:
+    name: bower
+    global: yes
+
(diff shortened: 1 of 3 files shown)


retlehs:
we recommend you run those on the host machine, not on the vagrant development box

nonameolsson:
I did try to have that same workflow, with node/gulp/bower, on a Vagrant machine once. My experience with the performance was awful! It was really slow.

Running those stuff on the host machine will give a much faster performance. @henkler What is the reason that you want those things to be installed in the Vagrantmachine if the host is running Windows? I'm just curious :)

henkler:
Actually, I didn't realize node/gulp/bower would actually build correctly without pulling in the entire php dev stack (composer, etc).

Tried it and it seems to be working.  Much less painful than doing it through Vagrant and mitigating symlink issues.  Thanks for pointing me in the right direction.

nonameolsson:
@henkler great! :)

PR #514, opened 2016-03-14
Remove execute permissions from files which don't need them

tremby:


swalkinshaw:
Good catch. Never noticed until now that those files are red in my terminal :unamused:

tremby:
Green in mine. Always bothers me. ☺

Issue #328, opened 2015-09-03
Let’s Encrypt integration


swalkinshaw:
Placeholder issue to investigate easier SSL certificate integration/creation once Let's Encrypt launches: https://letsencrypt.org/

louim:
:+1: :heart_eyes:

retlehs:
if anyone hears or sees anything about beta signups, please let us know!

> According to https://letsencrypt.org/2015/08/07/updated-lets-encrypt-launch-schedule.html72, the Let's Encrypt CA is planning for general availability of its services in the week of November 16, 2015. If you use the client before then (**except as part of a beta test program**), you would receive a test certificate that is not signed by a publicly-trusted CA, and that is not accepted by browsers.

related discussion: https://community.letsencrypt.org/t/is-there-a-beta-test-program-i-can-sign-up-for/119

retlehs:
https://github.com/thefinn93/ansible-letsencrypt

mxxcon:
According to https://letsencrypt.org/2015/11/12/public-beta-timing.html
>Let’s Encrypt will enter Public Beta on December 3, 2015. Once we’ve entered Public Beta our systems will be open to anyone who would like to request a certificate. There will no longer be a requirement to sign up and wait for an invitation.

mAAdhaTTah:
Looks like someone is developing a wp-cli command: https://github.com/tollmanz/lets-encrypt-wp

Maybe we should use that, and add the related cronjob so it auto-renews every 90 days?

kalenjohnson:
Applicable: https://deliciousbrains.com/http2-https-lets-encrypt-wordpress/

swalkinshaw:
I highly doubt we'll want to use the WP plugin. I'm assuming that's more for people without the benefit of tools like Ansible :)

mAAdhaTTah:
@kalenjohnson that was my source for the plugin :D

evanfuture:
On a live server, I've had no trouble setting up my certs (well, with Apache anyway).  The trouble comes in development, when I'm using a non-existent url, site_slug.dev, and it won't complete.  Other than that, it seems easy enough to ansible-ize the whole process... (he says confidently, then, weeks later, shakes his head in shame)

joshrickert:
Seems more practical to use LetsEncrypt for the live server and a self-signed cert for the development VM.

mxxcon:
[...]

PR #518, opened 2016-03-18
Let’s Encrypt integration

swalkinshaw:
Issue: https://github.com/roots/trellis/issues/328

This PR adds Let's Encrypt integration. It will automatically create a valid, production-ready SSL certificate through LE and configure your WP site to use it.

These certifications expire every 90 days so there's a cronjob to renew them monthly.

There's a new `provider` option under the `ssl` object for each WP site. Valid options are:

* `self-signed` - used in development
* `manual` - when you already have a key and cert you just want copied over
* `letsencrypt` - invokes automatic LE certificate flow

### Testing

*  You need a publicly accessible server
* Make sure a WP site has
```yaml
ssl:
  enabled: true
  provider: letsencrypt
```
* Optionally remove `letsencrypt_ca` from `group_vars/all/main.yml` if you want a real production cert. Otherwise you are using Let's Encrypt's staging server and the CA isn't verified in browsers. Note: LE production has rate limits. So don't exhaust the limit just testing unless you need a real cert.
* Provision server as usual

Todo:
- [x] Renewal script does not generate bundled certificate
- [x] Improve idempotence
- [x] Only run `letsencrypt` (toggle via variable)
- [x] Add the "reverse" (www or non-www) hosts for multi domain certs
- [x] More testing
- [x] Set default CA to production
- [x] Docs - https://github.com/roots/docs/blob/ssl-docs-add-providers/trellis/ssl.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Let's Encrypt integration ([#518](https://github.com/roots/trellis/pull/518))
 * Improve Git repo format validation [#516](https://github.com/roots/trellis/pull/516)
 * Fix #505 - Git ignore \*.retry file
 * Fix Ansible deprecations for bare variables ([#510](https://github.com/roots/trellis/pull/510))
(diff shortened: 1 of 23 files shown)


richvida:
Great work!

swalkinshaw:
Updated this with ea20374 thanks to @fullyint so that the manual `letsencrypt_enabled` variable isn't needed anymore. We'll automatically detect if any WP sites use `letsencrypt`.

PR #519, opened 2016-03-18
private_network is specific to Virtualbox

DavidSchargel:
Move private_network config into virtualbox block since it is not needed for VMware. Minor, but eliminates NFS warning messages on VMware 'vagrant up'.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -39,9 +39,6 @@ Vagrant.configure('2') do |config|
   # https://github.com/mitchellh/vagrant/issues/1673#issuecomment-28288042
   config.ssh.shell = %{bash -c 'BASH_ENV=/etc/profile exec bash'}
 
-  # Required for NFS to work
-  config.vm.network :private_network, ip: ip, hostsupdater: 'skip'
-
   hostname, *aliases = wordpress_sites.flat_map { |(_name, site)| site['site_hosts'] }
   config.vm.hostname = hostname
   www_aliases = ["www.#{hostname}"] + aliases.map { |host| "www.#{host}" }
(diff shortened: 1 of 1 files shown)


DavidSchargel:
My first PR ever…hope I did this right.

DavidSchargel:
As per https://www.vagrantup.com/docs/synced-folders/nfs.html, private_network is a prerequisite only due to a limitation of VirtualBox's built-in networking, not for VMware.

austinpray:
Doesn't the fixed IP have to be there for hostsupdater to work?

DavidSchargel:
Hmmm… @austinpray…I'm not even sure if that `hostsupdater: 'skip'` works any longer with the switch from vagrant-hostsupdater to vagrant-hostmanager a few months ago in #442. Since I'm not on VirtualBox, I can't test to see what happens definitively, but I would suspect that `hostsupdater: 'skip'` isn't working any longer anyway. 

However, for similar hostmanager functionality, I wonder if Virtualbox users should see the addition of `config.hostmanager.ignore_private_ip = false` to the config.hostmanager block of the VagrantFile as per the [vagrant-hostmanager docs](https://github.com/devopsgroup-io/vagrant-hostmanager#usage). I would think that VirtualBox users would have reported something earlier if this was the case.

QWp6t:
`hostsupdater: 'skip'` was added _because_ we switched to hostmanager. https://github.com/roots/trellis/pull/458

DavidSchargel:
Got it; I no longer have vagrant-hostsupdater installed. It obviously needs to stay in place then. 

Since 1) that `hostsupdater: "skip"` line is only available when used on the `config.vm.network :private_network` entry I am moving, and 2) my PR is VMware-specific and only suppresses a warning message, let's not make that change. How do I withdraw/close this PR?

PR #521, opened 2016-03-21
Disable MariaDB binary logging by default

### LINKED ISSUE #520: MariaDB config has binlogging enabled by default
I have a 100MB database, and my production server fell over in the middle of the night after `/var/log/mysql` filled up with `mariadb-bin` files. I don't need binary logs because I'm not doing replication, and I expect that's true of a majority of users.

Per this [StackOverflow](http://dba.stackexchange.com/a/120693) answer, it is possible to define a custom override in `/etc/mysql/conf.d` which could either disable binlogging, or set the log retention period to something less that the 10 days it currently stands at.

Ansible novice here, but happy to attempt a PR if this would be useful.

duncanjbrown:
Fixes #520

### DIFF
--- a/roles/mariadb/tasks/main.yml
+++ b/roles/mariadb/tasks/main.yml
@@ -25,10 +25,10 @@
     group: root
   when: binary_logging_disabled
 
-- name: Start MariaDB MySQL Server
+- name: Restart MariaDB MySQL Server
   service:
     name: mysql
-    state: started
+    state: restarted
     enabled: true
 
 - name: Set root user password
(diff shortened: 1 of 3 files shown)


roles/mariadb/tasks/main.yml
```diff
@@ -17,10 +17,18 @@
     name: mariadb-server
     state: present
 
-- name: Start MariaDB MySQL Server
+- name: Disable MariaDB binary logging
+  when: binary_logging_disabled
```
swalkinshaw: Small thing but can the `when` condition be put after the `template` action? That way it's consistent with the rest of the codebase.

swalkinshaw:
Thanks @duncanjbrown. Looks good other than my one comment :+1:

duncanjbrown:
Updated

swalkinshaw:
Thanks :)

Issue #391, opened 2015-10-23
Consider using WP-CLI nightly for local dev


danielbachhuber:
Hi, would you consider using WP-CLI nightly for dev? It's stable enough for local development, and is compatible with WordPress trunk. You just need to replace https://raw.githubusercontent.com/wp-cli/builds/gh-pages/phar/wp-cli.phar with https://raw.githubusercontent.com/wp-cli/builds/gh-pages/phar/wp-cli-nightly.phar

swalkinshaw:
Only if you promise there's no bugs :)

But yeah this is a good idea for development only. See https://github.com/roots/trellis/pull/392

austinpray:
Probably just use nightly everywhere, right? I test my scripts on development and then promote them to production. I don't want my script failing due to it relying on nightly behavior when production has an earlier release.

retlehs:
should we use nightly at all? if we want parity then we should be defining specific versions of wp-cli to use across all environments.

if we switch to nightly, someone could very easily wind up with one nightly version of wp-cli on dev that doesn't match up with remote (or vice-versa).

austinpray:
Oh shoot. Ben is totally right.

danielbachhuber:
> if we switch to nightly, someone could very easily wind up with one nightly version of wp-cli on dev that doesn't match up with remote (or vice-versa).

Why not use nightly for dev, and the latest stable everywhere else?

retlehs:
https://roots.io/twelve-factor-10-dev-prod-parity/

you could end up in a situation where dev works but prod doesn't

for instance - isn't wp-cli going to require a new min version once wp 4.4 is out? what if your remote server has the older version? we need dev and remote to be the same

mAAdhaTTah:
It doesn't look like the task is currently locked to a version as it is. If someone provisions their remote machine and their local env between version releases, they'll have different versions.

retlehs:
correct, #392 should be updated to specify a version

swalkinshaw:
Unfortunately there's no way to link a specific version from the WP-CLI builds at https://github.com/wp-cli/builds/tree/gh-pages/phar.

@danielbachhuber any suggestions? Maybe tags on that repo for versions?

QWp6t:
@swalkinshaw why not just use https://github.com/wp-cli/wp-cli/releases ?

https://github.com/wp-cli/wp-cli/releases/download/v0.20.4/wp-cli-0.20.4.phar
[...]

PR #523, opened 2016-03-22
Fix some deprecation warnings while deploying and use current stable WP-CLI

mockey:
WP-CLI URL taken from: http://wp-cli.org/#install


swalkinshaw:
Re: the WP-CLI link. We had discussion about it in https://github.com/roots/trellis/issues/391 and decided to lock the version down.

retlehs:
@mockey WP-CLI 0.23.0 came out yesterday, can you please update this PR to use that version?

then squash the commits and this is good to go :+1:

mockey:
OK like this?

retlehs:
thanks!

PR #528, opened 2016-03-25
wordpress_sites improvements

swalkinshaw:
* Add `to_env` filter
* Use `combine` + `to_env` for better .env templates
* Remove the need for `wp_env` and set automatically based on `env`
* Set `disable_wp_cron` by default

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Simplify and improve `wordpress_sites` with better defaults ([#528](https://github.com/roots/trellis/pull/528))
 * Improve Let's Encrypt challenge pre-flight tests ([#526](https://github.com/roots/trellis/pull/526))
 * `reverse_www` filter improvements (ignore subdomains) ([#525](https://github.com/roots/trellis/pull/525))
 * Fix #520 - Disable MariaDB binary logging by default ([#521](https://github.com/roots/trellis/pull/521))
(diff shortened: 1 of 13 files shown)


swalkinshaw:
Some doc updates should accompany this.

henkler:
I think there may be an issue here with 'roles/deploy/templates/env.j2'.  Getting a failure on deploy.

failed: [staging.example.com] => (item={u'dest': u'.env', u'src': u'roles/deploy/templates/env.j2', u'name': u'.env config'}) => {"failed": true, "item": {"dest": ".env", "name": ".env config", "src": "roles/deploy/templates/env.j2"}, "msg": "AnsibleUndefinedVariable: 'dict object' has no attribute 'value'"}

henkler:
Maybe this instead for env.j2?
```
{{ wordpress_env_defaults | combine(wordpress_sites[site].env, vault_wordpress_sites[site].env) | to_env }}
```

vdrnn:
Yep, ran into the exact same issue after updating an existing installation. 
It goes through with @henkler's proposition `{{ wordpress_env_defaults | combine(wordpress_sites[site].env, vault_wordpress_sites[site].env) | to_env }}`

fullyint:
Thank you @henkler for helping catch that this `env.j2` template in the `deploy` role must be different from the `env.j2` template in the `wordpress-install` role.

Your edit works. There is a bit of convention in the `deploy` role to prefer using the `project` dict, which is only slightly different:
```
{{ wordpress_env_defaults | combine(project.env, vault_wordpress_sites[site].env) | to_env }}
```

fullyint:
@henkler Would you be interested in submitting a pull request for this?

henkler:
Sure.  Let me run another deploy first and I'll get it submitted.

henkler:
PR #530 submitted.

PR #530, opened 2016-03-26
Fix deploy env template to use valid ansible vars

henkler:
When running deploy.sh, ansible fails with an error similar to the following:
```
failed: [staging.example.com] => (item={u'dest': u'.env', u'src': u'roles/deploy/templates/env.j2', u'name': u'.env config'}) => {"failed": true, "item": {"dest": ".env", "name": ".env config", "src": "roles/deploy/templates/env.j2"}, "msg": "AnsibleUndefinedVariable: 'dict object' has no attribute 'value'"}
```

It apepars that PR #528 introduced an error with some invalid ansible keys.  This should fix them.

Thanks to @fullyint for helping with Ansible conventions.  I'm new to Trellis (and Ansible), and you make it easy!


fullyint:
Thank you @henkler !

PR #531, opened 2016-03-28
Add custom formatting to Ansible output

fullyint:
This PR proposes more control over Ansible's CLI output. This requires nothing from users.

**Verbosity.** The docs at roots/docs#28 show examples of reducing verbosity.

**Format and color.** The playbook below demonstrates format and color. To try it, pull this branch, copy the code below to a file named `output_demo.yml` in the branch root, and run `ansible-playbook output_demo.yml`. Also run `ansible-playbook output_demo.yml -e custom_output=false` to see this PR's features disabled.

```
- name: Output demo
  gather_facts: false
  hosts: localhost
  tasks:
    - name: Show available Ansible colors and current color settings
      debug:
        msg: |
          AVAILABLE COLORS
          ----------------
          {{ ansible_colors }}

          CURRENT COLOR SETTINGS
          ----------------------
          {{ color_settings }}

    - name: Demonstrate a few colors in context
      debug:
        msg: |
          Here is the default color. Color will usually be determined dynamically based on event (e.g., debug message, failed task, etc.).

          C_WARNHere is the warning colorC_WARN, regular text, C_ERRORerror textC_ERROR, and C_OKok textC_OK.

          Here is `code in backticks`. If `colorize_code = False` (in `trellis.cfg`) then the backticks will just display, instead of being replaced with ANSI escape codes for color.

          URLs look nice in a slightly different shade, e.g., C_BOLDhttps://roots.io/trellis/docs/ssl/C_BOLD

    - name: Demonstrate the failure output
      fail:
        msg: |
          Please specify SSL `provider`:
          ```
          wordpress_sites:
            example.com:
              ssl:
                enabled: true
                provider: C_OKletsencryptC_OK

          `Backticks` always display inside fenced code blocks and are not colorized. Codeblocks do not wrap, which is good for displaying commands users may copy/paste.
          ```
```

**Basic features**
* pretty prints `msg` and error output
* if error, prints user's version info for Trellis, Ansible, and Vagrant (for pasting to help forum)
* cleans up colors and formatting for output from Vagrant shell provisioner (`windows.sh`)
* wraps wp_env keys and salts in `{% raw %}` so users no longer have to avoid `{{` and `{%`
* includes options to suppress output that is typically unnecessary (include tasks, skipped task info, displaying entire dicts like `wordpress_sites`, etc.)
* users can omit parameters from `trellis.cfg` and/or override via environment variables
* users can use extra vars to one-off enable/disable via `custom_output` and/or adjust `wrap_width`

**Structure via new `lib/trellis` directory**
```
lib/
  trellis/
    modules/
      deploy_helper.py
    plugins/
      callback/
        __init__.py         <- TrellisCallbackBase
        output_overrides.py
        90_output_format.py
      filter/
        trellis_filters.py
```
* structure mirrors Ansible's `lib/ansible` directory
* tidies up: pulls python files into `lib/trellis` (modules, filter plugin, callback plugins)
* enables us to import `TrellisCallbackBase` from `trellis.plugins.callback`

The two callback plugins here pull in shared code from `TrellisCallbackBase` in `__init__.py`. The `__init__.py` file [will be ignored](http://github.com/ansible/ansible/blob/27b02e8f6cabb343d06c36569b47022728381635/lib/ansible/plugins/__init__.py#L345) by Ansible's glob loading of callback plugins. Two future plugins could use `TrellisCallbackBase` too (custom vars and custom messages).

The `90_` prefix is include in `90_output_format.py` because Ansible loads callback plugins using [glob](https://github.com/ansible/ansible/blob/2f118e546f9d899cd52994c6149e0b2f58d14102/lib/ansible/plugins/__init__.py#L164), which is [sorted](https://github.com/ansible/ansible/blob/2f118e546f9d899cd52994c6149e0b2f58d14102/lib/ansible/plugins/__init__.py#L356). The prefix helps establish execution order, following the model of [sysctl.d](http://manpages.ubuntu.com/manpages/wily/en/man5/sysctl.d.5.html): "It is recommended to prefix all filenames with a two-digit number and a dash, to simplify the ordering of the files." The point is that future plugins should probably run first by having a lower value prefix, processing whatever data before the output_format plugin formats it.

I don't understand the point of the `CALLBACK_TYPE` attribute except to limit to [just one of type `stdout`](https://github.com/ansible/ansible/blob/4d42f1992b1df254e4c18dd983b41fef5d384094/lib/ansible/executor/task_queue_manager.py#L166). The one callback plugin with `CALLBACK_TYPE = 'stdout'` in this PR is `output_overrides.py`. This fact makes it not subject to the glob loading/sorting that requires an integer prefix to the filename (mentioned above). As of Ansible 2.0.1 it will always execute first among the callback plugins.

It would help to have a few people confirm that this doesn't break anything for them. The playbook above is useful for testing. Also, here's an easy way to see errors with `server.yml`:
```
ansible-playbook server.yml -e "env=production minimum_ansible_version=5"
```


fullyint:
I am closing this and will break it into smaller chunks instead. The first is in #538.

PR #533, opened 2016-03-28
Automatically set `wp_home` and `wp_siteurl`

swalkinshaw:
Using the "dynamic" HTTP_HOST/SERVER_NAME is a good enough default. In
the majority of cases users should have to worry about setting these
variables and overriding them.

This also allows for a more seamless transition to HTTPS since there's
one less place to configure or potentially forget. It also just reduces
the chances of misconfiguration in general and a mismatch of values
(site hosts and wp home/siteurl).


jalada:
With this change nothing works. All my Wordpress links etc. have literal ``${WP_HOME}` etc. in them. Confused...

PR #537, opened 2016-03-30
Multiple SSL sites redirect problem

swaincreates:
So I had a problem with two sites using SSL wildcard certs. Site 2 would not redirect the www version because it was using Site 1 wildcard. If I specify the certs and include ssl conf files in redirect block then everything works fine. Discourse: [here](https://discourse.roots.io/t/2-different-ssl-wildcards-issue-on-same-server/6333). Not sure if I did this exactly right because ansible/jinja syntax and my trellis is on an older version with less configs.

### DIFF
--- a/roles/wordpress-setup/templates/https.conf.j2
+++ b/roles/wordpress-setup/templates/https.conf.j2
@@ -1,0 +1,22 @@
+include h5bp/directive-only/ssl.conf;
+include h5bp/directive-only/ssl-stapling.conf;
+
+ssl_dhparam /etc/nginx/ssl/dhparams.pem;
+ssl_buffer_size 1400; # 1400 bytes to fit in one MTU
+
+{% set hsts_max_age = item.value.ssl.hsts_max_age | default(nginx_hsts_max_age) %}
+{% set hsts_include_subdomains = item.value.ssl.hsts_include_subdomains | default(nginx_hsts_include_subdomains) | ternary('includeSubdomains', None) %}
+{% set hsts_preload = item.value.ssl.hsts_preload | default(nginx_hsts_preload) | ternary('preload', None) %}
+add_header Strict-Transport-Security "max-age={{ [hsts_max_age, hsts_include_subdomains, hsts_preload] | reject('none') | join('; ') }}";
+
+{% if item.value.ssl.provider | default('manual') == 'manual' and item.value.ssl.cert is defined and item.value.ssl.key is defined -%}
+  ssl_certificate         {{ nginx_path }}/ssl/{{ item.value.ssl.cert | basename }};
+  ssl_certificate_key     {{ nginx_path }}/ssl/{{ item.value.ssl.key | basename }};
+{%- elif item.value.ssl.provider | default('manual') == 'letsencrypt' -%}
+  ssl_certificate         {{ nginx_path }}/ssl/letsencrypt/{{ item.key }}-bundled.cert;
+  ssl_certificate_key     {{ nginx_path }}/ssl/letsencrypt/{{ item.key }}.key;
+{%- elif item.value.ssl.provider | default('manual') == 'self-signed' -%}
+  ssl_certificate         {{ nginx_path }}/ssl/{{ item.key }}.cert;
+  ssl_trusted_certificate {{ nginx_path }}/ssl/{{ item.key }}.cert;
+  ssl_certificate_key     {{ nginx_path }}/ssl/{{ item.key }}.key;
+{%- endif -%}
--- a/roles/wordpress-setup/templates/wordpress-site.conf.j2
+++ b/roles/wordpress-setup/templates/wordpress-site.conf.j2
@@ -33,28 +33,7 @@ server {
   add_header Fastcgi-Cache $upstream_cache_status;
 
   {% if item.value.ssl is defined and item.value.ssl.enabled | default(false) -%}
-    include h5bp/directive-only/ssl.conf;
-    include h5bp/directive-only/ssl-stapling.conf;
-
-    ssl_dhparam /etc/nginx/ssl/dhparams.pem;
-    ssl_buffer_size 1400; # 1400 bytes to fit in one MTU
-
-    {% set hsts_max_age = item.value.ssl.hsts_max_age | default(nginx_hsts_max_age) %}
-    {% set hsts_include_subdomains = item.value.ssl.hsts_include_subdomains | default(nginx_hsts_include_subdomains) | ternary('includeSubdomains', None) %}
-    {% set hsts_preload = item.value.ssl.hsts_preload | default(nginx_hsts_preload) | ternary('preload', None) %}
-    add_header Strict-Transport-Security "max-age={{ [hsts_max_age, hsts_include_subdomains, hsts_preload] | reject('none') | join('; ') }}";
-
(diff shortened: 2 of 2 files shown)


swalkinshaw:
Thanks, this looks correct. I'm amazed we missed this and that it wasn't completely broken. I don't know even how an Nginx `server` block with `listen 443 ssl http2` would even work without any SSL details (like certificate).

swalkinshaw:
My only concern is duplicated configuration now. And there's still a few things like HSTS and tweaks missing from this server block that exist in the main one. It might be better to generate a separate `https` config and include it in both places. It would contain this:

```jinja
{% if item.value.ssl is defined and item.value.ssl.enabled | default(false) -%}
    include h5bp/directive-only/ssl.conf;
    include h5bp/directive-only/ssl-stapling.conf;

    ssl_dhparam /etc/nginx/ssl/dhparams.pem;
    ssl_buffer_size 1400; # 1400 bytes to fit in one MTU

    {% set hsts_max_age = item.value.ssl.hsts_max_age | default(nginx_hsts_max_age) %}
    {% set hsts_include_subdomains = item.value.ssl.hsts_include_subdomains | default(nginx_hsts_include_subdomains) | ternary('includeSubdomains', None) %}
    {% set hsts_preload = item.value.ssl.hsts_preload | default(nginx_hsts_preload) | ternary('preload', None) %}

    {% if item.value.ssl.provider | default('manual') == 'manual' and item.value.ssl.cert is defined and item.value.ssl.key is defined -%}
      ssl_certificate         {{ nginx_path }}/ssl/{{ item.value.ssl.cert | basename }};
      ssl_certificate_key     {{ nginx_path }}/ssl/{{ item.value.ssl.key | basename }};
    {%- elif item.value.ssl.provider | default('manual') == 'letsencrypt' -%}
      ssl_certificate         {{ nginx_path }}/ssl/letsencrypt/{{ item.key }}-bundled.cert;
      ssl_certificate_key     {{ nginx_path }}/ssl/letsencrypt/{{ item.key }}.key;
    {%- elif item.value.ssl.provider | default('manual') == 'self-signed' -%}
      ssl_certificate         {{ nginx_path }}/ssl/{{ item.key }}.cert;
      ssl_trusted_certificate {{ nginx_path }}/ssl/{{ item.key }}.cert;
      ssl_certificate_key     {{ nginx_path }}/ssl/{{ item.key }}.key;
    {%- endif -%}
  {% endif %}
```

Then you'd just include it like:

```jinja
{% for host in item.value.site_hosts if item.value.www_redirect | default(true) %}
server {
  {% if item.value.ssl is defined and item.value.ssl.enabled | default(false) -%}
    listen 443 ssl http2;
    include https.conf;
  {% else -%}
    listen 80;
  {% endif -%}

  server_name {{ host | reverse_www(append=false) }};
  return 301 $scheme://{{ host }}$request_uri;
}
{% endfor %}

**Comment by swalkinshaw** [comment]
@swaincreates would you want to take a crack at this? It would just involve creating a new template like here: https://github.com/roots/trellis/blob/fc44e94d6f174792b00833d79a89dd92d4b1daf8/roles/wordpress-setup/tasks/nginx.yml#L50-L55

**Comment by swaincreates** [comment]
Sure, will give it a shot.

**Comment by swaincreates** [comment]
So, after taking a look at this can you maybe help me out? If we want a separate https.conf file it will have to be unique to each site right? Because they'll have unique cert names and paths. 

So i'll have to create a site1.com-https.conf and a site2.com-https.conf and then include site1.com-https.conf if its ssl? And in that case its kind of not duplication.. more like a partial that we're including. In that case, do we still go this path? OR, do we create an https.conf file of only things that don't have unique info:

```
    include h5bp/directive-only/ssl.conf;
    include h5bp/directive-only/ssl-stapling.conf;

    ssl_dhparam /etc/nginx/ssl/dhparams.pem;
    ssl_buffer_size 1400; # 1400 bytes to fit in one MTU

    {% set hsts_max_age = item.value.ssl.hsts_max_age | default(nginx_hsts_max_age) %}
    {% set hsts_include_subdomains = item.value.ssl.hsts_include_subdomains | default(nginx_hsts_include_subdomains) | ternary('includeSubdomains', None) %}
    {% set hsts_preload = item.value.ssl.hsts_preload | default(nginx_hsts_preload) | ternary('preload', None) %}
```

and then just duplicate ssl cert paths/names? Thoughts?

**Comment by swalkinshaw** [comment]
Oh yeah, good point. Okay let's just stick with your original code then for now at least. Could you just ensure it's the same as the main config? Think it's just this missing:

```
  ssl_dhparam /etc/nginx/ssl/dhparams.pem;
  ssl_buffer_size 1400; # 1400 bytes to fit in one MTU

  {% set hsts_max_age = item.value.ssl.hsts_max_age | default(nginx_hsts_max_age) %}
  {% set hsts_include_subdomains = item.value.ssl.hsts_include_subdomains | default(nginx_hsts_include_subdomains) | ternary('includeSubdomains', None) %}
  {% set hsts_preload = item.value.ssl.hsts_preload | default(nginx_hsts_preload) | ternary('preload', None) %}
```

**Comment by swaincreates** [comment]
Yup, coming right up

**Comment by swalkinshaw** [comment]
Thanks @swaincreates. Would you be able to squash these commits into 1?

**Comment by swaincreates** [comment]
Sure will figure it out

**Comment by fullyint** [comment]
@swaincreates I recently learned about Ansible's [template lookups](http://docs.ansible.com/ansible/playbooks_lookups.html#more-lookups). What do you think about...
* moving this https config to a new template `roles/wordpress-setup/templates/https.conf.j2`
* replacing the 1) existing https config and 2) your proposed new https config with the following:

`{{ lookup('template', 'https.conf.j2') }}`

**Comment by swaincreates** [comment]
Yea I was thinking about that as well when I was researching options... Guess thats just a matter of project style. I personally would rather not have to go look up the template, but other people aren't fond of duplication and lastly.. I'm not maintaining this thing :) You guys let me know if thats the direction you want to go and I'll recommit.

**Comment by swalkinshaw** [comment]
:+1: to @fullyint's solution. Although the duplication isn't bad here I could easily see us updating one of these code blocks in the future and forgetting about the other one.

Thanks for being patient @swaincreates :)

**Comment by swaincreates** [comment]
Tried to add line breaks appropriately based on current format - let me know if you want some formatting changes though.

**Comment by swalkinshaw** [comment]
Tested and working, thanks!

**Comment by swaincreates** [comment]
Its working for you? I just tested it on something and I had a path issue. Was writing this post as you commented/merged.

**Comment by swalkinshaw** [comment]
What was your issue? I tried a new server and tested main site + redirects and some other pages just now.

**Comment by swaincreates** [comment]
Sweet! I am on ansible 1.9.3 when I tested so that could be part of problem. For me when I tried it I got this and you can see how the path is off:

```
TASK: [wordpress-setup | Create WordPress configuration for Nginx] ************
fatal: [default] => {'msg': 'AnsibleError: unable to read /Users/swain/ASPPH_Trellis/ansible/https.conf.j2', 'failed': True}
fatal: [default] => {'msg': 'One or more items failed.', 'failed': True, 'changed': False, 'results': [{'msg': 'AnsibleError: unable to read /Users
/swain/ASPPH_Trellis/ansible/https.conf.j2', 'failed': True}]}
```

swaincreates:
Yup that appears to be the problem (being on lower version of Ansible) according to this issue https://github.com/ansible/ansible/issues/9551

swaincreates:
Glad its working - thanks for helping me along. Love the project!

PR #538, opened 2016-03-30
Move modules and plugins to lib/trellis directory

fullyint:
**Structure via new `lib/trellis` directory**
```
lib/
  trellis/
    modules/
      deploy_helper.py
    plugins/
      callback/         <-- coming soon
      filter/
        filters.py
```
* structure mirrors Ansible's `lib/ansible` directory
* tidies up: pulls python files into `lib/trellis` (modules, filter plugin, callback plugins)


swalkinshaw:
:+1:

PR #542, opened 2016-04-01
Fix #489 - Add $realpath_root to fastcgi_cache_key

### LINKED ISSUE #489: Missing assets after deploy with FastCGI cache
currently there's a possibility of a user running into missing theme assets after a deploy if fastcgi caching is enabled and their site uses filename based asset revving 

background info: https://discourse.roots.io/t/missing-sage-assets-after-trellis-deploy-with-fastcgi-caching-turned-on/5870/2

swalkinshaw:
This acts as an automatic cache buster when a deploy happens.
`$realpath_root` contains the real/un-symlinked path to the web root.

Example: `/srv/www/example.com/releases/20160101` instead of
`/srv/www/example.com/current`

This will prevent some known bugs where a cached page refers to a now
non-existent "revved" asset like Sage generates (a CSS or JS file with a
hash in the filename).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Fix #489 - Add $realpath_root to fastcgi_cache_key ([#542](https://github.com/roots/trellis/pull/542))
 * Move modules and plugins to `lib/trellis` directory ([#538](https://github.com/roots/trellis/pull/538))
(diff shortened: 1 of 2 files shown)


swalkinshaw:
In #489 @retlehs had said that adding `$realpath_root` didn't work. However in my tests, adding it to the cache key correctly resulted in a cache miss on the next request.

PR #544, opened 2016-04-02
Add plugin to pretty print Ansible msg output

fullyint:
This PR proposes a new ['stdout_callback'](http://docs.ansible.com/ansible/intro_configuration.html#stdout-callback) named `output.py` to augment Ansible's [`default.py`](https://github.com/ansible/ansible/blob/bb6cadefa2d68ed2a668fba14dc027947e043ae5/lib/ansible/plugins/callback/default.py). It extracts and displays the `msg` attribute, interpreting newline characters and applying textwrap.

This PR adds a couple .py files, still mirroring Ansible project's structure.
```
lib/
  trellis/
    modules/
      deploy_helper.py
    plugins/
      callback/         <-- new
        output.py       <-- new
      filter/
        filters.py
    utils/              <-- new
      output.py         <-- new
```

Run the playbook below on Trellis master then again on this PR's branch to see the different formatting.
```
- name: Output demo 
  gather_facts: false
  hosts: localhost
  vars_files:
    - group_vars/staging/wordpress_sites.yml
  tasks:
    - debug:
        msg: If `with_dict`, `item` below displays dict key only (unless `-v`)
      register: results

    - debug:
        msg: If `with_items` based on dict, attempts to display only key (unless `-v`)
      with_items: "{{ results.results }}"

    - assert:
        that: "{{ item.name == 'bar' }}"
        msg: |
          {{ item.msg }}

          {{ item.msg }}

          {{ item.msg*2 }}
      with_items:
        - {'name': 'foo', 'msg': 'Fail for item zero.  '}
        - {'name': 'bar', 'msg': 'Example ok/skipped item'}
        - {'name': 'foo', 'msg': 'The horizontal rule between hosts/items helps distinguish between multiple hosts/items, especially when messages have blank lines. Blank lines would no longer clearly mark the start/end of the message per host/item.  '}
```
This is an improved piece of #531.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Add plugin to pretty print Ansible msg output ([#544](https://github.com/roots/trellis/pull/544))
 * Fix #482 - Multisite is-installed deploy check ([#543](https://github.com/roots/trellis/pull/543))
 * Skip setting permalink for multisite installs ([#546](https://github.com/roots/trellis/pull/546))
 * Fix #489 - Add $realpath_root to fastcgi_cache_key ([#542](https://github.com/roots/trellis/pull/542))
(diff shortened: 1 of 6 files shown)


swalkinshaw:
So much better :+1:

PR #545, opened 2016-04-02
Handle skipped items in "Notify of challenge failures"

fullyint:
**Problem:**
```
TASK [letsencrypt : Notify of challenge failures] ********
ERROR! 'dict object' has no attribute 'failed_hosts'
```
**Steps to reproduce:** Create two sites in `wordpress_sites`, one with SSL enabled with `letsencrypt` and the other with SSL not enabled. Run `server.yml`. Admittedly a bit of an edge case.

**Why this happens:** 
* The `Test Acme Challenges` task skips the site that doesn't have SSL enabled. The task produces the registered var `letsencrypt_test_challenges`. 
* The `Notify of challenge failures` task loops through `letsencrypt_test_challenges.results`, evaluating the `failed_hosts` attribute for each item. The problem is that skipped items don't have a `failed_hosts` attribute.

**Resolution:** This PR adds a check `when: not item | skipped` so the task no longer fails.


swalkinshaw:
:100:

PR #549, opened 2016-04-04
Display system info with fail output

fullyint:
To help with debugging and support requests, this PR augments fail output to include version info for Ansible, Vagrant, and Trellis. Also displays the host machine's platform.
```
TASK [common : Validate Ansible version] ***************************************
System info:
  Ansible 1.9.4; Vagrant 1.8.1; Linux
  Trellis at "Add plugin to pretty print Ansible msg output"
Your Ansible version is too old. Trellis requires at least 2.x. Your version is 1.9.4
fatal: [host 1]: FAILED! => {"assertion": false, "changed": false, "evaluated_to": false, "failed": true}
Your Ansible version is too old. Trellis requires at least 2.x. Your version is 1.9.4
fatal: [host 2]: FAILED! => {"assertion": false, "changed": false, "evaluated_to": false, "failed": true}
```

I would have preferred to have the system info in a footer, but displaying it at the top allows simpler code. Having the system info at the top may be better, however, possibly making it more likely that users will include it in the output they paste to the help forum. I can provide details if desired.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -73,6 +73,7 @@ Vagrant.configure('2') do |config|
   if Vagrant::Util::Platform.windows?
     config.vm.provision :shell do |sh|
       sh.path = File.join(ANSIBLE_PATH, 'windows.sh')
+      sh.args = [Vagrant::VERSION]
     end
   else
     config.vm.provision :ansible do |ansible|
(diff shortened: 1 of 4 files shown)


swalkinshaw:
Working for me :+1:

PR #556, opened 2016-04-08
Tidy up output.py by moving some methods to utils

fullyint:
This PR changes nothing but organization, moving a few methods from `plugins/callback/output.py` to `utils/output.py`. Now the callback plugin producing the Trellis custom output is particularly tidy and straightforward. This keeps its code manageable and conceptually clear as we add more output features.

Note. This reorganization puts faith in [this comment](http://stackoverflow.com/a/16084773): 
>Instances of `self` get passed to functions (and methods of other classes) all the time.

If someone believes that is bad python practice, let me know.


swalkinshaw:
Good improvement :+1:

PR #558, opened 2016-04-08
Update version check for Ansible 2.x

fullyint:
* Bumps `minimum_ansible_version` var and version check from `1.9.2` to `2.0.0.2`.
* Adds [`run_once`](http://docs.ansible.com/ansible/playbooks_delegation.html#run-once) to version check so error prints only once, even if multiple hosts.
* Edits `vars.py` to check and warn if Ansible 1.x. Otherwise the plugin executes and produces a traceback before the above version check executes. Trellis Ansible plugins rely on the structure of the Ansible 2.x refactor.


swalkinshaw:
:100:

PR #559, opened 2016-04-09
Add custom output options to ansible.cfg

fullyint:
See how output differs while toggling the new `[trellis_output]` settings in `ansible.cfg` and running `server.yml` (or just add/omit `-v`). A couple roles whose output is quite affected: `--tags users,wordpress`

Settings are explained in the docs: roots/docs#32

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -1,6 +1,7 @@
 [defaults]
 callback_plugins = ~/.ansible/plugins/callback_plugins/:/usr/share/ansible_plugins/callback_plugins:lib/trellis/plugins/callback
 stdout_callback = output
+display_skipped_hosts = False
 filter_plugins = ~/.ansible/plugins/filter_plugins/:/usr/share/ansible_plugins/filter_plugins:lib/trellis/plugins/filter
 force_handlers = True
 inventory = hosts
@@ -10,3 +11,9 @@ vars_plugins = ~/.ansible/plugins/vars_plugins/:/usr/share/ansible_plugins/vars_
 
 [ssh_connection]
 ssh_args = -o ForwardAgent=yes -o ControlMaster=auto -o ControlPersist=60s
+
+[trellis_output]
+display_include_tasks = False
+display_skipped_items = False
+truncate_items = True
+wrap_width = 80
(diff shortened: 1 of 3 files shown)


fullyint:
Closing as "a feature that was explored but discontinued." At present, there is no consensus that the "nice-to-have" benefits to output format outweigh risks: 
- Different output could confuse users ("Why does this output look different from usual Ansible?").
- Relying on a [possibly unstable Ansible plugin API](https://groups.google.com/forum/#!msg/ansible-devel/Vuhg_rElDjk/Qk80re2uEAAJ) may mean accepting that Ansible updates could "break" Trellis for features that are not "must-have." 

The original intention was to follow this PR with another PR enabling templating and colors so that Ansible output might feel accessible to users. I loaded a demo branch [`output-options`](https://github.com/roots/trellis/tree/output-options) into the `roots/trellis` repo so that anyone interested can review the potential features/implementation. The corresponding docs are in a [`trellis-output-options`](https://github.com/roots/docs/tree/trellis-output-options) branch. Roots can discard these branches in a few weeks.

Here is a screenshot of output demonstrating the templating and colors.

image

Here is the playbook to create the output above.
```
- name: Colors
  gather_facts: false
  hosts: localhost
  tasks:
    - name: Show available Ansible colors and current color settings
      debug:
        msg: |
          AVAILABLE COLORS
          --------------------
          {{ ansible_colors }}

          CURRENT COLOR SETTINGS
          --------------------
          {{ color_settings }}

    - name: Demonstrate a few colors in context
      debug:
        msg: |
          Here is the default color for all tasks <error>except failed tasks which use this color_error.</error> Also, here is <warn>the warning color,</warn> and the <ok>ok color</ok>.

          When a color setting is 'None' the string won't be colorized. For example, if `color_ok = None` in `ansible.cfg`, then the string won't be colorized (but the color tag will still be stripped).

          When there is `code in backticks`, the backticks will be stripped and the text colorized. However, if `color_code = None` (in `ansible.cfg`) then the backticks will just display and no change in color will be applied.

          URLs look nice in a slightly different shade, e.g., <bold>https://roots.io/trellis/docs/ssl/</bold>

    - name: Demonstrate the failure output
      fail:
        msg: |
          Please specify SSL `provider`:
          ```
          wordpress_sites:
            example.com:
              ssl:
                enabled: true
                provider: <ok>letsencrypt</ok>

          `Backticks` always display inside fenced code blocks and are not colorized. Codeblocks do not wrap, which is good for displaying commands users may copy/paste.
          ```
```

PR #560, opened 2016-04-11
Tidy up sudoer password definitions and comments

fullyint:
Removes sudoer password definitions from `group_vars/development` because `dev.yml` doesn't run the `users` role.

Edits comments to more directly instruct users to edit `vault_sudoer_passwords` in `vault.yml`, instead of pointing them to `group_vars/<environment>/main.yml` which in turn points them to `vault.yml`.


swalkinshaw:
development gets even simpler :dancers:

PR #562, opened 2016-04-12
Consolidate validations into new validations role

fullyint:
This consolidates validations -- those that can be moved -- into a new `validations` role. It adds a few new validations too. Forthcoming validations can be conveniently added to this `validations` role. 

The basic concept is that the validations task loops through the `validations` listed in `group_vars/all/validations.yml`. Each validation item is defined in `roles/validations/defaults` and each item can draw its `msg` from a template (`roles/validations/templates`) or just a string passed to the `msg`. Users can disable specific validations by commenting out items in `group_vars/all/validations.yml`, rewrite (override) validation items in `group_vars`, or write new validation items, etc.

This bumps the requirement to Ansible 2.0.1 (Feb 24, 2016) which enables system info to display in a failed task's footer using the `v2_playbook_on_no_hosts_remaining` callback. This event hook has been around all of 2.x, but seems not to fire on 2.0.0.2. I think [this commit](https://github.com/ansible/ansible/commit/8ee8872d6b7d34cbee4704d07743632cb616fb66#diff-823724d89e3b69d4236db3652b61ffcaR166) leading into 2.0.1 may be what fixed it so it _does_ fire on 2.0.1. (I didn't test to confirm that that was the exact commit.)

Ready to fail? Here are some shortcuts to trigger validation failures.

```
# validate_ansible_version
ansible-playbook server.yml -e env=production -e '{"ansible_requirements":[{"version":"5.0.1.0","operator":">"}]}'

# validate_sudo_user
ansible-playbook server.yml -e 'env=production sshd_permit_root_login=false' -e '{"users":[{"name":"nope", "groups":["nope"]}]}'

# validate_admin_user_is_not_root
ansible-playbook server.yml -e "env=production sshd_permit_root_login=false admin_user=root"

# validate_web_user_is_not_root
ansible-playbook server.yml -e "env=production web_user=root"

# validate_site_variable
./deploy.sh production nope.com

# validate_repo_definition
ansible-playbook deploy.yml -e 'env=production site=example.com' -e '{"wordpress_sites":{"example.com":{"repo":"malformed"}}}'

# validate_repo_customized
ansible-playbook deploy.yml -e "env=production site=example.com" -e '{"wordpress_sites":{"example.com":{"repo":"git@github.com:example/example.com.git"}}}'

# validate_one_env_group_only
# you'll have to manually edit `hosts` -- I don't know a way to assign groups via the cli
```

The default setting of `validate_hosts_separately: false` (in `group_vars/all/validations.yml`) enables the `run_once` task parameter because configs don't vary across hosts by default. The one validation that would ideally run for each host separately would be `validate_one_env_group_only`. It still works if there is just one host machine/server (probably 95% of users) or if the multi-group host is listed first in the inventory, but the validation won't be effective for hosts that do not meet these criteria (unless the user changes to `validate_hosts_separately: true`, of course).

---

**Edit.** Users can also disable a particular validation via the cli using extra-vars, e.g., 

```
ansible-playbook server.yml -e env=production -e '{"ansible_requirements":[{"version":"5.0.1.0","operator":">"}]}' -e validate_ansible_version={} --tags validations
```

The above `-e '{"ansible_requirements":[{"version":"5.0.1.0","operator":">"}]}'` would normally trigger the `validate_ansible_version` but it doesn't because of the added `validate_ansible_version={}`

Users can also one-off disable all validations via the cli using `validations=false` or `--skip-tags`:

```
ansible-playbook server.yml -e "env=production validations=false"

ansible-playbook server.yml -e env=production --skip-tags validations
```


roles/validations/defaults/main.yml
```diff
@@ -0,0 +1,75 @@
+# A validation will fail if any condition listed is true
```
mvasin: It will fail if ansible is >=2.0.2.0? That seems wrong for now.

PR #564, opened 2016-04-13
Pass vault cli args to remote-user ping task

fullyint:
**What works.** When files are encrypted and the `remote-user` task's ad hoc command pings the host, it succeeds in using the `vault_password_file = .vault_pass` specified in `ansible.cfg`. 

**What fails.** If a user chooses against this setting in `ansible.cfg`, opting instead to specify the vault password via the cli with an option such as `--ask-vault-pass` or `--vault-password-file`, the `remote-user` task's ping command will lack these options. In this case, the task yields an `ERROR! Decryption failed` and leads the `admin_user` to be selected instead of `root`.

**Solution.** The `vars.py` plugin can detect vault-related options passed to the cli and load them up in a new `cli_args_vault` variable appended to the `ping` command.

### DIFF
--- a/roles/remote-user/tasks/main.yml
+++ b/roles/remote-user/tasks/main.yml
@@ -1,6 +1,6 @@
 ---
 - name: Determine whether to connect as root or admin_user
-  local_action: command ansible {{ inventory_hostname }} -m ping{{ (inventory_file == None) | ternary('', ' -i ' + inventory_file | string) }} -u root
+  local_action: command ansible {{ inventory_hostname }} -m ping{{ (inventory_file == None) | ternary('', ' -i ' + inventory_file | string) }} -u root {{ cli_args_vault | default('') }}
   failed_when: false
   changed_when: false
   register: root_status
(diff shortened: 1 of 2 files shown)


roles/remote-user/tasks/main.yml
```diff
@@ -1,6 +1,6 @@
 ---
 - name: Determine whether to connect as root or admin_user
-  local_action: command ansible {{ inventory_hostname }} -m ping{{ (inventory_file == None) | ternary('', ' -i ' + inventory_file | string) }} -u root
+  local_action: command ansible {{ inventory_hostname }} -m ping{{ (inventory_file == None) | ternary('', ' -i ' + inventory_file | string) }} -u root {{ cli_args_vault | default('') }}
```
swalkinshaw: Is this `default` needed? Seems it should be taken care of from within the `else` branach of `cli_args_vault`

fullyint: I added `default` as a safety in case [`vars_plugins`](https://github.com/roots/trellis/blob/0667e4293d86b6920db838064433293be590a7d6/ansible.cfg#L9) ever happened to be disabled, in which case the var would be undefined.

swalkinshaw:
:+1:

PR #565, opened 2016-04-14
Enable Let's Encrypt to transition http sites to https

fullyint:
**The problem.** When the LE role runs for an existing http site, the site already has an Nginx conf in `sites-enabled/site.com.conf` with a server block that doesn't include a location block for the Acme Challenge. If the LE role [happens to create/load](https://github.com/roots/trellis/blob/0667e4293d86b6920db838064433293be590a7d6/roles/letsencrypt/tasks/nginx.yml#L8-L21) the [`nginx-challenge-site.conf`](https://github.com/roots/trellis/blob/0667e4293d86b6920db838064433293be590a7d6/roles/letsencrypt/templates/nginx-challenge-site.conf.j2), its server block will compete in sorting of the [glob inclusion](https://github.com/roots/trellis/blob/0667e4293d86b6920db838064433293be590a7d6/roles/nginx/templates/nginx.conf.j2#L136) and either the Acme Challenge or the actual site won't be served till the `nginx-challenge-site.conf` [is disabled](https://github.com/roots/trellis/blob/0667e4293d86b6920db838064433293be590a7d6/roles/letsencrypt/tasks/certificates.yml#L41). This means that either the [Acme Challenges](https://github.com/roots/trellis/blob/0667e4293d86b6920db838064433293be590a7d6/roles/letsencrypt/tasks/nginx.yml#L35) will fail, or the regular site will be down for a moment.

**Proposed solution.** This PR resolves the problem of competing server blocks by extending a strategy already in place. The already existing strategy is that when ssl is enabled, the [http->https redirection](https://github.com/roots/trellis/blob/0667e4293d86b6920db838064433293be590a7d6/roles/wordpress-setup/templates/wordpress-site.conf.j2#L83) section also loads the Acme Challenge location block so that LE cert renewals will pass challenge tests. 

It seems relatively harmless to have the Acme Challenge location block there in the conf all the time. This PR adds the Acme Challenge location block to the non-ssl conf. As a result, when the LE role runs on an existing http site, it will use the site's existing conf (or create a new Acme Challenge conf if somehow there is no conf already). Either way, the Acme Challenge tests pass. 

This creates a bit more crossover between the `letsencrypt` role and the `wordpress-setup` role, but I'm not sure it can be avoided.

-----

The one context privileged to not have to deal with the extra Acme Challenges location block is what should be the most common: 1) ssl enabled and 2) no www redirect necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+### HEAD
+* Enable Let's Encrypt to transition http sites to https ([#565](https://github.com/roots/trellis/pull/565))
+
 ### 0.9.7: April 10th, 2016
 * Fix #550 - Properly skip permalink setup for MU ([#551](https://github.com/roots/trellis/pull/551))
 * Escape salts and keys to avoid templating errors ([#548](https://github.com/roots/trellis/pull/548))
--- a/roles/common/handlers/main.yml
+++ b/roles/common/handlers/main.yml
@@ -10,6 +10,4 @@
     state: reloaded
 
 - name: reload nginx
-  service:
-    name: nginx
-    state: reloaded
+  include: reload_nginx.yml
--- a/roles/common/tasks/reload_nginx.yml
+++ b/roles/common/tasks/reload_nginx.yml
@@ -1,0 +1,12 @@
+---
+- name: test nginx conf
+  command: nginx -t
+  register: nginx_test
+  changed_when: false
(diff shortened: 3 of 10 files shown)


swalkinshaw:
Working great :+1:

swalkinshaw:
Add changelog then merge :)

PR #566, opened 2016-04-14
Adding dbus to essentials

jameskraus:
When the timezone is being set, hostnamectl is called, but it needs dbus to function.

This is one of the first things to error when you try to use trellis with Debian servers, so adding to the essentials should slightly lower the bar for Debian users. It comes installed on Ubuntu as of 14.04.

Tried this on Debian 8.3. Fixed the issue.


swalkinshaw:
Sure, why not :+1:

PR #568, opened 2016-04-16
Define development default for Let's Encrypt variable

fullyint:
Fixes https://discourse.roots.io/t/cant-start-a-new-virtual-machine-after-pull-request-565-was-merged/6475

```
TASK [wordpress-setup : Create Nginx conf for challenges location] *************
AnsibleUndefinedVariable: 'acme_tiny_challenges_directory' is undefined
fatal: [default]: FAILED! => {"changed": false, "failed": true}
```

The `acme_tiny_challenges_directory` default is not defined for the Vagrant VM because `dev.yml` omits the LE role and its default. Adding a default specific to `env=development` seems to be the cleanest solution.

### DIFF
--- a/group_vars/development/main.yml
+++ b/group_vars/development/main.yml
@@ -1,3 +1,4 @@
+acme_tiny_challenges_directory: "{{ www_root }}/letsencrypt"
 env: development
 ferm_enabled: false
 mysql_root_password: "{{ vault_mysql_root_password }}" # Define this variable in group_vars/development/vault.yml


group_vars/development/main.yml
```diff
@@ -2,3 +2,4 @@ env: development
 ferm_enabled: false
 mysql_root_password: "{{ vault_mysql_root_password }}" # Define this variable in group_vars/development/vault.yml
 web_user: vagrant
+acme_tiny_challenges_directory: "{{ www_root }}/letsencrypt"
```
swalkinshaw: :abc: order :angel:

swalkinshaw:
:+1:

PR #573, opened 2016-04-21
Fix #563 - Improve remote databases

### LINKED ISSUE #563: Provisioning Remote Server w/ Remote SQL DB ‘true’ Issues
- [X] This is a bug report

## Bug report

**Please provide steps to reproduce, including full log output:**

Provisioning a remote server with newest master branch of Trellis with the variable for remote sql set to true.
mysql_remote_database: true
I have tried it with it set to true from first provision (so no mariadb ever installed) and also provisioning first so mariadb gets installed then switching afterwards.

Either way Trellis will always fail when trying to connect to the remote db and will act like it can't find a local database even though db_host is set to an IP address.

Fails with the error "Access denied for root@localhost(using password YES)"

I also tried SSHing into the server and running 'wp db' commands after verifying that the .env file 100% had the REMOTE DB settings loaded in it. Every time it still wants to connect root@localhost

After trying a few things I decided to ssh into the server and install the package
mysql-client, which I don't think ended up helping.
(This is with no mariadb role run ever)

Then I edited /etc/mysql/my.cnf file and changed the variable 'bind-address' from the local ip 127.0.0.1 to my remote IP, then everything worked as expected.

So it seems the Ansible roles are running through the mysql socket which is set to localhost exclusively. Once I changed the bind address like I said I am able to run commands on the remote DB all day.

**Please describe your local environment:**

Ansible version: 2.0.0.2 and also tried 2.0.1.0

OS: Running the Trellis playbook from an Ubuntu 14.04 instance provisioned on Google Cloud Platform

DB Server: Google Cloud SQL

Note: I have successfully completed this exact provisioning numerous times using the same env variables.

**Where did the bug happen? Development or remote servers?**

Remote server, Ubuntu 14.04 Google Cloud Compute Instance

**Please provide a repository or your `wordpress_sites` config (if possible):**

```
  example.com:
    site_hosts:
      - staging.example.com
    local_path: ../example.com
    repo: git@bitbucket.org:example-sites/example.com.git
    branch: master
    db_create: false
    multisite:
      enabled: false
    ssl:
      enabled: true
      provider: letsencrypt
    cache:
      enabled: false
    env:
      disable_wp_cron: true
      db_host: 104.xxx.xxx.xxx
      db_user: example_staging
      db_name: example_staging
      db_prefix: ex_


  example.com:
    env:
      db_password: 'epic_db_password'
      # Generate your keys here: https://api.wordpress.org/secret-key/1.1/salt/
      pretend my salts are here

```

**Is there a related [Discourse](https://discourse.roots.io/) thread or were any utilized (please link them)?**

https://discourse.roots.io/t/provisioning-remote-server-w-remote-sql-db-true-issues/

Like I said as soon as I changed the socket IP it started working perfectly

swalkinshaw:
Previously you had to manually toggle the remote database option via
`mysql_remote_database` (which also didn't fully work).

Now Trellis automatically detects if the sites are using a remote
`db_host` or not (aka not `localhost`) and handles things accordingly.

This also installs the mysql-client now regardless.

### DIFF
--- a/roles/wordpress-setup/defaults/main.yml
+++ b/roles/wordpress-setup/defaults/main.yml
@@ -1,0 +1,1 @@
+site_uses_local_db: "{{ site_env.db_host == 'localhost' }}"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Fix #563 - Improve remote databases ([#573](https://github.com/roots/trellis/pull/573))
 * Fix #569 - Only skip subdomains for non-www domains ([#570](https://github.com/roots/trellis/pull/570))
 * Enable Let's Encrypt to transition http sites to https ([#565](https://github.com/roots/trellis/pull/565))
 
--- a/dev.yml
+++ b/dev.yml
@@ -10,7 +10,7 @@
     - { role: ferm, tags: [ferm] }
     - { role: ntp }
     - { role: sshd, tags: [sshd] }
-    - { role: mariadb, tags: [mariadb], when: not mysql_remote_database }
+    - { role: mariadb, tags: [mariadb] }
     - { role: ssmtp, tags: [ssmtp mail] }
     - { role: mailhog, tags: [mailhog mail] }
     - { role: php, tags: [php] }
--- a/group_vars/all/database.yml
+++ b/group_vars/all/database.yml
@@ -1,3 +1,0 @@
-mariadb_dist: trusty
-mysql_remote_database: false
-mysql_root_user: root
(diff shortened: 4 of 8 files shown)


ofauno:
@swalkinshaw that skips the tasks also if __127.0.0.1__ is defined, that was confusing and time-consuming to see 🐼

roles/wordpress-setup/defaults/main.yml
```diff
@@ -0,0 +1 @@
+site_uses_local_db: "{{ site_env.db_host == 'localhost' }}"
```
ofauno: guess we can put "127.0.0.1" also? 😄

swalkinshaw: We could. I kind of thought about that but figured there's no reason for anyone to put `127.0.0.1`? Trellis configures localhost automatically.

Want to do a PR?

PR #579, opened 2016-05-05
Require Ansible 2.0.2 and remove deploy_helper

swalkinshaw:
The `deploy_helper` module was vendored because we needed a fix which was not included in a stable Ansible yet. 2.0.2.0 includes the fix so we can bump our version requirements and un-vendor the module.


retlehs:
+ https://github.com/roots/docs/blob/master/trellis/installing-trellis.md

PR #578, opened 2016-05-04
Add connection-related cli options to ping command

fullyint:
The `remote-user` role's `ping` command tests the connection for the `root` user. Some `ping` history:
- #470 enabled `--inventory` option passed to `ansible-playbook` command to carry over to `ping`.
- #564 enabled vault cli args passed to `ansible-playbook` command to carry over to `ping`.
- #574 raised question of carrying `--ask-pass` over to `ping`, if present.

This PR implements a unified approach to collecting the connection-related cli options passed to the `ansible-playbook` command, carrying the options over to the `ping` command. This includes the options in the bullets above, and a few additional connection-related options (selected from the list displayed if you run `ansible-playbook` with no args or options).

PR #582, opened 2016-05-06
Pass cli options to user role's admin_user ping test

fullyint:
When #578 added connection-related CLI options to the `remote-user` role's `ping` command, it probably should have also updated the `users` role's `ping` command (introduced in #345). Here is a [discourse thread](https://discourse.roots.io/t/disabling-root-login-while-using-ask-vault-pass-error-decryption-failed/6655) demonstrating the need for this fix.

This PR:
- applies #578's `cli_options_ping` to the `users` role
- removes `--user` from `cli_options_ping`, adjusting each task to specify user explicitly, given that the two different ping tests rely on different users
- adds a default for the `cli_options_ping` variable as a safety in case [`vars_plugins`](https://github.com/roots/trellis/blob/472ce5c91c0ee5da3d295b1d830281500b7be590/ansible.cfg#L11) ever happened to be disabled, in which case the var would be undefined

### DIFF
--- a/lib/trellis/plugins/vars/vars.py
+++ b/lib/trellis/plugins/vars/vars.py
@@ -32,9 +32,6 @@ class VarsModule(object):
     def cli_options_ping(self):
         options = []
 
-        remote_user = getattr(self._options, 'remote_user')
-        options.append("--user='{0}'".format(remote_user if remote_user else 'root'))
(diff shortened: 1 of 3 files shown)


swalkinshaw:
👍

PR #584, opened 2016-05-08
Add Vagrant post_up_message

retlehs:
image


swalkinshaw:
See https://github.com/roots/trellis/pull/602

PR #585, opened 2016-05-11
Add wildcard subdomain to https redirect block

bostondv:
This adds a wildcard subdomain entry for each `site_hosts` to the `server_name` https redirect block when multisite subdomains are enabled. Previously subdomains were not being automatically redirected to https.


swalkinshaw:
Sorry for the delay on this. Make sense since we're doing this already in another place 👍

PR #586, opened 2016-05-11
Add nginx varible to configure fastcgi_buffer_size

bostondv:
I needed to increase `fastcgi_buffers` and `fastcgi_buffer_size` on a site I was working on (related to PDF output generation). Since Trellis only has the former value exposed as a variable I've added the later. The default is set to `8k` which matches [nginx's default](http://nginx.org/en/docs/http/ngx_http_fastcgi_module.html#fastcgi_buffer_size).

### DIFF
--- a/roles/nginx/defaults/main.yml
+++ b/roles/nginx/defaults/main.yml
@@ -3,6 +3,7 @@ nginx_path: /etc/nginx
 nginx_logs_root: /var/log/nginx
 nginx_user: www-data
 nginx_fastcgi_buffers: 8 8k
+nginx_fastcgi_buffer_size: 8k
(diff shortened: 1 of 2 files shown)


swalkinshaw:
I'm fine with adding this but the default is actually `4k` on Ubuntu.

> By default, the buffer size is equal to one memory page. This is either 4K or 8K, depending on a platform

```bash
$ getconf PAGESIZE
4096
```

bostondv:
Ah, thanks. I'll revise it to 4k.

bostondv:
Revised!

swalkinshaw:
Thanks :)

PR #591, opened 2016-06-04
Fix tags in the dev.yml.

primozcigler:
Tags for mail, mailhog and ssmtp where not delimited with a comma in
dev.yml, requiring to use something like `--tags 'mailhog mail'` in CLI.

Fixed the same way as the tags in server.yml.


swalkinshaw:
Good eye, thanks.

PR #594, opened 2016-06-13
Update contributing guidelines link in readme

ptibbetts:
This link used to point to `./CONTRIBUTING.md` which has now been moved to `.github/CONTRIBUTING.md` and so I've updated it to the one in the roots/guidelines repo, like the Sage readme does.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ A base Ubuntu 14.04 server is required for setting up remote servers.
 
 ## Contributing
 
-Contributions are welcome from everyone. We have [contributing guidelines](CONTRIBUTING.md) to help you get started.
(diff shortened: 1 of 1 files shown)


ptibbetts:
I've noticed that the roots-example-project.com, docs, and soil repos are missing a `CONTRIBUTING.md` file (in either `./` or `.github`).

Would you be open to some PRs that direct people to the guidelines repo? I was thinking of putting them all in `.github/CONTRIBUTING.md` (same contents as Sage & Trellis) and also adding a contributing link in the readme that points to the guidelines repo, like Sage goes.

I could add the issue template to those repos whilst I'm at it.

Whaddaya think?

swalkinshaw:
I think that sounds like a good idea.

/cc @retlehs

swalkinshaw:
Thanks 👍

PR #596, opened 2016-06-15
Switch to ondrej/nginx PPA for ALPN support

swalkinshaw:
Chrome 51 and newer only support ALPN. This is a negotiation protocol
that HTTP/2 uses. To support ALPN, you need OpenSSL 1.0.2+.

Ubuntu 14.04 unfortunately stops at OpenSSL 1.0.1 so we need to use a
custom PPA in order to support this.


swalkinshaw:
Closing this since it will be fixed by switching to 16.04.

PR #597, opened 2016-06-16
Update includeSubDomains captialization

allanchau:
Fix for Warning: Non-standard capitalization of includeSubDomains


swalkinshaw:
Thanks!

Issue #599, opened 2016-06-17
Can't create custom build before hook for a site


alexandcote:
## Submit a feature request or bug report

- [x] This is a feature request
- [ ] This is a bug report
- [x] This request isn't a duplicate of an [existing issue](https://github.com/roots/trellis/issues)
- [x] I've read the [docs](https://roots.io/trellis/docs) and followed them (if applicable)

---

**What is the current behavior?**

You can only have one build before hook for all websites

**What is the expected or desired behavior?**

If you have custom build before hook for one website, you can create a file for this site

---

## Feature Request

**Please provide use cases for changing the current behavior:**

Support old version of root/sage (grunt/gulp) and support other theme

PR #600, opened 2016-06-17
Support multiple site build before hook

alexandcote:
Related to #599.
Feel free to propose any change.

### DIFF
--- a/deploy.yml
+++ b/deploy.yml
@@ -8,7 +8,7 @@
   remote_user: "{{ web_user }}"
 
   vars:
-    deploy_build_before: "{{ playbook_dir }}/deploy-hooks/build-before.yml"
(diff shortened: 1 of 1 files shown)


swalkinshaw:
Thanks for proposing this @alexandcote but I don't think this is the proper solution yet.

1. I just realized that `deploy_build_before` isn't even used in Trellis by default so this line should be removed which I'll do.
2. If we were going to support hooks *per* site, we should enable it for all of them which is more complicated.

If you need to customize a hook in the meantime, I'd just use a `when` conditional on tasks within the hook file.

```yaml
- name: site specific task
  action: whatever
  when: site == 'mysite.com'
```

swalkinshaw:
Oops I was mistaken about my first point. That file does exist as an example file.

PR #608, opened 2016-06-30
Update per-item callback names for Ansible 2.0.2.0

fullyint:
This PR updates the Trellis output plugin to use the [callback names modified in Ansible 2.0.2.0](https://github.com/ansible/ansible/commit/9d46a16f3f38819d3c567cea119dc85d0da7618e#diff-ab09fa2a9291a017c38c5db420d7ee0fL323).

Otherwise textwrap will not be applied to the `msg` in tasks that use `with_items`.


swalkinshaw:
👍

PR #604, opened 2016-06-27
Change to Ubuntu 16.04 Xenial LTS

ptibbetts:
This PR is a work-in-progress, aiming to replace Ubuntu 14.04 Trusty with 16.04 Xenial. 

**Todo:**

- [x] confirm if remote-user role should change from ping to whoami
- [x] [wait for Vagrant 1.8.5 to be released](https://github.com/roots/trellis/pull/604#issuecomment-228943575)
- [x] switch Vagrant boxes to [`bento/ubuntu-16.04`](https://atlas.hashicorp.com/bento/boxes/ubuntu-16.04)
- [ ] add support for vmware_workstation
- [ ] add support for parallels

https://github.com/roots/trellis/pull/604/commits/9751f498265d28bee50085a58c8bae3c2a6bf552 is based on the method described in [this post on the Roots Discourse](https://discourse.roots.io/t/trellis-ubuntu-16-04-do/6551/3?u=paul_tibbetts) which

* changed the Vagrant box to `geerlingguy/ubuntu1604`, which ships with Python 2.7.11+
* changed the Ansible remote-user role from `ping`ing the external server to running a raw `whoami`. It also removes `cli_options_ping` from `vars.py` in the Trellis library.
* added [python-minimal](http://packages.ubuntu.com/xenial/python2.7-minimal) to the remote server using a raw command as the admin user. This is root on the first run of the playbook and if another admin user is chosen the command will run (and pass) as them in the future. [Debian describes](https://packages.debian.org/wheezy/python2.7-minimal) python-minimal as "… the interpreter and some essential modules. It can be used in the boot process for some basic tasks" which is all we need

PR #611, opened 2016-07-02
Support SSH password authentication

fullyint:
This PR adds support for SSH password authentication via the `--ask-pass` option.
Each commit message provides relevant details.

Props to @ptibbetts for discovering in #604 that the `--ask-pass` option does not function properly in [cli_options_ping](https://github.com/roots/trellis/blob/f36bcb5f04d453421fe1b73190bd8df8092e7da8/roles/remote-user/tasks/main.yml#L3). Adjusting the `ansible.cfg` section `[ssh_connection]` to include `pipelining = False` did not resolve the error `(25, 'Inappropriate ioctl for device')` as it may have for others, nor did adjusting the `local_action` to use the `shell` or `raw` module (instead of the `command` module).


swalkinshaw:
🙌

PR #613, opened 2016-07-07
Install ansible 2.0.2 in windows via pip

danstefancu:
Ansible 2.1.0 has a bug, and apt cannot install 2.0.2 version that is not afected.

See: https://github.com/ansible/ansible/issues/15915
See: https://discourse.roots.io/t/taskinclude-object-has-no-attribute-has-triggered/6834/5

### DIFF
--- a/windows.sh
+++ b/windows.sh
@@ -31,12 +31,11 @@ fi
 
 # Install Ansible and its dependencies if not installed.
 if [ ! -f /usr/bin/ansible ]; then
-  echo "Adding Ansible repository..."
-  sudo apt-add-repository -y ppa:ansible/ansible
-  echo "Updating system..."
-  sudo apt-get -y update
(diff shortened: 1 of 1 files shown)


danstefancu:
Well, it's not really an issue, as pip is only used to install ansible 2.0.2. Building the entire dependency tree for pip will just add time to the installation and nothing else.

Tested the ansible to setup a server and deploy a release, with this patched version, and everything wrks as expected.

danstefancu:
Note: this patch is for short term usage (couple of weeks), and can be removed once ansible 2.1.1 is stable and published to launchpad, to be installed via apt-get. Without this, trellis is unusable from a windows host.

fullyint:
> Tested the ansible to setup a server and deploy a release, with this patched version, and everything works as expected.

Works for me too. Thanks!

PR #614, opened 2016-07-09
Add vault_users for easier password management

fullyint:
**Automates password hashing** so people don't have to create their own hash for `sudoer_passwords`.

**Automates become pass submission** so people don't have to use `--ask-become-pass`.
When `--ask-become-pass` is used, its password will take precedence.

~~**Disables `root` login by default,** but only after confirmation that the `admin_user` can connect.~~
**Edit:** we can dedicate a separate PR to disabling `root` login.

roots/docs#43

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ### HEAD
-* Add `vault_users` for easier password management ([#614](https://github.com/roots/trellis/pull/614))
+* [BREAKING] Add `vault_users` for easier password management ([#614](https://github.com/roots/trellis/pull/614))
 * Fix #581 - Use WP-CLI to run WP cron ([#583](https://github.com/roots/trellis/pull/583))
-* Require explicit redirects and drop www_redirect ([#622](https://github.com/roots/trellis/pull/622))
+* [BREAKING] Require explicit redirects and drop `www_redirect` ([#622](https://github.com/roots/trellis/pull/622))
 * Fix #612 - Bump nginx_fastcgi_buffer_size to `8k` ([#620](https://github.com/roots/trellis/pull/620))
 * Setup permalink structure for multisite installs too ([#617](https://github.com/roots/trellis/pull/617))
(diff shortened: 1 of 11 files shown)


swalkinshaw:
Any ideas about encouraging the use of Ansible Vault? It becomes more important with this change.

At a minimum we should probably emphasize it more in the docs.

fullyint:
>Any ideas about encouraging the use of Ansible Vault?

I have a "tips" role largely worked out, following the pattern of #562's proposed "validations" role. Tips print to the end of playbook stdout. One is conditional on a custom `unencrypted_vault_files` _bool_ created using [`is_encrypted`](https://github.com/ansible/ansible/blob/v2.0.2.0-1/lib/ansible/parsing/vault/__init__.py#L100) from `VaultLib`:
> Tip: Some vault files are not encrypted. Set up Ansible Vault following the guide at ... [etc]

Users can disable tips and/or validations with a simple toggle in `group_vars` or `--extra-vars`.

Not wanting to overload, I was waiting to finalize/submit the "tips" PR till after the "validations" PR had been considered.

There could certainly be other ideas for encouraging Vault usage.

fullyint:
@swalkinshaw pointed out that the playbook fails if `vault_users` is not defined. It does not need to be defined when `sshd_permit_root_login: true` (current default).

I added a few instances of `vault_users | default([])` to prevent unnecessary failure.

swalkinshaw:
Trying this out and I get this:

> |password_hash requires the passlib python module to generate password hashes
on Mac OS X/Darwin

swalkinshaw:
After a `pip install password_hash` everything worked great 👍

fullyint:
Thanks for testing! I squashed in a test and message at the top of `vars.py` suggesting:
>Ansible on OS X requires the python passlib module to create user password hashes.
sudo easy_install pip
pip install passlib

swalkinshaw:
💯

PR #615, opened 2016-07-13
Add list of raw_vars whose values are auto-escaped

fullyint:
#548 wrapped WP salts/keys in `{% raw %}` to avoid templating failures due to values containing opening jinja delimiters `{{` and `{%`. This was not user-configurable (no easy way to opt-out of the magic) and it didn't wrap other variables likely to have randomly generated values, leaving the potential for problematic jinja delimiters.

This PR replaces the #548 method with a new list of `raw_vars` as a user-configurable method to wrap chosen variable values in `{% raw %}`.

This doesn't seem to be a very risky use of the potentially evolving Ansible plugin API. The only new API material introduced is the checking of Ansible-specific classes `AnsibleMapping`, `AnsibleSequence`, and `AnsibleUnicode` in the `raw_triage` method. These classes have [been around](https://github.com/ansible/ansible/blob/v2.0.0.0-1/lib/ansible/parsing/yaml/objects.py) all of Ansible 2.0.

-----
To see it work, pull this branch then add the vars below to the end of `group_vars/all/main.yml`:
```
simple_var: no {{ fail

complex_var:
  - name: "{{ admin_user }}"
    password: random_with_{{_jinja_delimiter
    salt: ran{%dom
    some_list:
      - no {{ fail
      - templated -> {{ admin_user }}
  - name: "{{ web_user }}"
    some_dict:
      key_1: no {% fail
      key_2: no {{ fail
      other_key: templated -> {{ web_user }}

raw_vars:
  - simple_var
  - complex_var.*.password
  - complex_var.*.salt
  - complex_var.*.some_list.0
  - complex_var.*.some_dict.key_*
```

Now add this task to the end of the `Determine Remote User` play in `server.yml`:
```
  tasks:
    - debug:
        msg: |
          {{ simple_var | to_nice_yaml }}
          {{ complex_var | to_nice_yaml }}
      tags: remote-user
```

Now run `server.yml` to print the vars, seeing that the stray jinja delimiters don't cause failures:
```
ansible-playbook server.yml -e env=production --tags remote-user
```

### DIFF
--- a/lib/trellis/plugins/vars/vars.py
+++ b/lib/trellis/plugins/vars/vars.py
@@ -41,6 +41,9 @@ class VarsModule(object):
         if 'raw_vars' not in hostvars:
             return
 
+        if not isinstance(hostvars['raw_vars'], AnsibleSequence):
+            raise AnsibleError('The `raw_vars` variable must be defined as a list.')
+
         patterns = [re.sub(r'\*', '(.)*', re.sub(r'\.', '\.', var)) for var in hostvars['raw_vars'] if var.split('.')[0] in hostvars]
         keys = set(pattern.split('\.')[0] for pattern in patterns)
         for key in keys:
--- a/lib/trellis/plugins/vars/vars.py
+++ b/lib/trellis/plugins/vars/vars.py
@@ -1,6 +1,8 @@
 from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
+import re
+
 from ansible import __version__
 from ansible.errors import AnsibleError
 
@@ -10,6 +12,7 @@ if __version__.startswith('1'):
 # These imports will produce Traceback in Ansible 1.x, so place after version check
 from __main__ import cli
 from ansible.compat.six import iteritems
+from ansible.parsing.yaml.objects import AnsibleMapping, AnsibleSequence, AnsibleUnicode
 
 
 class VarsModule(object):
@@ -20,14 +23,57 @@ class VarsModule(object):
         self.inventory_basedir = inventory.basedir()
         self._options = cli.options if cli else None
 
-    # Wrap salts and keys variables in {% raw %} to prevent jinja templating errors
-    def wrap_salts_in_raw(self, host, hostvars):
-        if 'vault_wordpress_sites' in hostvars:
-            for name, site in hostvars['vault_wordpress_sites'].iteritems():
-                for key, value in site['env'].iteritems():
-                    if key.endswith(('_key', '_salt')) and not value.startswith(('{% raw', '{%raw')):
-                        hostvars['vault_wordpress_sites'][name]['env'][key] = ''.join(['{% raw %}', value, '{% endraw %}'])
-            host.vars['vault_wordpress_sites'] = hostvars['vault_wordpress_sites']
+    def raw_triage(self, key_string, item, patterns):
+        # process dict values
+        if isinstance(item, AnsibleMapping):
+            dict = {}
+            for key,value in item.iteritems():
+                dict[key] = self.raw_triage('.'.join([key_string, key]), value, patterns)
+            return dict
+
+        # process list values
+        elif isinstance(item, AnsibleSequence):
+            list = []
+            for i,value in enumerate(item):
+                raw = self.raw_triage('.'.join([key_string, str(i)]), value, patterns)
+                list.append(raw)
+            return list
+
+        # wrap values if they match raw_vars pattern
+        elif isinstance(item, AnsibleUnicode):
+            matches = False
+            for pattern in patterns:
+                if re.match(pattern, key_string) is not None:
+                    matches = True
+                    break
+
+            if not item.startswith(('{% raw', '{%raw')) and matches:
+                item = ''.join(['{% raw %}', item, '{% endraw %}'])
+
+            return item
+
+    def raw_vars(self, host, hostvars):
+        if 'raw_vars' not in hostvars:
+            return
+
+        raw_vars = list((var for var in hostvars['raw_vars'] if var.split('.')[0] in hostvars))
+
+        # prepare regex match patterns
+        patterns = []
+        for pattern in raw_vars:
+            pattern = re.sub(r'\.', '\.', pattern)
+            pattern = re.sub(r'\*', '(.)*', pattern)
+            patterns.append(pattern)
+
+        # wrap matching vars under each key
+        keys = set()
+        for var in raw_vars:
+            key = var.split('.')[0]
+            if key in keys:
+                continue
+
+            host.vars[key] = self.raw_triage(key, hostvars[key], patterns)
+            keys.add(key)
 
     def cli_options(self):
         options = []
@@ -52,7 +98,7 @@ class VarsModule(object):
         return ' '.join(options)
 
     def get_host_vars(self, host, vault_password=None):
-        self.wrap_salts_in_raw(host, host.get_group_vars())
+        self.raw_vars(host, host.get_group_vars())
         host.vars['cli_options'] = self.cli_options()
         host.vars['cli_ask_pass'] = getattr(self._options, 'ask_pass', False)
         return {}
--- a/lib/trellis/plugins/vars/vars.py
+++ b/lib/trellis/plugins/vars/vars.py
@@ -26,54 +26,25 @@ class VarsModule(object):
     def raw_triage(self, key_string, item, patterns):
         # process dict values
         if isinstance(item, AnsibleMapping):
-            dict = {}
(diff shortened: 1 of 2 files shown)


swalkinshaw:
This is good but think I'd rather skip all the wildcard stuff for now. It's currently unused by default and not sure it's worth the potential hassle.

fullyint:
We wouldn't want the wildcard functionality to become a wildcard of its own. Surprises not welcome.

But to explain, the wildcard functionality stemmed from development on #614 with its new var:
```
vault_users:
  - name: "{{ admin_user }}"
    password: example_password
    salt: "generateme"
```
This var would require the wildcard, a la:
```
raw_vars:
  - vault_users.*.password
  - vault_users.*.salt
```

lib/trellis/plugins/vars/vars.py
```diff
@@ -20,14 +23,57 @@ def __init__(self, inventory):
...
+                    break
+
+            if not item.startswith(('{% raw', '{%raw')) and matches:
+                item = ''.join(['{% raw %}', item, '{% endraw %}'])
+
+            return item
+
+    def raw_vars(self, host, hostvars):
+        if 'raw_vars' not in hostvars:
+            return
```
swalkinshaw: I wonder if it's better to move the default vars into here.

Pros: users shouldn't have to worry about these too much so they are kept out of the way and harder to mess up.
Cons: to disable them would require  `raw_vars: []` and maybe less visibility?

lib/trellis/plugins/vars/vars.py
```diff
@@ -20,14 +23,57 @@ def __init__(self, inventory):
...
+        for pattern in raw_vars:
+            pattern = re.sub(r'\.', '\.', pattern)
+            pattern = re.sub(r'\*', '(.)*', pattern)
+            patterns.append(pattern)
+
+        # wrap matching vars under each key
+        keys = set()
+        for var in raw_vars:
+            key = var.split('.')[0]
+            if key in keys:
```
swalkinshaw: So this just prevent duplicate vars? Like in this case:


```yaml
raw_vars:
  - vault_foo.*
  - valut_foo.*.bar
```

The 2nd one would be skipped?

lib/trellis/plugins/vars/vars.py
```diff
@@ -20,14 +23,57 @@ def __init__(self, inventory):
...
+                list.append(raw)
+            return list
+
+        # wrap values if they match raw_vars pattern
+        elif isinstance(item, AnsibleUnicode):
+            matches = False
+            for pattern in patterns:
+                if re.match(pattern, key_string) is not None:
+                    matches = True
+                    break
```
swalkinshaw:
```python
match = next((pattern for pattern in patterns if re.match(pattern, key_string)), None)
```

Think that replaces the above 5 lines

lib/trellis/plugins/vars/vars.py
```diff
@@ -20,14 +23,57 @@ def __init__(self, inventory):
...
+                dict[key] = self.raw_triage('.'.join([key_string, key]), value, patterns)
+            return dict
+
+        # process list values
+        elif isinstance(item, AnsibleSequence):
+            list = []
+            for i,value in enumerate(item):
+                raw = self.raw_triage('.'.join([key_string, str(i)]), value, patterns)
+                list.append(raw)
+            return list
```
swalkinshaw: Comprehension maybe?

```python
return [self.raw_triage('.'.join([key_string, str(i)]), value, patterns) for i,value in enumerate(item)]
```

lib/trellis/plugins/vars/vars.py
```diff
@@ -20,14 +23,57 @@ def __init__(self, inventory):
...
-                    if key.endswith(('_key', '_salt')) and not value.startswith(('{% raw', '{%raw')):
-                        hostvars['vault_wordpress_sites'][name]['env'][key] = ''.join(['{% raw %}', value, '{% endraw %}'])
-            host.vars['vault_wordpress_sites'] = hostvars['vault_wordpress_sites']
+    def raw_triage(self, key_string, item, patterns):
+        # process dict values
+        if isinstance(item, AnsibleMapping):
+            dict = {}
+            for key,value in item.iteritems():
+                dict[key] = self.raw_triage('.'.join([key_string, key]), value, patterns)
+            return dict
```
swalkinshaw: Could use a comprehension here as well

lib/trellis/plugins/vars/vars.py
fullyint: The primary purpose of recursing only once through each top-level key is to avoid having multiple lists or dicts to merge for keys at various depths. No variable value is touched more than once. Each top-level key needs only one instance of the following and avoids merges:
```
host.vars[key] = self.raw_triage(key, hostvars[key], patterns)
```

A prior iteration had the potential to produce multiple dicts and lists per key, which I tried to merge using Ansible's [`merge_hash`](https://github.com/ansible/ansible/blob/v2.0.2.0-1/lib/ansible/utils/vars.py#L71_). There were edge-case merge failures, with nested values inappropriately overwitten (particularly for lists).

It is only a secondary benefit that we do no more work than for the simplest equivalent:
```yaml
raw_vars:
  - vault_foo
```

lib/trellis/plugins/vars/vars.py
fullyint: That "visibility" was my focus. My overarching goal was to transition this functionality from "hidden" -- which for users is possibly surprising and certainly non-configurable -- to "out in the open" and easy to use.

I agree most users won't understand it and won't touch it, in which case they won't mess it up. In that sense, I think it is like neighboring obscure vars in `group_vars/all/main.yml`, such as `wordpress_env_defaults` or `site_env`. 

For users who do try to use it, actually adding patterns versus merely disabling it, I think they'll have the easiest time if they readily see the full list of `raw_vars` default patterns right there in a group_vars file.

Would users really want to adjust patterns? I think a conceivable example would be that someone wants to add an env var in `vault_wordpress_sites` that should _not_ be wrapped in `{% raw %}`. They'd need to adjust the `raw_vars` to be more specific, just wrapping the right parts, such as:
```
raw_vars:
  - vault_wordpress_sites.*.*_password       # for `admin_password` (dev only) and `db_password`
  - vault_wordpress_sites.*.*_salt
  - vault_wordpress_sites.*.*_key
```

On the other hand, if we decide that visibility and configurability aren't the goal or aren't worth it, then I suppose moving the defaults into `vars.py` makes sense. In that case, we could consider whether the python code would be "safer" being more targeted to those defaults, not supporting the wildcard stuff and being more "hard-coded" like it has been for WP keys and salts.

lib/trellis/plugins/vars/vars.py
swalkinshaw: Unless this isn't the same thing, it's clearer to filter the list first:

```python
raw_vars = set([var.split('.')[0] for key in raw_vars])

for var in raw_vars:
    host.vars[key] = self.raw_triage(key, hostvars[key], patterns)
```

fullyint:
I incorporated @swalkinshaw's suggested comprehensions; a remarkable streamlining of the code.

----
I added a commit to raise an error if `raw_vars` is not defined as a list.

-----
The `vars.py` plugin accesses vars before they are templated. This is what enables us to wrap vars in `{% raw %}` before a templating problem occurs. 

However, this means that the `raw_vars` variable will not be templated/expanded here unless we do it ourselves. I added a commit to template the `raw_vars` variable before working with it.

```yaml
some_var:
  - vault_mail_password
  - vault_mysql_root_password
  - vault_wordpress_sites

raw_vars: "{{ some_var }}"
```

If a user were to create the unlikely scenario above, I think it would be a baffling surprise/mystery why `some_var` wasn't being templated/expanded in the definition of `raw_vars`. I introduce some code to avoid such a problem. I don't think the code has too high of a price.

- The [`Templar` params](https://github.com/ansible/ansible/blob/v2.0.0.0-1/lib/ansible/template/__init__.py#L125) seem relatively stable, unchanged during history of Ansible 2.x
- The [Templar.template() params](https://github.com/ansible/ansible/blob/v2.0.0.0-1/lib/ansible/template/__init__.py#L272) have had only one [minor irrelevant change](https://github.com/ansible/ansible/commit/dc15eb806e1b5c040aedb0565218d26c369eb453#diff-c6c175f84d40c7d6fd35ebe53793bdddR278) during history of 2.x
- [`DataLoader`](https://github.com/ansible/ansible/blob/v2.0.0.0-1/lib/ansible/parsing/dataloader.py) has been around all of 2.x

I put `self.loader` in `__init__` so it is accessible to other methods to use for templating in the future.

swalkinshaw:
Works great 🎉

PR #619, opened 2016-07-21
Fix subdomain detection in reverse_www

swalkinshaw:
Replaces our naive detection with querying the tldextract service.

This isn't 100% ideal since it relies on an external request which might fail. But for now it's better than trying to bundle tldextract itself in Trellis and have to maintain the TLD data list.


swalkinshaw:
Going to close this. It will be replaced with a proper solution to come soon and we shouldn't rely on these network calls which could cause problems.

PR #623, opened 2016-07-27
Updated windows.sh

TheAggressive:
Added apt update before installing Ansible to fix a issue in windows.

### DIFF
--- a/windows.sh
+++ b/windows.sh
@@ -32,8 +32,8 @@ fi
 # Install Ansible and its dependencies if not installed.
 if [ ! -f /usr/bin/ansible ]; then
   echo "Installing pip..."
-  sudo apt-get -y install python-pip
   sudo apt-get -y update
+  sudo apt-get -y install python-pip
   echo "Installing Ansible with pip..."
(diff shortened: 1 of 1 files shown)


windows.sh
```diff
@@ -33,6 +33,7 @@ fi
 if [ ! -f /usr/bin/ansible ]; then
   echo "Installing pip..."
   sudo apt-get -y install python-pip
+  sudo apt-get -y update
```
swalkinshaw: Shouldn't this be before the line above it?:

```bash 
sudo apt-get -y update
sudo apt-get -y install python-pip
```

TheAggressive: It seems both ways work the same. I could change it to how you have it before installing pip.

swalkinshaw: Seems like it should be before. Are you testing from scratch to test? Anyway, let's put it before 👍 and we're good

TheAggressive: I am testing from scratch to verify it's working correctly. I just made that change so its before.

swalkinshaw:
Thanks 👍

PR #628, opened 2016-08-03
Validate user salts for allowable characters

fullyint:
#614 introduced ansible's [password_hash filter](http://docs.ansible.com/ansible/playbooks_filters.html#hashing-filters) to Trellis. The [filter uses either](https://github.com/ansible/ansible/blob/v2.0.2.0-1/lib/ansible/plugins/filter/core.py#L238-L241) the Python crypt function ([v2](https://docs.python.org/2/library/crypt.html) or [v3](https://docs.python.org/3/library/crypt.html#module-functions)) or [passlib.hash.sha512_crypt](https://pythonhosted.org/passlib/lib/passlib.hash.sha512_crypt.html#passlib.hash.sha512_crypt), both of which limit salts to 1-16 characters from the regexp range `[./a-zA-Z0-9]`.

Currently, if a salt has a disallowed character, the "Setup users" task fails with 
```
the field 'args' has an invalid value ([u'users']), and could not be
converted to an dict. Error was: invalid characters in sha512_crypt salt
```

Currently, if a salt has more than 16 characters, the task fails with 
```
the field 'args' has an invalid value ([u'users']), and could not be
converted to an dict. Error was: salt too large (sha512_crypt requires <= 16
chars)
```

This PR adds
- a validation step to halt and warn users about salts with disallowed characters
- a comment about character limitations next to the `vault_users` definition
- a slice to use only the first 16 characters of any `salt` (`user.salt[:16]`)

### DIFF
--- a/group_vars/all/helpers.yml
+++ b/group_vars/all/helpers.yml
@@ -11,3 +11,4 @@ site_env: "{{ wordpress_env_defaults | combine(item.value.env | default({}), vau
 site_hosts_canonical: "{{ item.value.site_hosts | map(attribute='canonical') | list }}"
 site_hosts_redirects: "{{ item.value.site_hosts | selectattr('redirects', 'defined') | sum(attribute='redirects', start=[]) | list }}"
 site_hosts: "{{ site_hosts_canonical | union(site_hosts_redirects) }}"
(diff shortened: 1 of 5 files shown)


fullyint:
replaced by #629

PR #629, opened 2016-08-04
Transform vault_user salts into allowable chars

fullyint:
Explanation in #628

Opting for this simpler solution suggested by @swalkinshaw which makes for a superior user experience. In this current implementation, the user doesn't have to do a thing.

This PR limits the salt length to the 16 character max and replaces disallowed characters with an allowable character.

This slightly limits the entropy in the password hashing, but the salts were already limited (char length and char set). The tradeoff seems worth it for the sake of not bothering the user with a failed playbook and requirement to manually edit salts down to allowable characters.


swalkinshaw:
👍

PR #630, opened 2016-08-05
Enable Let's Encrypt to detect updated site_hosts

fullyint:
### Update CSRs and certs

The primary problem to solve: If a user has already created Let's Encrypt CSRs and certs then modifies `site_hosts`, the LE role will fail to update the CSRs/certs.

If an LE cert exists, it will not update [if not older than 60 days](https://github.com/roots/trellis/blob/6ad6fcbb6e0c0cfdaaf7ecb734c2e867ff0437a7/roles/letsencrypt/templates/renew-certs.py#L22). If a CSR exists, it [will not be recreated](https://github.com/roots/trellis/blob/6ad6fcbb6e0c0cfdaaf7ecb734c2e867ff0437a7/roles/letsencrypt/tasks/certificates.yml#L21) because the [`creates` parameter](http://docs.ansible.com/ansible/shell_module.html) specifies the CSR file, and "when it already exists, this step will not be run." 

This PR updates the LE role to detect changes to `site_hosts` and update the CSR. The PR also updates the `renew-certs.py` to detect changes to the CSR and update the cert, even if younger than 60 days.

~~The CSR updates become idempotent via a new rsync like Trellis already uses for [the .env file](https://github.com/roots/trellis/blob/6ad6fcbb6e0c0cfdaaf7ecb734c2e867ff0437a7/roles/wordpress-install/tasks/main.yml#L5-L17) (details on rationale of using [rsync for idempotence](https://github.com/roots/trellis/pull/268#issuecomment-119903331)).~~

### Add acme challenge location to conf

**Edit: The remainder of this comment is now obsolete after internal discussion and a decision to avoid the `replace` module.**

If a user has just updated to a version of Trellis that includes the LE role, the LE role's [Test Acme Challenges](https://github.com/roots/trellis/blob/6ad6fcbb6e0c0cfdaaf7ecb734c2e867ff0437a7/roles/letsencrypt/tasks/nginx.yml#L38) would fail because the `wordpress-setup` role hasn't yet had a chance to insert `include acme-challenge-location.conf;` into the conf. This challenge location could also be missing for users with current Trellis with ssl [enabled with a non-LE provider](https://github.com/roots/trellis/blob/6ad6fcbb6e0c0cfdaaf7ecb734c2e867ff0437a7/roles/wordpress-setup/templates/wordpress-site.conf.j2#L86). This missing challenge location has required users to run the `wordpress` role first to update their conf before running the `letsencrypt` role, an unexpected hassle.

This PR updates the LE role to detect the missing challenge location and inserts it into the conf so that users no longer must run the `wordpress` role first. This feature is convenient but relies on the [replace module](http://docs.ansible.com/ansible/replace_module.html), which introduces some risk. There are some reassurances, however.
- The `replace` task will not even run if the challenge location is already in the conf.
- If an unforeseen regex failure messes up the conf, the user can fix it by running the `wordpress` tag, which they would have had to do anyway without this attempted `replace`. Then the challenge location will be in the conf and the `replace` won't ever run again for that user/project.
- The `replace` regex is designed to remove nothing. It should return everything matched, just adding the challenge location conf.

Explaining some regex (suggestions welcome)
example: `(listen 80;(?:[^\{]*\n)*)(.*return 301 \$scheme:\/\/.*\$request_uri;)`
- `(listen 80;(?:[^\{]*\n)*)` -- backref 1
  - `listen 80;` -- this portion of the match must begin with `listen 80;`
  - `(?:[^\{]*\n)*`
    - The `?:` prevents the creation of a separate backref for this `(?:pattern)`.
    - The `[^\{]*\n` will match any character(s) -- _except_ `{` -- followed by a newline. Excluding the `{` prevents matching across server/location blocks. For example, we wouldn't want a match to start at the `listen 80;` from the "ssl redirect server block" and finish on the `return 301` in the following "general redirects server block."
    - The final `*` lets there be multiple instances of `[^\{]*\n`
- `(.*return 301 \$scheme:\/\/.*\$request_uri;)` -- backref 2

The match patterns appear applicable to the conf for the past year (back to [ff0fbff](https://github.com/roots/trellis/blob/ff0fbff1f0ef863ae936cf357972815d0ea4ded2/roles/wordpress-setup/templates/wordpress-site.conf.j2), June 30, 2015).
### Handlers

**Edit:** The handler adjustment separated out to #722.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Enable Let's Encrypt to detect updated `site_hosts` ([#630](https://github.com/roots/trellis/pull/630))
 * Add `SKIP_GALAXY` env var to skip galaxy install in Vagrant ([#734](https://github.com/roots/trellis/pull/734))
 * Avoid `loop.first` variable in conditional jinja loops ([#729](https://github.com/roots/trellis/pull/729))
 * Use dynamic `local_path` to accommodate Ansible running on VM ([#725](https://github.com/roots/trellis/pull/725))
--- a/roles/letsencrypt/defaults/main.yml
+++ b/roles/letsencrypt/defaults/main.yml
@@ -1,7 +1,7 @@
 sites_using_letsencrypt: "[{% for name, site in wordpress_sites.iteritems() if site.ssl.enabled and site.ssl.provider | default('manual') == 'letsencrypt' %}'{{ name }}',{% endfor %}]"
-letsencrypt_enabled: "{{ sites_using_letsencrypt | count > 0 }}"
+letsencrypt_enabled: "{{ sites_using_letsencrypt | count }}"
 site_uses_letsencrypt: "{{ ssl_enabled and item.value.ssl.provider | default('manual') == 'letsencrypt' }}"
-sites_need_confs: "False in [{% for item in nginx_confs.results if 'stat' in item %}{{ item.stat.exists }},{% endfor %}]"
+missing_hosts: "{{ site_uses_letsencrypt | ternary(site_hosts, []) | difference((current_hosts.results | selectattr('item.key', 'equalto', item.key) | selectattr('stdout_lines', 'defined') | sum(attribute='stdout_lines', start=[]) | map('trim') | list | join(' ')).split(' ')) }}"
 
 acme_tiny_repo: 'https://github.com/diafygi/acme-tiny.git'
 acme_tiny_commit: '5a7b4e79bc9bd5b51739c0d8aaf644f62cc440e6'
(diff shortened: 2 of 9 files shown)


fullyint:
After generous and painstaking review from @swalkinshaw, I think this is ready.

This PR originally intended to solve the problem that CSRs and certs would not detect changes in `site_hosts` and would fail to update.

Upon further review, I see that other crucial changes would go undetected, but should trigger cert regeneration, i.e., changes in any of the following:

* var: `site_hosts` (the original issue of this PR)
* var: `letsencrypt_intermediate_cert_sha256sum`
* var: `acme_tiny_commit`
* var: `letsencrypt_ca`
* file: `/etc/nginx/ssl/letsencrypt/{{ site }}.key`
* file: `/var/lib/letsencrypt/account.key`

The revised 2nd commit hashes the combination of the above, per site, inserting a corresponding 7-digit ID into CSR and cert filenames. If any item above changes, the required filename will change, triggering creation of new CSRs and certs (if they don't already exist with this ID).

-----

## How to update

### New servers
New servers provisioned with Trellis (after this PR is merged) will accommodate changes to `site_hosts` or any of the above items.

### Existing servers that do NOT need to add hosts to `site_hosts`
1. Update trellis (get changes from this PR)
2. Run `ansible-playbook server.yml -e env=<environment> --tags letsencrypt`

Then, any future changes to `site_hosts` etc. will work no problem for this project.

These steps should also work for existing servers that need to REMOVE hosts from `site_hosts`.

### Existing servers that DO need to add hosts to `site_hosts`
1. Update trellis (get changes from this PR)
2. Set ssl `enabled: false` for all sites in `group_vars/<environment>/wordpress_sites.yml`
3. Run `ansible-playbook server.yml -e env=<environment> --tags wordpress`
4. Set ssl `enabled: true` for applicable sites in `group_vars/<environment>/wordpress_sites.yml`
5. Run `ansible-playbook server.yml -e env=<environment> --tags letsencrypt`

Then, any future changes to `site_hosts` etc. should work no problem for this project.

Existing servers that use non-Let's Encrypt certs and are transitioning to Let's Encrypt certs will probably need to use these 5 steps.

swalkinshaw:
Let's go forward with this 👍 

Needs changelog + maybe docs?

swalkinshaw:
Awesome work :shipit:

PR #642, opened 2016-08-27
Create WordPress php-fpm conf after web_root exists

fullyint:
Resolves the problem -- and implements the suggestion -- in https://discourse.roots.io/t/failed-starting-php7-0-fpm-service/7444


swalkinshaw:
👏

PR #644, opened 2016-08-28
Update README requirements re: Ansible and passlib

fullyint:
- clarifies Ansible version requirement
- makes OS X `passlib` requirement explicit in README (for provisioning remote servers only) 
- omits `passlib` test/msg from dev (dev lacks `users` role; `passlib` not needed to hash password)

Note: `passlib` will most likely already be on OS X machines for users who generated a [crypted password](http://docs.ansible.com/ansible/faq.html#how-do-i-generate-crypted-passwords-for-the-user-module) for the defunct `vault_sudoer_passwords`. #614 removed `vault_sudoer_passwords` but added the requirement for `passlib` for OS X users provisioning remote servers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Trellis will configure a server with the following and more:
 
 Make sure all dependencies have been installed before moving on:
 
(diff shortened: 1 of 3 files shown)


swalkinshaw:
👍  much better

Issue #646, opened 2016-08-31
WordPress 4.6 issues


swalkinshaw:
This is the new main tracking issue for 4.6 issues.

Related issues:

* https://github.com/roots/trellis/issues/645
* https://github.com/roots/trellis/issues/639

And Discourse threads:

* https://discourse.roots.io/t/uncaught-phpmailerexception-error-on-current-trellis-install-wp/7436
* https://discourse.roots.io/t/cannot-deploy-wp-4-6-with-trellis/7461

There are 2 distinct issues here:

* Emails are failing causing WP install to be shown as failed in development. WP actually gets installed and the site does work. The error message looks like this:

```
TASK [wordpress-install : Install WP] ******************************************
System info:
  Ansible 2.0.2.0; Vagrant 1.8.5; Darwin
  Trellis at "Create WordPress php-fpm conf after web_root exists"
---------------------------------------------------
PHP Fatal error:  Uncaught phpmailerException: Invalid address:
wordpress@${http_host} in /srv/www/example.com/current/web/wp/wp-includes
/class-phpmailer.php:946
Stack trace:
#0 /srv/www/example.com/current/web/wp/wp-includes/pluggable.php(352):
PHPMailer->setFrom('wordpress@${htt...', 'WordPress')
#1 /srv/www/example.com/current/web/wp/wp-admin/includes/upgrade.php(392):
wp_mail('admin@example.d...', 'New WordPress S...', 'Your new WordPr...')
#2 /srv/www/example.com/current/web/wp/wp-admin/includes/upgrade.php(95):
wp_new_blog_notification('example.com', 'http://${HTTP_H...', 1, 'The
password yo...')
#3 phar:///usr/bin/wp/php/commands/core.php(763): wp_install('example.com',
'admin', 'admin@example.d...', true, '', 'admin')
#4 phar:///usr/bin/wp/php/commands/core.php(573):
Core_Command->_install(Array)
#5 [internal function]: Core_Command->install(Array, Array)
#6 phar:///usr/bin/wp/php/WP_CLI/Dispatcher/CommandFactory.php(67):
call_user_func(Array, Array, Array)
#7 [internal function]: WP_CLI\Dispatcher\CommandFactory::WP_C in
/srv/www/example.com/current/web/wp/wp-includes/class-phpmailer.php on line
946
[...]

PR #647, opened 2016-08-31
WP 4.6 Compat: set WP_HOME/SITEURL directly

swalkinshaw:
https://github.com/WordPress/WordPress/commit/905f4ec0f835dc9eb0a34f37c297bd6a02d519ba
changed email address handling with PHPMailer. It's not validating email
addresses. This was failing since Trellis was using string interpolation
with the `HTTP_HOST` constant. It wasn't being evaluated in that
context.

The solution is to set these values directly with the first canonical
site host eliminating the need for the magic constant and string
interpolation.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -100,6 +100,10 @@ Vagrant.configure('2') do |config|
         'development' => ['default']
       }
 
+      if tags = ENV['ANSIBLE_TAGS']
+        ansible.tags = tags
+      end
+
       ansible.extra_vars = {'vagrant_version' => Vagrant::VERSION}
       if vars = ENV['ANSIBLE_VARS']
         extra_vars = Hash[vars.split(',').map { |pair| pair.split('=') }]
--- a/group_vars/all/helpers.yml
+++ b/group_vars/all/helpers.yml
@@ -4,8 +4,8 @@ wordpress_env_defaults:
   db_user: "{{ item.key | underscore }}"
   disable_wp_cron: true
   wp_env: "{{ env }}"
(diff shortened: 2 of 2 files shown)


runofthemill:
Continuing the conversation from #646 - Ansible provides easy access to system facts that can be used here and elsewhere: http://docs.ansible.com/ansible/playbooks_variables.html#information-discovered-from-systems-facts

I think it may be good to use `{{ ansible_fqdn }}` instead of a filter - I think it was in [this](https://www.ansible.com/ansible-roles-by-acquia) video where Jeff talks about the idea that Ansible roles/playbooks shouldn't attempt to use Ansible as a programming language (i.e. using lots of complex logic, filters, etc.) and as a whole Trellis could probably benefit from a reduction in complexity - this (IMO) is a good place to take advantage of that!

thisolivier:
@runofthemill Since ansible_fqdn relies on an external lookup, this could result in some very strange behaviour, see [here.](https://github.com/ansible/ansible/issues/9972)- I'm imagining if you're deploying the production version while the domain is still in use elsewhere. My apologies if that's not relevant here.

fullyint:
Suppose a user setting up a new DO droplet doesn't customize the hostname to be the fqdn, leaving some default value like this:
image
Then `hostname --fqdn` and `ansible_fqdn` are both `ubuntu-512mb-sfo1-01`. Trellis could add a task to set the hostname, but I'd be inclined to stick with `{{ site_hosts_canonical | first }}`.

runofthemill:
@thisolivier @fullyint I see your points - I was using it with vagrant in the context of the local dev server, which I mentioned in the other post (but not here); my only other thought is I'd just suggest leaving `wp_siteurl` as ` "${WP_HOME}/wp"` to keep things DRY.

PR #649, opened 2016-08-31
Don't run xdebug in the CLI

kalenjohnson:
This is especially important if running Composer via the command line in the vagrant box.

Should speed up anything running via the PHP CLI. 99.9% of the time xdebug does not need to be run with the CLI.


kalenjohnson:
I am noticing that on re-provisions this task still is set to 'changed', might need to check if xdebug is actually activated in the CLI before running this command

swalkinshaw:
Awesome 👍  Yeah making it idempotent would be better

kalenjohnson:
Idempotent

PR #650, opened 2016-08-31
backup feature

guilro:
This is a proposition for a backup feature inspired by @MWDelaney work. 

The role uses [Souts.backup](https://github.com/Stouts/Stouts.backup) to create two duply backup profiles for each `wordpress_sites` which has `backup: {enabled: true}`.

The configuration for each `wordpress_sites` is the following : 

``` yaml
backup:
    enabled: true # creates a backup profile if true (default: false)
    auto: true # set up a cron job for backup each night if true (default: false)
    target: scp://user@example.com/website_name # any valid duplicity location
```

The parameter `auto` allows to create backup profile without installing the cron job. This is usefull if you want to use the backup profile only to restore backups made from another server, by running `sudo duply website_name_database restore` and `sudo duply website_name_uploads restore` . This can be an interesting solution to synchronize database and uploads between production and staging for example.

### DIFF
--- a/group_vars/production/wordpress_sites.yml
+++ b/group_vars/production/wordpress_sites.yml
@@ -19,3 +19,9 @@ wordpress_sites:
       provider: letsencrypt
     cache:
       enabled: false
+    backup:
+      enabled: false # switch to true to enable backup
+      auto: true
+      target: scp://user@example.com/example.com_backups # any location supported by duplicity
+      schedule: '0 4 * * *' # change this value
+      purge: false # switch to true to enable automatic purging of old backups
--- a/group_vars/staging/wordpress_sites.yml
+++ b/group_vars/staging/wordpress_sites.yml
@@ -19,3 +19,12 @@ wordpress_sites:
       provider: letsencrypt
     cache:
       enabled: false
+    backup:
+      enabled: false
+      auto: false
+      target: scp://user@example.com/example.com_backups # any location supported by duplicity
+      schedule: '0 4 * * *' # change this value
+      purge: false # switch to true to enable automatic purging of old backups
+      # let "enabled: true" and "auto: false" to install backup profiles but not
+      # actually making backups, for example to restore your production database
+      # on staging
--- a/roles/backup/defaults/main.yml
+++ b/roles/backup/defaults/main.yml
@@ -1,0 +1,3 @@
+site_uses_backup: "{{ item.value.backup is defined and item.value.backup.enabled | default(false) }}"
+site_purge_backup: "{{ item.value.backup is defined and item.value.backup.purge | default(false) }}"
+site_purge_backup: "{{ item.value.backup is defined and item.value.backup.enabled and item.value.backup.auto and item.value.backup.purge | default(false) }}"
--- a/roles/backup/tasks/main.yml
+++ b/roles/backup/tasks/main.yml
@@ -1,0 +1,48 @@
+---
+- name: Ensure python pip and paramiko dependencies are installed
+  apt:
+    name:
+      - python-pip
+      - python-dev
+      - libffi-dev
+    state: present
+
+- name: Ensure paramiko is installed
+  pip:
+    name: paramiko
+    state: present
+
+- name: List backup jobs
+  set_fact:
+    backup_user: "{{ admin_user }}"
+    backup_group: sudo
+
+    backup_mysql_user: "{{ mysql_root_user }}"
+    backup_mysql_pass: "{{ mysql_root_password }}"
+
+    # Define the backup jobs
+    backup_profiles:
+    # Backup uploads
+      - name: "{{ item.key }}_uploads"
+        schedule: "{{ item.value.backup.schedule | default(omit) }}"
+        source: "{{ www_root }}/{{ item.key }}/shared/uploads"
+        target: "{{ item.value.backup.target }}/uploads"
+        target_user: "{{ site_env.backup_target_user }}"
+        target_pass: "{{ site_env.backup_target_pass }}"
+        action: "{{ site_purge_backup | ternary('purge_backup --force', 'backup') }}"
+
+    # Backup database
+      - name: "{{ item.key }}_database"
+        schedule: "{{ item.value.backup.schedule | default(omit) }}"
+        source: "mysql://{{ item.key | underscore }}_{{ env }}"          # Backup prefixes: postgresql://, maysql://, mongo://
+        target: "{{ item.value.backup.target }}/database"
+        target_user: "{{ site_env.backup_target_user }}"
+        target_pass: "{{ site_env.backup_target_pass }}"
+        action: "{{ site_purge_backup | ternary('purge_backup --force', 'backup') }}"
+
+  when: site_uses_backup
+  with_dict: "{{ wordpress_sites }}"
+  register: backup_jobs
+
+- name: Configure backup jobs
+  set_fact: backup_profiles="{{ backup_jobs.results | selectattr('ansible_facts', 'defined') | map(attribute='ansible_facts.backup_profiles.0') | list }} + {{ backup_jobs.results | selectattr('ansible_facts', 'defined') | map(attribute='ansible_facts.backup_profiles.1') | list }}"
--- a/group_vars/production/vault.yml
+++ b/group_vars/production/vault.yml
@@ -22,3 +22,6 @@ vault_wordpress_sites:
       secure_auth_salt: "generateme"
       logged_in_salt: "generateme"
       nonce_salt: "generateme"
+      # Uncomment if you use backups
+      # backup_target_user: example_user
+      # backup_target_pass: example_backup_password
--- a/group_vars/staging/vault.yml
+++ b/group_vars/staging/vault.yml
@@ -22,3 +22,6 @@ vault_wordpress_sites:
       secure_auth_salt: "generateme"
       logged_in_salt: "generateme"
       nonce_salt: "generateme"
+      # Uncomment if you use backups
+      # backup_target_user: example_user
+      # backup_target_pass: example_backup_password
--- a/requirements.yml
+++ b/requirements.yml
@@ -20,3 +20,7 @@
 - name: mailhog
   src: geerlingguy.mailhog
   version: 2.1.0
+
+- name: Stouts.backup
+  src: Stouts.backup
+  version: 3.4.5
--- a/server.yml
+++ b/server.yml
@@ -40,3 +40,5 @@
     - { role: wp-cli, tags: [wp-cli] }
     - { role: letsencrypt, tags: [letsencrypt], when: sites_using_letsencrypt | count }
     - { role: wordpress-setup, tags: [wordpress, wordpress-setup, letsencrypt] }
+    - { role: backup, tags: [backup] }
+    - { role: Stouts.backup, tags: [backup] }


swalkinshaw:
Thanks for this @guilro. I'm not opposed to adding a backup feature, but it will take us a bit to consider this and look it over.

inthedeepend:
This is great Guillaume, i've tried to use it and having some errors on provision. 
As soon as I remove 
`backup:
    enabled: true # creates a backup profile if true (default: false)
    auto: true # set up a cron job for backup each night if true (default: false)
    target: scp://user@example.com/website_name # any valid duplicity location`
from `wordpress_sites` the provision works fine... 

Any advice would be massively appreciated :)

guilro:
You can use `scheme://[user:password@]host[:port]/[/]path`, so you don't have to define `backup_target_user` and `backup_target_pass`.

I think we should not rely on this variables for the backup feature, as they are global to all the stouts.backup role and not only to one specific backup role. That would mean having the same backup host for all websites managed by trellis.

inthedeepend:
@guilro - makes perfect sense. 
I'm getting a different error on provision now... 

ERROR
```
The error appears to have been in '/Users/bradley/Sites/www.domainname.com/tr
ellis/roles/backup/tasks/main.yml': line 12, column 3, but may

- name: Create backup jobs
  ^ here

fatal: [crownruler.com]: FAILED! => {"failed": true}
```

This is what's in my wordpress_sites now : 
```
    backup:
      enabled: true # switch to true to enable backup
      auto: true
      target: s3://AWSACCESSKEYID:AWSSECRETACCESSKEY@s3-ap-southeast-2.amazonaws.com:443/bucket/directory
```

Your help would be most appreciated once again sir

inthedeepend:
Makes perfect sense @guilro but I'm having absolutely no luck getting this working with S3. 

I've tried every variation on the scheme I can think of and find and still get this error.

```
the field 'args' has an invalid value, which appears to include a variable
that is undefined. The error was: 'dict object' has no attribute 'target'
```

Have you tried this with backends that require auth?

MWDelaney:
That error looks more like a yml formatting issue than a configuration issue. Does it work with a simpler target like local?

inthedeepend:
It must be a yml formatting thing but I've copied everything detailed in the changes here...

Hmmm

inthedeepend:
Just can't get this to work full stop regardless of the target

guilro:
Ok, I will try to look at it next week :)

2016-11-18 11:18 GMT+01:00 Bradley <notifications@github.com>:

newtonne:
@inthedeepend @guilro I haven't tested this but it looks like the issue with the target is that:

`{{ item.value.target }}`

should be:

`{{ item.value.backup.target }}`

lucrussell:
Just to confirm, I had the same issue and tried `{{ item.value.backup.target }}` as suggested by @newtonne, this corrected the problem. 

I then received this error: 

    the field 'args' has an invalid value, which appears to include a variable
that is undefined. The error was: 'mysql_root_user' is undefined

I was able to resolve this by specifying the `mysql_root_user` variable in production/main.yaml. 

Not sure if this is the best approach, but it's working. Thanks all for the great work on this backup solution!

greatislander:
@lucrussell Haven't tested myself but are you running the playbook with just the backup role? `mysql_root_user` is defined in  the MariaDB role so if you have that role commented out it will be undefined as you are reporting. See https://github.com/roots/trellis/blob/b299b331f2112c2cb5ee7fa2e7512a9e5a663821/roles/mariadb/defaults/main.yml#L2

guilro:
Hello everyone. I just fixed the `{{ item.value.backup.target }}`, and now it is working.

greatislander:
One piece of feedback on this, @guilro: wouldn't it be better practice to put the credentials in `/group_vars/<env>/vault.yml` rather than inlining them as you suggest [here](https://github.com/roots/trellis/pull/650#issuecomment-261187432)?

guilro:
Good suggestion. I thought it was impossible because `backup_target_pass` was a setting global to Stouts.Backup. But you can also set up `target_pass` per backup profile. Should be working now.

Several caveats :
* with ssh backends, backup won't work out of the box because the target is not in `.known_hosts`. You first have to manually SSH from the server into your backup target,and validate the target private key.
* the cron time is hard-coded. It is not a good practice that every user backups at the same time. Working on that.

greatislander:
Looks good. Another thought—have you considered integrating a `purge` command as per @MWDelaney's role so that the backup server doesn't keep filling up over time?

guilro:
I agree such a feature would be usefull, but I don't think its a good idea to purge by default. So this would make one more setting.

group_vars/production/wordpress_sites.yml
```diff
@@ -19,3 +19,8 @@ wordpress_sites:
       provider: letsencrypt
     cache:
       enabled: false
+    backup:
+      enabled: false # switch to true to enable backup
+      auto: true
+      target: scp://user@example.com/example.com_backups # any location supported by duplicity
```
greatislander: Add to backup dict:

```
purge: false # switch to true to enable automatic purging of old backups
```

group_vars/staging/wordpress_sites.yml
```diff
@@ -19,3 +19,11 @@ wordpress_sites:
       provider: letsencrypt
     cache:
       enabled: false
+    backup:
+      enabled: false
+      auto: false
+      target: scp://user@example.com/example.com_backups # any location supported by duplicity
```
greatislander: Add to backup dict:

```
purge: false # switch to true to enable automatic purging of old backups
```

roles/backup/defaults/main.yml
```diff
@@ -0,0 +1,2 @@
+site_uses_backup: "{{ item.value.backup is defined and item.value.backup.enabled | default(false) }}"
+site_auto_backup: "{{ item.value.backup is defined and item.value.backup.auto | default(true) }}"
```
greatislander: Add:

```
site_purge_backup: "{{ item.value.backup is defined and item.value.backup.purge | default(false) }}"
```

roles/backup/tasks/main.yml
```diff
@@ -0,0 +1,43 @@
...
+        target: "{{ item.value.backup.target }}/database"
+        target_user: "{{ site_env.backup_target_user }}"
+        target_pass: "{{ site_env.backup_target_pass }}"
+
+  when: site_uses_backup
+  with_dict: "{{ wordpress_sites }}"
+  register: backup_jobs
+
+- name: Configure backup jobs
+  set_fact: backup_profiles="{{ backup_jobs.results | selectattr('ansible_facts', 'defined') | map(attribute='ansible_facts.backup_profiles.0') | list }} + {{ backup_jobs.results | selectattr('ansible_facts', 'defined') | map(attribute='ansible_facts.backup_profiles.1') | list }}"
```
greatislander: Add after the `Configure` task:

```
- name: Create uploads purge job
  cron: name="Purge old uploads backups" minute="0" hour="1"
        cron_file=purge
        user="{{ admin_user }}" job="duply /etc/duply/{{ item.key }}_uploads purge --force"
  when:
    - site_uses_backup
    - site_purge_backup

- name: Create database purge job
  cron: name="Purge old database backups" minute="0" hour="2"
        cron_file=purge
        user="{{ admin_user }}" job="duply /etc/duply/{{ item.key }}_database purge --force"
  when:
    - site_uses_backup
    - site_purge_backup
```

retlehs:
purge seems important to have on by default, similar to how we [rotate log files](https://github.com/roots/trellis/blob/cbe647a26a95126a3cf7fd8ace7f1f2f0454adf0/group_vars/all/logrotate.yml#L9)

MWDelaney:
I had a server eat itself recently because I didn't have purge working correctly and it used up all the free space. I think purge is important. Maybe daily backups until there's a werk's worth, and then weekly until there's a month, and then monthly for... however long? Purging as you go after each checkpoint is reached?

guilro:
@MWDelaney  Duply makes incremental backups, and only few full backups. So there is no gain at using grand-father/father/son model. As long as you have only one full backup, there is no duplicate data. Skipping a day only mean either having a bigger diff next day or creating a new full backup. This create "chains" of backups.
Purge use a combination of MAX_FULL_BACKUPS and MAX_AGE to delete old backup "chains".

My point was that backups are not made locally so it should not be critical if your server fills up (you should have a dedicated server for backups), and you may want to keep your backup longer (that is the point of backups), but I am ok to work on it.

guilro:
@greatislander just saw yours ! Actually I did exactly the same :)

newtonne:
@guilro @greatislander Instead of a separate cron for the purge, you could alternatively integrate the purge into the backup cron by adding the following to `backup_profiles` under the `target_pass` definition:

> action: "{{ 'purge_backup --force' if site_purge_backup else 'backup' }}"

I have tested this and it works as expected. Note that the underscore between 'purge' and 'backup' is a neutral separator that tells duply to run a purge and then a backup. It is possible to tell it to only do the backup if the purge succeeds or only if it fails. See the section on separators in the [duply docs](http://duply.net/wiki/index.php/Duply-documentation).

guilro:
Suggestion added ! I also turned the `purge` setting to `true` by default.

newtonne:
One issue I did run into is that it fails when you do a manual restore logged in as the `admin_user`. Running something like:

`/usr/bin/duply /etc/duply/mysite_uploads restore`

will fail because the `admin_user` does not have permission to modify any of the files/directories under `/srv/www`, since they are owned by the `web_user`.

Changing the `backup_user` to `web_user` does resolve this.

MWDelaney:
I know @swalkinshaw said they'd be willing to look at this as a feature, but I wonder if it would make sense to maintain the backup playbook as a sort of "recommended" or semi-official add-on to the Trellis core project. 

There are other projects that could fit in such a category --playbooks that are widely useful but not strictly within Trellis' project scope. I'm thinking here of Ben's database push/pull scripts, and media push/pull scripts I've seen around. There might be an opportunity to "make official" (and focus development and support time on) a list of add-ons for Trellis, the way Gravity Forms and WooCommerce officially promote 3rd party add-ons.

Just a thought.

jasperf:
Looking for a great backup solution I finally bumped into this pull request. Seems like an awesome addition. Hope we can merge it soon. @MWDelaney 's suggestion for a category playbooks on Roots would be a great opportunity too. And a way to keep Trellis core clean while at the same time allowing playbooks to be displayed and supported at the same time.

nathobson:
@guilro Thanks for the work on this. I've had a good amount of time to test this now. I've successfully got backups running smoothly to [Backblaze B2](https://www.backblaze.com/b2/cloud-storage.html) (nice that Duplicity supports this). 

Two questions:

- How does purging work exactly? Is there some sort of max_age set?
- How do restores work? I'm probably missing the obvious but haven't figured that one out.

Again, thanks for the time you've put into everything so far.

nathobson:
Scratch that, I've looked more into restores. However, facing an issue here. It seem (on my instance at least) that the `suplicity-full-signatures` for the example.com_database and example.com_uploads are the wrong way around, meaning the restores fail.

For example, my uploads backup has `duplicity-full-signatures.20170310T160306Z.sigtar.gz`and when I run `duply example.com_uploads restore 1D` I get:

`--- Start running command RESTORE at 17:48:23.131 ---
Synchronizing remote metadata to local cache...
Copying duplicity-full-signatures.20170310T160307Z.sigtar.gz to local cache.
Attempt 1 failed. HTTPError: HTTP Error 404: Not Found`

However, it's my database that has `duplicity-full-signatures.20170310T160307Z.sigtar.gz`. Equally, my database restore fails in the same way in reverse.

swalkinshaw:
This will likely be split out into its own role (via Galaxy I guess).

It's useful to have development happen here though for more visibility.

@guilro I'm going to go through and review this. After we can figure out some logistics for integrations. Sound good?

roles/backup/tasks/main.yml
```diff
@@ -0,0 +1,45 @@
+---
+- name: Ensure python pip and paramiko dependencies are installed
+  apt:
+    name:
+      - python-pip
+      - python-dev
+      - libffi-dev
+    state: present
+- name: Ensure paramiko is installed
```
swalkinshaw: Blank above above this

roles/backup/tasks/main.yml
```diff
@@ -0,0 +1,45 @@
...
+  pip: name=paramiko state=present
```
swalkinshaw: Use the YAML format here

```yaml
pip:
  name: paramiko
  state: present
```

roles/backup/tasks/main.yml
```diff
@@ -0,0 +1,45 @@
...
+    # Define the backup jobs
+    backup_profiles:
+    # Backup uploads
+      - name: "{{ item.key }}_uploads"
+        schedule: "{{ item.value.backup.schedule if site_auto_backup else '0 3 31 2 *' }}" # On February 31 if not auto
+        source: "{{ www_root }}/{{ item.key }}/shared/uploads"
+        target: "{{ item.value.backup.target }}/uploads"
+        target_user: "{{ site_env.backup_target_user }}"
+        target_pass: "{{ site_env.backup_target_pass }}"
+        action: "{{ 'purge_backup --force' if site_purge_backup else 'backup' }}"
```
swalkinshaw:
```
"{{ site_purge_backup | ternary('purge_backup --force', 'backup') }}"
```

roles/backup/tasks/main.yml
```diff
@@ -0,0 +1,45 @@
...
+
+    # Backup database
+      - name: "{{ item.key }}_database"
+        schedule: "{{ item.value.backup.schedule if site_auto_backup else '0 4 31 2 *' }}" # On February 31 if not auto
+        source: "mysql://{{ item.key | underscore }}_{{ env }}"          # Backup prefixes: postgresql://, maysql://, mongo://
+        target: "{{ item.value.backup.target }}/database"
+        target_user: "{{ site_env.backup_target_user }}"
+        target_pass: "{{ site_env.backup_target_pass }}"
+        action: "{{ 'purge_backup --force' if site_purge_backup else 'backup' }}"
```
swalkinshaw:
```
"{{ site_purge_backup | ternary('purge_backup --force', 'backup') }}"
```

roles/backup/tasks/main.yml
```diff
@@ -0,0 +1,45 @@
...
+    backup_group: sudo
+
+    backup_mysql_user: "{{ mysql_root_user }}"
+    backup_mysql_pass: "{{ mysql_root_password }}"
+
+    # Define the backup jobs
+    backup_profiles:
+    # Backup uploads
+      - name: "{{ item.key }}_uploads"
+        schedule: "{{ item.value.backup.schedule if site_auto_backup else '0 3 31 2 *' }}" # On February 31 if not auto
```
swalkinshaw: There ideally shouldn't be a default schedule here. Can this be omitted?

```
schedule: "{{ item.value.backup.schedule | default(omit) }}"
```

`schedule` will be skipped when `item.value.backup.schedule` is not defined

roles/backup/tasks/main.yml
```diff
@@ -0,0 +1,45 @@
...
+        source: "{{ www_root }}/{{ item.key }}/shared/uploads"
+        target: "{{ item.value.backup.target }}/uploads"
+        target_user: "{{ site_env.backup_target_user }}"
+        target_pass: "{{ site_env.backup_target_pass }}"
+        action: "{{ 'purge_backup --force' if site_purge_backup else 'backup' }}"
+
+    # Backup database
+      - name: "{{ item.key }}_database"
+        schedule: "{{ item.value.backup.schedule if site_auto_backup else '0 4 31 2 *' }}" # On February 31 if not auto
```
swalkinshaw: Omit this too is possible

nathobson:
Please do let me know if any testing is required. I've now had a good chance to experiment with several of the targets/backends including B2, S3 and SFTP... with varying success.

medfreeman:
:+1:  for @MWDelaney 's suggestion for a category playbooks.
I'd even add a 'recipes' category for the kinds of additions which don't belong in a playbook.
i.e.: setup-health checks on amazon elb, redirect to https behind an elb, etc..

MWDelaney:
Roots has an aesthetic theme to its naming, so additional playbooks could be called "seeds", and "how-to" recipe type info could be called "guides".

inthedeepend:
This thread is getting exciting! Can't wait to see this pan out.

guilro:
@swalkinshaw thanks for your review and sorry for the delay.

Tell me what is your decision for integration, I can create a separated role if you ask for.

swalkinshaw:
@guilro separate role to start with at least. Thanks for all the work on this 🎉

guilro:
Does [this](https://galaxy.ansible.com/guilro/trellis-backup/) seem correct ?

partounian:
Sorry to bother you, can you take a moment to review @guilro's updates? @swalkinshaw

swalkinshaw:
@guilro looks good to me 👍 Role ended up being pretty simple in the end which is good.

I'd encourage you to add to our https://github.com/roots/docs repo. Maybe a new page on Backups?

PR #656, opened 2016-09-09
Fix errors when compiling wheel

kalenjohnson:
Yay, another Windows problem.

So I was having issues on a new project. pip when installing Ansible has a couple dependencies, and wheel was failing to compile. Adding these packages looks like it solves the issue.


swalkinshaw:
👏

PR #657, opened 2016-09-13
Statics cache max age

pacotole:
Added a rule for nginx wordpres.conf template to serve static files with the max age for cache expiration.


swalkinshaw:
I'm not sure we can just add this rule as a default. People get might cached files without realizing it if they don't have any cache invalidation mechanism.

If we did add this, we'd probably use the [h5bp config](https://github.com/h5bp/server-configs-nginx/blob/master/h5bp/location/expires.conf) since we already use the others.

retlehs:
it would be pretty dope if we could have an option to enable h5bp's expires conf for each wordpress site. sage based themes for example wouldn't have to worry about cache invalidation

at the least we can add some documentation

swalkinshaw:
@pacotole thanks for this but I'm going to close this for the time being due to the reasons mentioned above until you or we can come up with a better solution.

This is actually currently possible to add on a per-site basis via [Nginx includes](https://roots.io/trellis/docs/nginx-includes/). We do this on https://roots.io.

Add a file like  `roles/wordpress-setup/templates/includes.d/roots.io/assets-expiry.conf.j2`:

```nginx
# Media: images, icons, video, audio, HTC
location ~* \.(?:jpg|jpeg|gif|png|ico|cur|gz|svg|svgz|mp4|ogg|ogv|webm|htc)$ {
  expires 1M;
  access_log off;
  add_header Cache-Control "public";
  gzip_vary on;
}

# CSS and Javascript
location ~* \.(?:css|js)$ {
  expires 1y;
  access_log off;
  add_header Cache-Control "public";
  gzip_vary on;
}
```

PR #658, opened 2016-09-14
Update WP's timezone_string

regularjack:
Update WP's `timezone_string` with the value from the `default_timezone` configuration.

Note:
I can't for the sake of me figure out how to escape single quotes inside the `changed_when` so I simplified it. Ideally, it would be: `changed_when: "'Success: Updated ''timezone_string'' option.' in wp_timezone_update_results.stdout"`

I can change the PR when I get feedback on the above note.

### DIFF
--- a/roles/wordpress-install/tasks/main.yml
+++ b/roles/wordpress-install/tasks/main.yml
@@ -63,6 +63,15 @@
   with_dict: "{{ wordpress_sites }}"
   when: wp_install_results | changed or wp_multisite_install_results | changed
 
+- name: Setup WP Timezone
+  command: wp option update timezone_string {{ default_timezone }} --allow-root
+  args:
+    chdir: "{{ www_root }}/{{ item.key }}/{{ item.value.current_path | default('current') }}/"
+  register: wp_timezone_update_results
+  with_dict: "{{ wordpress_sites }}"
+  changed_when: "'Success: Updated' in wp_timezone_update_results.stdout"
+  when: wp_install_results | changed
+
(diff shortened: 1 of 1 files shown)


retlehs:
thanks for the PR! we discussed this briefly the other day — consensus was that the best practice is to keep the server at UTC and to modify the timezone in the WP admin for what you'd like to display

there's some good discussion here as well: http://stackoverflow.com/questions/2532729/daylight-saving-time-and-time-zone-best-practices

regularjack:
I don't understand: if the best practice is to keep the server in UTC, why does the `default_timezone` configuration exist? See #301

retlehs:
i don't have an answer for you, but we can revisit that and possibly remove that code...

swalkinshaw:
Does WP not read the default PHP timezone setting as a default?

regularjack:
@swalkinshaw The `timezone_string` options is by default `NULL`, which means it falls back to UTC.

swalkinshaw:
@regularjack thanks for this contribution but I'm going to pass on this.

There's a lot of potential WP options we could set on install and I don't want to set a precedent about which to include or not. I know we already have permalinks but that doesn't mean we should keep adding more. So I'm kind of arbitrarily drawing the line here :(

Maybe you/anyone else could just a custom post install role to run further customizations.

Re: https://github.com/roots/trellis/pull/301. I'd actually consider removing it, but it's pretty hard to do. Just removing the code would change anyone's PHP settings back to the default without warning which is not nice.

PR #665, opened 2016-10-01
Added additional XDebug variables

nathanielks:
I've went ahead and added additional, "nice to have" variables to our XDebug configuration in order to make working with XDebug a bit easier out of the box. This isn't _all_ the variables XDebug offers, but it is most of them and they all are quite nice to use. Whenever I'm profiling or tracing, I end up modifying most of these, so doing this via ansible instead of manually doing it would be great.


nathanielks:
Oh, they've been added with their default values in XDebug, so no behavior has been changed.

nathanielks:
Though now that I think of it, I've got a few more changes to make. WIP

nathanielks:
Okie doke, I've also added a script that will create an ssh tunnel to a configured remote environment. The way it's used is this:

```
bin/xdebug-tunnel.sh production some_host_prod
```

`some_host_prod` corresponds to a defined host in a hosts file, like so:

```
# let's pretend hosts/production

some_host_prod ansible_ssh_host=12.34.56.78

[production]
some_host_prod

[web]
some_host_prod
```

It will create a tunnel using a control socket which enables the tunnel to persist as long as the socket is open. The socket is created at `/tmp/trellis-xdebug-{{ host }}`, in this case it would be `/tmp/trellis-xdebug-some_host_prod`. The tunnel routes local port 9000 to remote port 9000 on the host.

To close the connection, run
```
bin/xdebug-tunnel.sh production some_host_prod close
```

nathanielks:
@swalkinshaw @fullyint any thoughts here?

nathanielks:
Notes from a conversation with @fullyint 
- [ ] Replace instances of `ansible_ssh_*` with `ansible_*`
- [ ] Remove reference to `xdebug_tunnel`
- [ ] Remove `env` requirement, as well as the `variable-check.yml` playbook
- [ ] Simplify to new xdebug playbook:
      
  1. define these vars github.com/roots/trellis/blob/master/group_vars/development/php.yml#L8-L15 somewhere NOT in `group_vars/development`
  2. run `server.yml --tags php`
  3. run `bin/xdebug-tunnel.sh production some_host_prod`
  4. play
  5. run `bin/xdebug-tunnel.sh production some_host_prod close`
  6. change back to `php_xdebug_remote_enable: false`
  7. run `server.yml --tags php`

  - If so, consider this attempt to eliminate steps 1,2,6,7:
    - A new xdebug role includes the vars from step 1 in a `roles/xdebug/defaults/main.yml` file
    - The `xdebug-tunnel.yml` playbook runs the new xdebug role, thus handling steps 2 and 7
    - Running the playbook implicitly sets `php_xdebug_remote_enable: true` and running it with the `close` flag implicitly sets the var back to `php_xdebug_remote_enable: false` (handling step 6

nathanielks:
Separating this into different PR's. First: https://github.com/roots/trellis/pull/676

PR #670, opened 2016-10-10
add $http_origin to fastcgi_cache_key

olapersson:
Wordpress API adds the origin to CORS headers, if requesting from two different origins while the cache is hot headers will be incorrect unless http_origin is added to the cache key.

Had to add this to a server where I have more than one possible origin pulling content from the wordpress api, causing CORS errors. I think it makes sense having this as the default behaviour.


swalkinshaw:
Makes sense. `$http_origin` should just be empty on non-CORS requests so cache key is still good in that case too.

Thanks 👍

PR #683, opened 2016-11-07
Use to_text to prevent to_unicode deprecation warning

fullyint:
Ansible 2.2+ prints this warning:
```
[DEPRECATION WARNING]: ansible.utils.unicode.to_unicode is deprecated. 
Use ansible.module_utils._text.to_text instead.
This feature will be removed in version 2.4.
```

This PR prevents the deprecation warning by importing and using [`to_text`](https://github.com/ansible/ansible/blob/v2.2.0.0-1/lib/ansible/module_utils/_text.py#L135) when available (Ansible 2.2+) and otherwise falling back to [`to_unicode`](https://github.com/ansible/ansible/blob/v2.1.1.0-1/lib/ansible/utils/unicode.py#L43). The fallback will no longer be needed once Trellis requires Ansible >= 2.2.

### DIFF
--- a/lib/trellis/utils/output.py
+++ b/lib/trellis/utils/output.py
@@ -8,7 +8,12 @@ import re
 import textwrap
 
 from ansible import __version__
-from ansible.utils.unicode import to_unicode
+
+# to_unicode will no longer be needed once Trellis requires Ansible >= 2.2
+try:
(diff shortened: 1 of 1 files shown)


swalkinshaw:
Great idea with the import alias 👍

PR #685, opened 2016-11-08
Disable wp-cron emails

retlehs:
see https://discourse.roots.io/t/wp-cron-sending-emails-like-crazy/5276

this was previously fixed in #421 but we changed how wp-cron worked with f07bf72bcd5d7107c6f1aeeb042eb8b06cf8cee3 and forgot about this


swalkinshaw:
pr accepted

PR #688, opened 2016-11-15
Modularize Nginx wordpress-site.conf.j2 and add Nginx conf hooks

fullyint:
Trellis Nginx includes have been limited to one location, inside the primary `server` block. This PR adds hooks to enable users to include files or inject snippets in other locations (e.g., outside `server` block). Here is [an example](https://discourse.roots.io/t/nginx-error-server-name-hash-bucket-size-exceeded/7905/3) of needing to inject a directive in the `http` context.

The docs in [roots/docs#57](https://github.com/roots/docs/pull/57) will be most helpful for understanding the intent of this PR. Builds on [#687](https://github.com/roots/trellis/pull/687)

This PR is simpler than it looks. The first commit accounts for most all the PR's additions/deletions, and doesn't change code. It just breaks the WordPress site conf out into separate modules/files. The second commit (Nginx hooks) adds only 13 lines.

The modularization of the Nginx confs is for convenience, to make the sections clear amongst all the jinja templating complexity. This is the approach already in use for `roles/wordpress-setup/templates/https.conf.j2`.

The one change made in the modularization is that this PR moves the base WordPress conf out of the `nginx` role into the `wordpress-setup` role. This changes this base WP conf from an `include` whose content is general to all sites, to content injected into each site's conf such that it can be customized per site, via the `nginx_location` hook.

### DIFF
--- a/roles/nginx/defaults/main.yml
+++ b/roles/nginx/defaults/main.yml
@@ -19,3 +19,6 @@ nginx_cache_size: 250m
 nginx_cache_inactive: 1h
 nginx_skip_cache_uri: /wp-admin/|/xmlrpc.php|wp-.*.php|/feed/|index.php|sitemap(_index)?.xml
 nginx_skip_cache_cookie: comment_author|wordpress_[a-f0-9]+|wp-postpass|wordpress_no_cache|wordpress_logged_in
+
+# Nginx hooks
+nginx_http: ''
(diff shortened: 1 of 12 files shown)


fullyint:
I rebased to resolve merge conflicts that arose from the recently merged #692. Both PRs involve(d) caching stuff in Nginx confs.

I used the rebase as an opportunity to add a few `tags: nginx-includes` on tasks involving the new nginx hooks.

nathanielks:
@phil was about to create  a PR adding some support for trellis living behind a server that does SSL termination, but I remembered this existed and was curious where we're at with this getting merged?

nathanielks:
Nevermind! I was able to do it pretty simply.

fullyint:
closed and replaced by the better solution in #740

PR #691, opened 2016-11-29
Add Percona as a database server option

dlundgren:
This allows Percona to be chosen as an alternative to MariaDB.


QWp6t:
Have you tested provisioning with mariadb and then provisioning the same server with percona, and vice versa? Should one be completely uninstalled before installing the other?

dlundgren:
I have not tested that particular scenario. MariaDB & Percona are mutually exclusive and replace/break each other.

I should be able to take a look at having it detect which is installed, and handling that scenario, later this week.

swalkinshaw:
These changes seem pretty minor so I'd be open to adding this. Can the duplication in the provider tasks be minimized? Maybe just include a task file for provider install which would only have the first ~4 tasks to actually install it.

dlundgren:
I moved the block into the main task, and left the installers in their own files.

I also noticed an issue with which release I was installing from when using `percona_dist`, I dropped that in favor of using `ansible_distribution_release`.

roles/mysql-server/defaults/main.yml
```diff
@@ -1,8 +1,15 @@
-mariadb_binary_logging_disabled: true
+# change this to percona if you want to use percona
```
swalkinshaw: Maybe: `# providers: mariadb or percona`

dev.yml
```diff
@@ -1,16 +1,15 @@
 ---
-- name: "WordPress Server: Install LEMP Stack with PHP 7.0 and MariaDB MySQL"
+- name: "WordPress Server: Install LEMP Stack with PHP 7.0 and chosen MySQL"
```
swalkinshaw: While you're here, let's just simplify this to: "WordPress Server: Install LEMP stack"

server.yml
```diff
@@ -17,7 +17,7 @@
     - name: Install Python 2.x
       raw: sudo apt-get install -qq -y python-simplejson
 
-- name: WordPress Server - Install LEMP Stack with PHP 7.0 and MariaDB MySQL
+- name: "WordPress Server - Install LEMP Stack with PHP 7.0 and chosen MySQL"
```
swalkinshaw: Same here

roles/mysql-server/tasks/percona.yml
```diff
@@ -0,0 +1,24 @@
+---
+- name: Add Percona MySQL apt-key
+  apt_key:
+    url: "http://keyserver.ubuntu.com/pks/lookup?op=get&fingerprint=on&search={{ percona_keyserver_fingerprint }}"
```
swalkinshaw: Let's use `https` here

roles/mysql-server/tasks/mariadb.yml
```diff
@@ -0,0 +1,24 @@
+---
+- name: Add MariaDB MySQL apt-key
+  apt_key:
+    url: "http://keyserver.ubuntu.com/pks/lookup?op=get&fingerprint=on&search={{ mariadb_keyserver_fingerprint }}"
```
swalkinshaw: And `https` here as well :)

roles/mysql-server/defaults/main.yml
```diff
@@ -1,8 +1,15 @@
...
+
 mariadb_keyserver_fingerprint: "0xcbcb082a1bb943db"
 mariadb_mirror: nyc2.mirrors.digitalocean.com
 mariadb_version: "10.0"
 mariadb_dist: trusty
+
+percona_keyserver_fingerprint: "0x9334A25F8507EFA5"
+percona_version: "5.7"
+
+binary_logging_disabled: true
```
swalkinshaw: `mysql_binary_logging_disabled`

swalkinshaw:
I'll test this out but otherwise it should be good to go 👍

swalkinshaw:
@dlundgren I'm sorry about this but I've changed my mind about including this in Trellis :( Mostly I don't want to incur the overhead of supporting both MariaDB and Percona. It might not be much, but it's obviously more than just MariaDB alone and increases the support/maintenance burden. This is also the first request of Percona we've had in ~2 years and anyone who needs it can easily add it themselves.

I also hadn't thought about what would happen once we switched to the official Ubuntu distro packages for Maria in https://github.com/roots/trellis/pull/693. Now that the custom apt repo is gone, these two database options diverge just a little more.

Thanks for the good work on this and sorry for not deciding this sooner before I got you to do additional work.

dlundgren:
That's fine, would you consider the option to allow mysql vs maria using distro packages?

Issue #436, opened 2015-12-15
Requests for missing .php files result in incorrect 404


retlehs:
`http://example.com/doesnt-exist.php` returns "File not found" instead of the WordPress 404 

`http://example.com/doesnt-exist.html` is good
`http://example.com/doesnt-exist` is good

pacotole:
I have the same problem again with the last versions of Trellis.

retlehs:
@pacotole unable to reproduce. please use https://discourse.roots.io/ if you still need support.

![image](https://cloud.githubusercontent.com/assets/115911/20719792/3cb04f10-b61a-11e6-9059-e7e1f250eb2c.png)

pacotole:
@retlehs yes, in vagrant works as expected but in server (staging or production) somefile.php returns "File not found" instead of the WordPress 404.

I'll try on a server with all the WP plugins disabled to discard that possibility...

fullyint:
@pacotole would love to have you test the proposed fix in #692 if you're interested

pacotole:
@fullyint Problem solved!! Thanks =)

PR #692, opened 2016-12-01
Fix 404s by moving skip_cache conditions to server block

fullyint:
**Problem**
Fixes a version of #436 where a system default 404 page is returned instead of the WordPress 404 (via index.html) . To reproduce, provision with cache [`enabled: true`](https://github.com/roots/trellis/blob/78400a99e953c6db3c628adde8b866e4fe13ece4/group_vars/production/wordpress_sites.yml#L21) and do any of the following:

- visit `example.com/missing.php?p=1` (triggers the `$query_string` condition in code block below)
- visit `example.com/xmlrpc.php` (triggers [`nginx_skip_cache_uri`](https://github.com/roots/trellis/blob/78400a99e953c6db3c628adde8b866e4fe13ece4/roles/nginx/defaults/main.yml#L20) condition in code block below)
- log in to `wp-admin` and then visit `example.com/missing.php` (triggers [`nginx_skip_cache_cookie`](https://github.com/roots/trellis/blob/78400a99e953c6db3c628adde8b866e4fe13ece4/roles/nginx/defaults/main.yml#L21) in code block below)

**Solution**
The PR diff display overcomplicates the changes. This PR simply moves the following portion of the caching section up out of the location block context and into the server block context.
```
set $skip_cache 0;

if ($query_string != "") {
  set $skip_cache 1;
}

# Don't cache uris containing the following segments
if ($request_uri ~* "{{ item.value.cache.skip_cache_uri | default(nginx_skip_cache_uri) }}") {
  set $skip_cache 1;
}

# Don't use the cache if cookies includes the following
if ($http_cookie ~* "{{ item.value.cache.skip_cache_cookie | default(nginx_skip_cache_cookie) }}") {
  set $skip_cache 1;
}
```

**Explanation**
Why does it fix the problem to move these `if` conditions out of the `location ~ \.php$` block? Because [If Is Evil](https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/), particularly when inside a location block. Here is a summary of the affected Trellis location block:

```
location ~ \.php$ {
  try_files $uri =404;
  error_page 404 /index.php;

  ... if conditions in code block above ...
}
```

The excerpt above appears to exhibit the problem demonstrated in the [example](https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/#examples) titled `# try_files wont work due to if`. If one of the `if` conditions is met, the `try_files` directive won't even be run. In our case, this means the WP `index.php` does not load and the system 404 is used instead. I am guessing this is related to the various descriptions [here](http://agentzh.blogspot.com/2011/03/how-nginx-location-if-works.html) about how 'Nginx traps into the "if" inner block because its condition ... was met.'

**Justification**
Here are two examples recommending placing the `if` condition cache exceptions in the server block instead of the location block: [easyengine.io](https://easyengine.io/wordpress-nginx/tutorials/single-site/fastcgi-cache-with-purging/) and [digitalocean.com](https://www.digitalocean.com/community/tutorials/how-to-setup-fastcgi-caching-with-nginx-on-your-vps), the latter even suggesting that these exceptions "must be used in the **server{ }** context."

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Fix 404s by moving skip_cache conditions to server block ([#692](https://github.com/roots/trellis/pull/692))
 * Nginx includes: Move templates dir, fix 'No such file' error ([#687](https://github.com/roots/trellis/pull/687))
 * [BREAKING] Move shell scripts to bin/ directory ([#680](https://github.com/roots/trellis/pull/680))
 * Add myhostname to nsswitch.conf to ensure resolvable hostname ([#686](https://github.com/roots/trellis/pull/686))
--- a/roles/wordpress-setup/templates/wordpress-site.conf.j2
+++ b/roles/wordpress-setup/templates/wordpress-site.conf.j2
@@ -40,30 +40,31 @@ server {
     include acme-challenge-location.conf;
   {% endif %}
 
+  {% if item.value.cache is defined and item.value.cache.enabled | default(false) -%}
+    set $skip_cache 0;
+
+    if ($query_string != "") {
+      set $skip_cache 1;
+    }
+
+    # Don't cache uris containing the following segments
+    if ($request_uri ~* "{{ item.value.cache.skip_cache_uri | default(nginx_skip_cache_uri) }}") {
+      set $skip_cache 1;
+    }
+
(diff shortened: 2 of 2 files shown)


swalkinshaw:
Good find :shipit:

PR #693, opened 2016-12-02
Switch to distro packages for MariaDB

RiFi2k:
Use official Ubuntu managed packages for simplicity and less dependencies to deal with.

#659


swalkinshaw:
This might be one of the best Trellis PRs yet 😄. Removes a bunch of code and improves things!

PR #694, opened 2016-12-02
Composer install: Use composer module and run as non-root

fullyint:
Prevents the cause of the `Do not run Composer as root/super user!` message when provisioning the Vagrant VM (example [thread](https://discourse.roots.io/t/ssh-fowarding-fails-during-provisioning-works-on-vagrant-ssh/7938)).


swalkinshaw:
👏

PR #695, opened 2016-12-02
PHP 7.1 update

kalenjohnson:
Tested on fresh `vagrant up` as well as provisioning from master then 7.1 on top for some backwards compatibility testing.

Provisioning over top of a PHP 7.0 box appears to run fine, but the `php7.0-fpm` continues to run, and the 7.1 service cannot start until the 7.0 is stopped. I'm not sure why Ansible did not throw an error.

This is also probably on hold until WP 4.7, as there is a big ugly warning thrown on first provision with the 2015 theme: https://core.trac.wordpress.org/ticket/37772


eshimischi:
Changed trellis  yml files to install 7.1, but php7.1-fpm can't be restarted at the end of deploying to digitalocean and stucked for a long time :(

PR #701, opened 2016-12-05
Add `main` format to wordpress-sites access_log

hooverlunch:
Seems this is an error. Just took me half an hour to figure out why the `log_format` in the main nginx.conf file was not actually being used.

Thanks for all you do!


swalkinshaw:
Good catch. Makes sense since we have the `log_format` defined 👍

Issue #561, opened 2016-04-12
More secure sshd defaults


swalkinshaw:
## Submit a feature request or bug report
- [x] This is a feature request
- [x] This request isn't a duplicate of an [existing issue](https://github.com/roots/trellis/issues)

---

**What is the current behavior?**

`sshd` may result in insecure-defaults like using RC4 cypher.

**What is the expected or desired behavior?**

Just like with SSL we should have better more secure defaults for sshd.
## Feature Request

**Please provide use cases for changing the current behavior:**

Security!

**Other relevant information:**

See https://discourse.roots.io/t/do-we-need-ssh-rc4-cipher-enabled/6438/6

aried3r:
I'm very interested in this, any idea how is this going to be tackled? Using something like [ansible-ssh-hardening](https://github.com/dev-sec/ansible-ssh-hardening) or doing your own?

swalkinshaw:
@aried3r that role looks decent from a quick glance. We could use something like that or just add our own changes to the sshd config.

We're actually already using another role that we imported: https://github.com/roots/trellis/tree/master/roles/sshd

So we could switch or just update that template. If you're interested in helping let us know 👍

aried3r:
I'm interested. :)

Personally, I'd use ansible-ssh-hardening because it is maintained so changes there don't have to be reflected and implemented in your own config, especially with vulnerabilities showing up.

What do you think?

swalkinshaw:
@aried3r sorry never replied to this. https://github.com/dev-sec/ansible-ssh-hardening looks good. Would you be interested in implementing it?

isynergy-development:
@swalkinshaw Is this still something you want to see implemented? I never saw ansible-ssh-hardening before and now I'm most likely going to integrate it with my setup.

swalkinshaw:
@isynergy-development yep 👍

RiFi2k:
@swalkinshaw I got this pretty much finished, I currently have tested it locally and have it deployed on my live staging server.

Just need to test and verify sftp is still going to be working as expected because ssh-hardening is using internal-sftp with chroot directories (which is safer anyways) but obviously users that are used to using sftp will most likely want to access the web root instead of the user home folder which is default. I only use ssh in my normal day to day but I'm going to try and make a point to finish up this test and submit a PR soon.

fullyint:
closed by #744

PR #702, opened 2016-12-06
Switch roles for more secure defaults in ssh

RiFi2k:
* Depreciate the sshd role (#561)
* Use more secure defaults for client and server ssh connections
* Securing the moduli file

Originally I had been trying to integrate the ansible-ssh-hardening role as a requirement but there were a few major roadblocks which ultimately lead me to a modified version of the original ported in as an included role. I'll explain my reasoning.

- When adding the role in as a requirement we lose the ability to offer `sshd_password_authentication` for the server without adding in a hack after the fact because they do not offer a variable at all for that line in sshd_config

- Probably my biggest issue though from a user standpoint is that if you do not have selinux configured and running on your server (which Trellis obviously doesn't configure, nor do I ever see that being integrated because new users have a hard enough time, without selinux permissions and errors which can be difficult for a everyday linux user even) it causes the role to display errors in red text during the provisioning which I feel like even though the role completes will scare a lot of users into thinking they have something broken on the server. The first time I even got confused for a second because they aren't 100% descriptive either.

- There is a lot of extra files and steps we will never use because they pertain to RHEL / CentOS, etc. Which isn't a big deal by itself but with everything else I just felt like it couldn't be added gracefully as a requirement.

So I took a crack and stripping out all the extras and keeping the same defaults we have now but also using all of their hardening in sshd_config plus I added the moduli file hardening. There is a reference to that being a good idea https://stribika.github.io/2015/01/04/secure-secure-shell.html as well as many other places online. Removing all strings under 2048 which are known to be insecure.

If this does go in then the documentation will need a pull request for a few small things as well, naming of variables, extra options, etc. I can do that if needed.

Also I have been running this on my live, staging, and local dev environments for over a week messing with all the different variables which work as expected.


fullyint:
@RiFi2k Thanks for all this work. This will be great.

**TL;DR** I propose the following:
- @RiFi2k: you squash the first 3 commits described below into your branch
- we merge your PR into the new `ssh-hardening` branch I've created as the base branch for this PR
- I'll add 3 more commits and make a PR to merge the branch into master
- People focus their review efforts on that latter PR as the complete revised sshd implementation

-----

I've added a [`ssh-hardening-demo`](https://github.com/roots/trellis/commits/ssh-hardening-demo) branch with changes that make sense to me. I've tested it quite thoroughly. Here are the commits:

0. `Switch roles for more secure defaults in ssh` (@RiFi2k's original commit)
1. `Streamline ssh-hardening role to Ubuntu, OpenSSH_6.6+`
2. `Enable AllowAgentForwarding, UsePAM; Restore MaxAuthTries 6`
3. `Use lineinfile to make moduli task more succinct and idempotent`
4. `Use stronger sshd HostKeys and ssh client HostKeyAlgorithms` (see [stribika article](https://stribika.github.io/2015/01/04/secure-secure-shell.html))
5. `Fix ssh config to handle custom options per Host` (see dev-sec/ansible-ssh-hardening#83)
6. `List only one Port in ssh config` (see dev-sec/ansible-ssh-hardening#84)

If you approve of commits 1-3, I think it'd be great if you would squash them into your original commit. The commits are reviews/changes I could have tried to put into the thread here but preferred to code up and test first. Now it's easiest for me to just share the my branch itself. Details below.

### 1. Streamline ssh-hardening role to Ubuntu, OpenSSH_6.6+
You already removed the non-Ubuntu stuff from the original role's tasks. This commit does the same for the templates. I also removed some variables whose names included `53` `59` and `66`. These refer to OpenSSH versions 5.3, 5.9, 6.6. Given that Ubuntu 14.04 and 16.04 both include > 6.6, we can just use the `66` variables, which I renamed with the `66` omitted.

According to [stribika](https://stribika.github.io/2015/01/04/secure-secure-shell.html) and https://wiki.mozilla.org/Security/Guidelines/OpenSSH I added `umac-128@openssh.com` to the list of MACs. However, I added it to the `ssh_macs_weak` instead of `ssh_macs_default`, just to be safe, because it was not mentioned by the authors at https://bettercrypto.org/static/applied-crypto-hardening.pdf

### 2. Enable AllowAgentForwarding, UsePAM; Restore MaxAuthTries 6
I changed `AllowAgentForwarding` to `yes` for a successful git clone on deploy.

I changed `UsePAM` to `yes` to avoid Debian [bug #751636](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=751636) with `openssh-server`. We can change to `UsePAM no` once the [Canonical Main repository](https://launchpad.net/ubuntu/xenial/+source/openssh) (`openssh 1:7.2p2-4`) includes [`openssh 1:7.2p2-6`](https://launchpad.net/ubuntu/+source/openssh/1:7.2p2-6). Trellis users haven't encountered this because we've had `UsePAM yes` so far.

I changed `MaxAuthTries` from `2` back to the `6` that has been in Trellis so far. If someone has 3-6 SSH keys and Trellis has managed to try them all till now, I want to avoid suddenly having Trellis try only 2 keys then fail. That could be challenging for people to debug if they don't think to run `ssh -v` to see which keys are being tried and _not_ tried. I'm not aware of `MaxAuthTries 6` being particularly risky, but correct me if I'm wrong.

I added `SendEnv` and `AcceptEnv` (empty by default) in case someone might need them.

### 3. Use lineinfile to make moduli task more succinct and idempotent
This change is in the spirit of Ansible's recommendation to use an Ansible module instead of the `command` or `shell` modules, where possible. This implementation using the `lineinfile` module makes the task idempotent, so it isn't marked 'changed' every time.

Any line in `/etc/ssh/moduli` that matches the regex will be removed. An example line to remove:
```
# Time Type Tests Tries Size Generator Modulus
20150522025931 2 6 100 1535 5 F4EE15F22E5F49997A027769656D...
```
The fifth column (`Size` of `1535`) is less than 2000 so we need to remove the line.

Explaining the regex: `^(\d+\s){4}1`
- `^`  -- the line must start with the pattern that follows
- `(\d+\s){4}`  -- 4 occurrences of one or more digits followed by a space (matches columns for `Time Type Tests Tries`)
- `1`  -- next column (`Size`) starts with `1` (i.e., is in range of 1000-1999)

Moduli `Size` on DigitalOcean Ubuntu 12.04, 14.04, and 16.04 is always a 4-digit number 1000 or higher, so the last bullet above should work: just matching a `Size` that starts with `1`.

If for some reason we believe moduli `Size` could occasionally be less than 4 digits, we would need regex like `^(\d+\s){4}(\d{1,3}|[0-1]\d{3})\s` where 
- `(\d{1,3}|[0-1]\d{3})\s`  -- matches `Size` less than 2000
  - `\d{1,3}`  -- a number of 1-3 digits (i.e., less than 1000)
  - `|[0-1]\d{3}` -- OR a 4-digit number starting with `0` or `1`
  - `\s`  -- a space after the 1-4 digit number, to ensure we don't match a 5+ digit number

RiFi2k:
My man!

It seems like you put a bunch of time into research and improving this even more, from quickly reading over what you listed above I'm glad you took the time to really dig in even further.

**1. Streamline ssh-hardening role to Ubuntu, OpenSSH_6.6+**
Good call on that, stribika is super aggressive with his rules anyways, last time I tried using that whole set I couldn't barely do anything without work arounds.

**2. Enable AllowAgentForwarding, UsePAM; Restore MaxAuthTries 6**
Good catch on all these, I have my own system for deploys so I completely forgot to even test that. Also I can for sure see people having issues with the 2 keys, I don't see any reason that would affect the overall security, especially when it's all getting a huge upgrade anyways.

**3. Use lineinfile to make moduli task more succinct and idempotent**
The moduli steps were actually secretly annoying me using shell commands, the lineinfile was an elegant solution for sure.

Just wrapping up a fun filled DDoS attack mitigation so I'm gonna crash out but tomorrow I should have some free time, if you want me to just add those in it won't be a problem, unless you want me to just remove my moduli section and make the few small edits, then you could throw a commit on top adding in the moduli part. It is technically outside the scope of even the original ssh-hardening role.

Just let me know, and thanks again for taking the time to do all this!

fullyint:
:sparkles::christmas_tree:DDoS:christmas_tree::sparkles:

Thanks for the quick reply!

Good point about the moduli being outside the original role's scope. I'd be happy for you to do the `lineinfile` moduli task in a separate commit of your own, still in this same PR.

- So, I'd say definitely squash commit `1 - Streamline ssh-hardening role to Ubuntu, OpenSSH_6.6+` into your original commit, just so the initial appearance of the role omits the non-Ubuntu stuff.
- Then maybe squash `2 -  Enable AllowAgentForwarding, UsePAM; Restore MaxAuthTries 6` in as well, or keep it a separate commit, given that it changes settings from the original role. Your choice. But I'd like that to be your commit. It's just some revisions I'm requesting on your PR.
- Then yeah, add the lineinfile moduli task in a separate commit from you. Again, it's yours, but just includes some revisions suggested in this PR.
- Then we'll probably merge and I'll add commits 4-6, then open a PR to master, etc.

Thanks!

RiFi2k:
@fullyint Thanks for the revisions! Have a happy holiday!

I think that push should take care of everything

fullyint:
Looks great! Thanks for all your work in making this happen.

Issue #697, opened 2016-12-02
Unable to open /requirements.yml


gabrielgiordan:
- [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/guidelines/blob/master/CONTRIBUTING.md)
- [ ] This is a feature request
- [x] This is a bug report
- [x] This request isn't a duplicate of an [existing issue](https://github.com/roots/trellis/issues)
- [x] I've read the [docs](https://roots.io/trellis/docs) and followed them (if applicable)

---

**What is the current behavior?**

I cannot vagrant provision the latest Trellis (master) on Windows 10.

**What is the expected or desired behavior?**

Be able to provision

---

## Bug report

**Please provide steps to reproduce, including full log output:**

I can ssh into it, but when vagrant up or vagrant provision it outputs this log:

`vagrant provision`

```
==> default: Running Ansible Galaxy install
==> default:  [ERROR]: Unable to open /requirements.yml: [Errno 2] No such file or
==> default: directory: '/requirements.yml'
==> default: Running Ansible Playbooks
==> default: ERROR! the playbook: dev.yml could not be found
The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what went wrong.
```

`vagrant provision --debug`

```
 INFO interface: info: ==> default: You are using pip version 8.1.1, however version 9.0.1 is available.
==> default: You should consider upgrading via the 'pip install --upgrade pip' command.
==> default: You are using pip version 8.1.1, however version 9.0.1 is available.
==> default: You should consider upgrading via the 'pip install --upgrade pip' command.
DEBUG ssh: Sending SSH keep-alive...
 INFO interface: info: Requirement already satisfied (use --upgrade to upgrade): markupsafe in /usr/local/lib/python2.7/dist-packages
 INFO interface: info: ==> default: Requirement already satisfied (use --upgrade to upgrade): markupsafe in /usr/local/lib/python2.7/dist-packages
==> default: Requirement already satisfied (use --upgrade to upgrade): markupsafe in /usr/local/lib/python2.7/dist-packages
DEBUG ssh: stderr: You are using pip version 8.1.1, however version 9.0.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.

 INFO interface: info: You are using pip version 8.1.1, however version 9.0.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
[...]

PR #705, opened 2016-12-09
Create /home/vagrant/trellis bindfs mount with proper permissions

fullyint:
Fixes an [issue](https://github.com/roots/trellis/issues/697#issuecomment-264594492) where Ansible, running in a VM on a Windows host, throws an error while interpreting a `.vault_pass` file as executable.

```
ERROR! Problem running vault password script / v a g r a n t / . v a u l t _ p a s s ([Errno 8]
Exec format error). If this is not a script, remove the executable bit from the file.
```

This PR implements a variation of [`@aoe`'s approach](https://discourse.roots.io/t/shared-trellis-folder-permissions-vault-problem/6640/7), using vagrant-bindfs to mount the `/vagrant` share with non-executable permissions. This PR makes a separate bindfs mount for `/vagrant/bin` executable, the only items that need to be executable, as far as I know.

Punishing details below.

### NFS performance
I could have wrapped the new vagrant-bindfs mounts (lines 85-87) in a conditional for Windows only, but it seems simpler without. I compared [times](http://stackoverflow.com/a/29132716/6847025) from 20 trials of a test playbook that runs various Ansible `local_action` tasks on the VM from a regular VirtualBox share (current Trellis) and from an NFS share (result of this PR). The NFS mean time of 23.245 seconds (_SD_ = 2.018) was slightly faster than the VB share mean of 23.603 seconds (_SD_ = 2.401), but the difference was not statistically significant, _p_ = 0.613. This suggests that changing to NFS doesn't incur any performance hit.

### Pipelining
Performance tests while preparing this PR also examined setting [`pipelining = True`](http://docs.ansible.com/ansible/intro_configuration.html#pipelining) in `ansible.cfg`.
>By default, this option is disabled to preserve compatibility with sudoers configurations that have requiretty (the default on many distros), but is highly recommended if you can enable it.

The Trellis Ubuntu Xenial distro does not have `requiretty` in `/etc/sudoers`, so :+1: pipelining. 

Repeating the performance tests mentioned above, but with pipelining enabled, the NFS mean time of 18.238 seconds (_SD_ = 0.323) was again slightly faster than the VB share mean of 18.618 seconds (_SD_ = 0.188), and the difference statistically significant, _p_ < 0.001 (very small difference nonetheless). This again suggests that changing to NFS doesn't incur any performance hit.

### Mount options
I tested various combinations of [`mount_options`](http://manpages.ubuntu.com/manpages/xenial/en/man5/nfs.5.html)`: ['rw', 'vers=3', 'tcp', fsc', 'actimeo=1']` that appear in various internet reports of how to speed up NFS performance on Vagrant. According to `nfsstat -m` (command run on the VM) our nfs shares automatically apply the `rw` and `vers=3`. The other options occasionally achieved tiny improvements that were statistically significant, but their effect size was so small they seemed negligible and not worth including.

The `fsc` would require installing and enabling `cachefilesd` on the VM, then remounting the NFS shares, which would complicate the `vagrant up` process, unless we start using a vagrant box with `cachefilesd` already installed/enabled. 

The `tcp` option is said to be slower, but helpful if NFS is used over "lossy networks." This didn't feel applicable, and testing results didn't show a clear benefit.

The `actimeo` option to set file attribute cache times would perhaps improve `gulp watch` response times if `gulp` were being run on the VM ([example](https://www.inovex.de/blog/doh-my-vagrant-nfs-is-slow/)), but the Roots recommendation is to run `gulp` on the host machine.

I also tested these mount options on the NFS share with the `site` files. The 20 iterations of `touch site/web/app/themes/sage/assets` then measuring the `gulp watch` build times and Browsersync page load times with a VM running roots/sage 8.5.0 again occasionally showed statistically significant differences, but again they had negligibly small effect sizes, so I omitted all mount options for `site`. Perhaps someone will want to revisit these mount options with Sage 9 when it's released.

### Permissions on .vault_pass
A nice thing about [`@aoe`'s approach](https://discourse.roots.io/t/shared-trellis-folder-permissions-vault-problem/6640/7) of `.vault_pass` into a new `vault/` dir is the `p: '0000,u=rD'` so `g` and `o` have no permissions for `.vault_pass`. However, for Trellis to sync a `vault/` directory by default, the directory would have to always be present, whether added to Trellis core, or created by every user, which seems impractical. Users who want tighter permissions on just `.vault_pass` can create their own directory and sync it as `@aoe` suggested. Even then, `.vault_pass` would still be readable by `vagrant` user, the user that I imagine will be used by all accessing the VM.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -66,11 +66,14 @@ Vagrant.configure('2') do |config|
     fail_with_message "vagrant-hostmanager missing, please install the plugin with this command:\nvagrant plugin install vagrant-hostmanager"
   end
 
+  bin_path = File.join(ANSIBLE_PATH.sub(__dir__, '/vagrant'), 'bin')
+
   if Vagrant::Util::Platform.windows? and !Vagrant.has_plugin? 'vagrant-winnfsd'
     wordpress_sites.each_pair do |name, site|
       config.vm.synced_folder local_site_path(site), remote_site_path(name, site), owner: 'vagrant', group: 'www-data', mount_options: ['dmode=776', 'fmode=775']
     end
-    config.vm.synced_folder File.join(ANSIBLE_PATH, 'hosts'), File.join(ANSIBLE_PATH.sub(__dir__, '/vagrant'), 'hosts'), mount_options: ['dmode=755', 'fmode=644']
+    config.vm.synced_folder '.', '/vagrant', mount_options: ['dmode=755', 'fmode=644']
+    config.vm.synced_folder File.join(ANSIBLE_PATH, 'bin'), bin_path, mount_options: ['dmode=755', 'fmode=755']
   else
     if !Vagrant.has_plugin? 'vagrant-bindfs'
       fail_with_message "vagrant-bindfs missing, please install the plugin with this command:\nvagrant plugin install vagrant-bindfs"
@@ -79,6 +82,9 @@ Vagrant.configure('2') do |config|
         config.vm.synced_folder local_site_path(site), nfs_path(name), type: 'nfs'
         config.bindfs.bind_folder nfs_path(name), remote_site_path(name, site), u: 'vagrant', g: 'www-data', o: 'nonempty'
       end
+      config.vm.synced_folder '.', '/vagrant-nfs', type: 'nfs'
+      config.bindfs.bind_folder '/vagrant-nfs', '/vagrant', o: 'nonempty', p: '0644,a+D'
+      config.bindfs.bind_folder bin_path, bin_path, perms: '0755'
     end
   end
 
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -7,7 +7,8 @@ ip = '192.168.50.5' # pick any local IP
 cpus = 1
 memory = 1024 # in MB
 
-ANSIBLE_PATH = __dir__ # absolute path to Ansible directory
+ANSIBLE_PATH = __dir__ # absolute path to Ansible directory on host machine
+ANSIBLE_PATH_ON_VM = '/home/vagrant/trellis' # absolute path to Ansible directory on virtual machine
 
 # Set Ansible paths relative to Ansible directory
 ENV['ANSIBLE_CONFIG'] = ANSIBLE_PATH
@@ -66,13 +67,13 @@ Vagrant.configure('2') do |config|
     fail_with_message "vagrant-hostmanager missing, please install the plugin with this command:\nvagrant plugin install vagrant-hostmanager"
   end
 
-  bin_path = File.join(ANSIBLE_PATH.sub(__dir__, '/vagrant'), 'bin')
+  bin_path = File.join(ANSIBLE_PATH_ON_VM, 'bin')
 
   if Vagrant::Util::Platform.windows? and !Vagrant.has_plugin? 'vagrant-winnfsd'
     wordpress_sites.each_pair do |name, site|
       config.vm.synced_folder local_site_path(site), remote_site_path(name, site), owner: 'vagrant', group: 'www-data', mount_options: ['dmode=776', 'fmode=775']
     end
-    config.vm.synced_folder '.', '/vagrant', mount_options: ['dmode=755', 'fmode=644']
+    config.vm.synced_folder ANSIBLE_PATH, ANSIBLE_PATH_ON_VM, mount_options: ['dmode=755', 'fmode=644']
     config.vm.synced_folder File.join(ANSIBLE_PATH, 'bin'), bin_path, mount_options: ['dmode=755', 'fmode=755']
   else
     if !Vagrant.has_plugin? 'vagrant-bindfs'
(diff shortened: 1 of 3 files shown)


Vagrantfile
```diff
@@ -67,11 +67,13 @@ Vagrant.configure('2') do |config|
     fail_with_message "vagrant-hostmanager missing, please install the plugin with this command:\nvagrant plugin install vagrant-hostmanager"
   end
 
+  bin_path = File.join(ANSIBLE_PATH.sub(__dir__, '/vagrant'), 'bin')
```
swalkinshaw: blank line after this please 👼

arashohadi:
Hi @fullyint! 

I just tried implementing the changes from this pull request (was on https://github.com/roots/trellis/commit/8453c53213eb722e5293dded8cceff3ebf33a493).

I am on a Windows host and using Winnfsd. On https://github.com/roots/trellis/commit/8453c53213eb722e5293dded8cceff3ebf33a493 everything works fine and Winnfsd is able to mount the shared folders that I have.

But when running ```vagrant up``` with the changes from this file, it seems the winnfsd.exe starts up but instantly crashes (noted in Task Manager).

Running ```vagrant up --debug``` gives me this information when trying to mount (which is not so strange considering the .exe has crashed).
 
```
set -e
mkdir -p /vagrant-nfs
mount -o vers=3,udp,vers=3,udp,nolock 192.168.50.1:/C/Projects/Trellis /vagrant-nfs
if command -v /sbin/init && /sbin/init --version | grep upstart; then
  /sbin/initctl emit --no-wait vagrant-mounted MOUNTPOINT=/vagrant-nfs
fi


Stdout from the command:



Stderr from the command:

mount.nfs: Connection timed out
```

Just let me know if you need more information from me to help you.

arashohadi:
Some more information:

- Tried upgrading vagrant-winnfsd from ```1.3.0``` to ```1.3.1```. Didn't help unfortunately, same behavior as above.
- I also tried upgrading vagrant-bindfs from ```0.4.9``` to ```1.0.1```. This gave another interesting error directly after running ```vagrant up``` command.

```
vagrant-bindfs:
* Destination path '/vagrant/bin' is part of a reserved subtree for Vagrant use
```

I looked at the repo and found this commit https://github.com/gael-ian/vagrant-bindfs/commit/f07ab7c18059e0f4e5c245577c8f6a4937d887f7. Especially this line is interesting 

```Remember that Vagrant use `/vagrant` on guest side as a shared directory dedicated to provisioning and configuration. Binding a folder to `/vagrant` or one of its subfolders will fail.```

**Comment by fullyint** [comment]
@arashohadi Thank you for testing. Your information is very helpful. After updating my `vagrant-bindfs` plugin, I too get the `'/vagrant/bin' is part of a reserved subtree` error.

If `vagrant-bindfs` won't allow a mountpoint in `/vagrant`, one option could be to put the mountpoint in the `vagrant` user's home directory, changing the PR to something like this:

```diff
- bin_path = File.join(ANSIBLE_PATH.sub(__dir__, '/vagrant'), 'bin')
+ bin_path = File.join(ANSIBLE_PATH.sub(__dir__, '/home/vagrant/trellis'), 'bin')

    if Vagrant::Util::Platform.windows? and !Vagrant.has_plugin? 'vagrant-winnfsd'
      wordpress_sites.each_pair do |name, site|
        config.vm.synced_folder local_site_path(site), remote_site_path(name, site), owner: 'vagrant', group: 'www-data', mount_options: ['dmode=776', 'fmode=775']
      end
-     config.vm.synced_folder '.', '/vagrant', mount_options: ['dmode=755', 'fmode=644']
+     config.vm.synced_folder '.', '/home/vagrant/trellis', mount_options: ['dmode=755', 'fmode=644']
      config.vm.synced_folder File.join(ANSIBLE_PATH, 'bin'), bin_path, mount_options: ['dmode=755', 'fmode=755']
    else
      if !Vagrant.has_plugin? 'vagrant-bindfs'
        fail_with_message "vagrant-bindfs missing, please install the plugin with this command:\nvagrant plugin install vagrant-bindfs"
      else
        wordpress_sites.each_pair do |name, site|
          config.vm.synced_folder local_site_path(site), nfs_path(name), type: 'nfs'
          config.bindfs.bind_folder nfs_path(name), remote_site_path(name, site), u: 'vagrant', g: 'www-data', o: 'nonempty'
        end
-       config.vm.synced_folder '.', '/vagrant-nfs', type: 'nfs'
+       config.vm.synced_folder '.', '/vagrant', type: 'nfs'
-       config.bindfs.bind_folder '/vagrant-nfs', '/vagrant', o: 'nonempty', p: '0644,a+D'
+       config.bindfs.bind_folder '/vagrant', '/home/vagrant/trellis', o: 'nonempty', p: '0644,a+D'
        config.bindfs.bind_folder bin_path, bin_path, perms: '0755'
      end
    end

    provisioner = Vagrant::Util::Platform.windows? ? :ansible_local : :ansible
-   provisioning_path = Vagrant::Util::Platform.windows? ? ANSIBLE_PATH.sub(__dir__, '/vagrant') : ANSIBLE_PATH
+   provisioning_path = Vagrant::Util::Platform.windows? ? ANSIBLE_PATH.sub(__dir__, '/home/vagrant/trellis') : ANSIBLE_PATH
```

This change would require users to `cd ~/trellis` after `vagrant up` (instead of `cd /vagrant`).

I don't know if that will resolve the `winnfsd` crashing problem, however.
I will set myself up with a Windows box for testing, but I can't get to it for a few days.
Again, thank you very much for testing!

fullyint:
I added a commit based on the diff above. It works on my Windows 10 with encrypted `vault.yml` files and `.vault_pass`.

On the VM, the `/vagrant` directory will still exist with the same content as before but there will be a new `/home/vagrant/trellis` directory with the needed permissions on files and directories.

Windows users would switch to running Ansible commands from `~/trellis`. Although non-Windows users could continue with `/vagrant`, we might as well make it the general rule to say, "find your Trellis files on the VM in `~/trellis`." I augmented the `post_up_message` to that effect.

Does anyone see any problem with this change to `/home/vagrant/trellis`?

Note that if the `Vagrantfile` is moved to the parent directory on the host machine, `/vagrant` on the VM will have the `site` and `trellis` subdirectories like in the past, but `~/trellis` on the VM will still have just the Trellis files, 

I think only the [Windows docs](https://roots.io/trellis/docs/windows/) would need a few minor updates.

PR #706, opened 2016-12-10
Create project_shared_children files if they do not exist

fullyint:
Fixes this [discourse thread](https://discourse.roots.io/t/create-symlinks-for-files/7086).

The "[Ensure shared sources are present](https://github.com/roots/trellis/blob/8453c53213eb722e5293dded8cceff3ebf33a493/roles/deploy/tasks/share.yml#L5-L10)" task uses the [file module](http://docs.ansible.com/ansible/file_module.html) to create directories or files in the `shared` directory.

The [comments for `project_shared_children`](https://github.com/roots/trellis/blob/8453c53213eb722e5293dded8cceff3ebf33a493/roles/deploy/defaults/main.yml#L25) in the deploy/defaults indicates that items in the list "are created if they don't exist." The problem is that if a user specifies `type: file`, this is fed to the `state` option for the file module, but when `state` is "`file`, the file will NOT be created if it does not exist." 

In addition, the file module will not create the subdirectories a file may be in, if they are missing. 

This PR splits the "Ensure shared sources are present" task in two. The first task creates all the shared directories, including the parent directories of shared files. The second task touches the files.
> If `touch` (new in 1.4), an empty file will be created if the `path` does not exist, while an existing file or directory will receive updated file access and modification times  [-- ref](http://docs.ansible.com/ansible/file_module.html) 

Touch is necessary if we actually want to create files, but unfortunately it is never idempotent.

This approach should still allow for `type`/`state` values such as `absent`.

### DIFF
--- a/roles/deploy/tasks/share.yml
+++ b/roles/deploy/tasks/share.yml
@@ -5,7 +5,7 @@
 - name: Ensure shared sources are present -- directories
   file:
     path: "{{ deploy_helper.shared_path }}/{{ project_shared_file | ternary(item.src | dirname, item.src) }}"
-    state: "{{ project_shared_file | ternary('directory', item.type | default('directory')) }}"
+    state: "{{ project_shared_file | ternary('directory', item.type | default('directory')) | lower }}"
     mode: "{{ project_shared_file | ternary('0755', item.mode | default('0755')) }}"
   with_items: "{{ project_shared_children }}"
 
@@ -17,18 +17,26 @@
   with_items: "{{ project_shared_children }}"
   when: project_shared_file
 
+- name: Ensure parent directories for shared paths are present
+  file:
+    path: "{{ deploy_helper.new_release_path }}/{{ item.path | default('') | dirname }}"
+    state: directory
+  with_items: "{{ project_shared_children }}"
+
 - name: Ensure shared paths are absent
   file:
-    path: "{{ deploy_helper.new_release_path }}/{{ item.path }}"
+    path: "{{ deploy_helper.new_release_path }}/{{ item.path  | default('') }}"
     state: absent
   with_items: "{{ project_shared_children }}"
+  when: item.type | default('directory') | lower != 'absent'
 
 - name: Create shared symlinks
   file:
-    path: "{{ deploy_helper.new_release_path }}/{{ item.path }}"
+    path: "{{ deploy_helper.new_release_path }}/{{ item.path | default('') }}"
     src: "{{ deploy_helper.shared_path }}/{{ item.src }}"
     state: link
   with_items: "{{ project_shared_children }}"
+  when: item.type | default('directory') | lower != 'absent'
 
 - include: "{{ deploy_share_after | default('../hooks/example.yml') }}"
   tags: deploy-share-after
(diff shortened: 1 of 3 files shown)


fullyint:
Thank you @swalkinshaw for pointing out that the symlink creation will fail if the parent directory of `project_shared_children[*].path` doesn't exist in the `new_release_path`. I added a commit to create the parent directory.

I also added the `lower` filter to prevent the `file` module from failing if a user defines `type: File`. 

I also added a comment in `roles/deploy/defaults/main.yml`...
```
Use `type: absent` to remove an item from `/shared`.
```

I also made it explicit that the `mode` value should be quoted to avoid the potential confusion around how "python and yaml (pyyaml) work with each other in regards to octal values" ([discussion](https://github.com/ansible/ansible/issues/9196#issuecomment-57168074)).

roles/deploy/tasks/share.yml
```diff
@@ -2,25 +2,41 @@
...
-    path: "{{ deploy_helper.new_release_path }}/{{ item.path }}"
+    path: "{{ deploy_helper.new_release_path }}/{{ item.path  | default('') }}"
     state: absent
   with_items: "{{ project_shared_children }}"
+  when: item.type | default('directory') | lower != 'absent'
 
 - name: Create shared symlinks
   file:
-    path: "{{ deploy_helper.new_release_path }}/{{ item.path }}"
+    path: "{{ deploy_helper.new_release_path }}/{{ item.path | default('') }}"
```
swalkinshaw: Shouldn't we enforce the `path` here? Items set to `absent` aren't included here anyway.

roles/deploy/tasks/share.yml
```diff
@@ -2,25 +2,41 @@
...
+- name: Ensure parent directories for shared paths are present
+  file:
+    path: "{{ deploy_helper.new_release_path }}/{{ item.path | default('') | dirname }}"
+    state: directory
   with_items: "{{ project_shared_children }}"
 
 - name: Ensure shared paths are absent
   file:
-    path: "{{ deploy_helper.new_release_path }}/{{ item.path }}"
+    path: "{{ deploy_helper.new_release_path }}/{{ item.path  | default('') }}"
```
swalkinshaw: Same with here

roles/deploy/tasks/share.yml
fullyint: The `absent` option allows an easy way for someone to remove something from `/shared` if it will no longer be used in the next deployment(s). Although it is not necessary to remove items, and although people could remove them manually, facilitating this `absent` option requires no extra tasks, so I thought it was convenient.

However, the `absent` option does require the the two tasks you've marked to include `when` conditions and `item.path | default('')`. Yes these two tasks will skip for items set to `absent`, but Ansible still attempts to template all the variables in the task. So, without a default for `item.path`, the task would fail when people omit `path` for items set to `absent`. 

If we keep this `absent` option, I would like people to be able to omit `path` because an item to be removed from `/shared` will have no `path` in the release directory, so the `path` parameter doesn't make sense to include.

fullyint:
Latest commit `Simplify handling ...`
- removes the `absent` feature
- splits the first "ensure shared directories are present" task in two:
   - first create directories for items with `type: directory`
   - next create parent directories for items with `type: file`

These changes favor "more tasks that are simple" over "fewer tasks that are complex." I routinely have to fight my inclination toward the latter.

PR #707, opened 2016-12-10
Remove obsolete references to vars_plugins path

fullyint:
When #548 added `lib/trellis/plugins/vars/vars.py` it added the `vars_plugins` config to `ansible.cfg`. 

Later, #684 removed `lib/trellis/plugins/vars/vars.py` but neglected to remove the `vars_plugins` config from `ansible.cfg` and `Vagrantfile`. This PR fixes that.


swalkinshaw:
👏

PR #710, opened 2016-12-13
Remove creation of unused socket directory

RiFi2k:
The php7.1-fpm socket is created at /var/run so no need for this directory to be created


fullyint:
Thanks for spotting this! 👍

PR #715, opened 2016-12-23
Update build routine suggestions for Sage 9

davekiss:
These changes ensure that Sage 9's build routine is run successfully on deployment by removing the bower install process in favor of composer and compiling assets with Webpack instead of Gulp


QWp6t:
Should prob change the npm stuff to yarn too while we're at it.

`npm install` -> `yarn`
`npm run xxxx` -> `yarn run xxxx`

swalkinshaw:
Thanks! 👍

PR #717, opened 2016-12-26
Fix #522 - Check if branch is master and prompt to continue

### LINKED ISSUE #522: Add warning during deploys about current branch
If someone is deploying from a non `master` branch, we should probably add a warning message. And maybe even a prompt to continue.

mitramejia:
Before executing the deploy command. Verify if the actual branch is master, if not prompt a warning asking to continue or abort the deploy.

### DIFF
--- a/bin/deploy.sh
+++ b/bin/deploy.sh
@@ -5,6 +5,9 @@ DEPLOY_CMD="ansible-playbook deploy.yml -e env=$1 -e site=$2"
 ENVIRONMENTS=( hosts/* )
 ENVIRONMENTS=( "${ENVIRONMENTS[@]##*/}" )
 NUM_ARGS=2
+BRANCH_NAME="$(git symbolic-ref HEAD 2>/dev/null)" ||
+BRANCH_NAME="(unnamed branch)"     # detached HEAD
+BRANCH_NAME=${BRANCH_NAME##refs/heads/}
 
 show_usage() {
   echo "Usage: deploy <environment> <site name>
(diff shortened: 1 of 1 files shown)


bin/deploy.sh
```diff
@@ -5,6 +5,9 @@ DEPLOY_CMD="ansible-playbook deploy.yml -e env=$1 -e site=$2"
 ENVIRONMENTS=( hosts/* )
 ENVIRONMENTS=( "${ENVIRONMENTS[@]##*/}" )
 NUM_ARGS=2
+BRANCH_NAME="$(git symbolic-ref HEAD 2>/dev/null)" ||
+BRANCH_NAME="(unnamed branch)"     # detached HEAD
+BRANCH_NAME=${BRANCH_NAME##refs/heads/}
```
swalkinshaw: Think we can simplify this a bit more:

```bash
BRANCH_NAME="$(git rev-parse --abbrev-ref HEAD)"
```

Nothing else should be necessary since we're only checking if it's not `master`.

swalkinshaw:
@mitramejia sorry about this but we realized we lacked some details in the original issue. I added more details in https://github.com/roots/trellis/issues/522.

These checks will need to be done in the deploy role/playbook in order to properly check a WP site's `branch`. I'll close this PR for now, you can re-open if you want to re-use this one or just open another PR.

swalkinshaw:
Oh and thanks for the work on this :) If you want to continue that's appreciate, if not let us know and we can take over the work.

bin/deploy.sh
mitramejia: Great. Thanks for pointing this out.

mitramejia:
Great @swalkinshaw I'll look into it. 👍 

And you're welcome! Thank you for making this project

PR #720, opened 2016-12-29
Fix #718 - change method of updating theme paths

### LINKED ISSUE #718: Deploy can break sidebars/widgets if multiple sites defined
[Update WP theme paths](https://github.com/roots/trellis/blob/3e32e2ae9cc78010e8b36ba63cb52da52ed1582c/roles/deploy/hooks/finalize-after.yml#L10-L14) is the culprit.

needs investigating, related discourse threads:

https://discourse.roots.io/t/wordpress-4-7-update-moved-all-widgets/8277/6
https://discourse.roots.io/t/update-wp-theme-paths-what-does-this-hook-do/8296

swalkinshaw:
After a deploy we were running `switch_theme` would had the side effect of sometimes making sidebar widgets inactive.

This uses a simpler, more direct method of updating `stylesheet_root` and `template_root` which should be less error-prone since it does less.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Fix #718 - improve method of updating theme paths ([#720](https://github.com/roots/trellis/pull/720))
 * Create `/home/vagrant/trellis` bindfs mount with proper permissions ([#705](https://github.com/roots/trellis/pull/705))
 
 ### 0.9.9: December 14th, 2016
--- a/roles/deploy/hooks/finalize-after.yml
+++ b/roles/deploy/hooks/finalize-after.yml
@@ -24,11 +24,23 @@
     chdir: "{{ deploy_helper.current_path }}"
   when: wp_installed | success and project.multisite.enabled | default(false)
 
-- name: Update WP theme paths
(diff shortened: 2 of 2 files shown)


coreyworrell:
The reason menus and sidebar widgets become inactive is because of the two functions registered in `wp-includes/default-filters.php`

```php
add_action( 'after_switch_theme', '_wp_menus_changed' );
add_action( 'after_switch_theme', '_wp_sidebars_changed' );
```

I still have no idea why the issue only happens occasionally, but it appears that you can bypass those by pretending the theme switch came from the Customizer:

**wp-includes/theme.php** [`check_theme_switched()`] (this is bound to `init` action, so it would only happen on first page load after the deployment)

```php
if ( get_option( 'theme_switched_via_customizer' ) ) {
	remove_action( 'after_switch_theme', '_wp_menus_changed' );
	remove_action( 'after_switch_theme', '_wp_sidebars_changed' );
	update_option( 'theme_switched_via_customizer', false );
}
```

The new method you have here should probably work fine too, but if you didn't want to hard code a path in there then the following _should_ work:

```bash
wp eval 'wp_clean_themes_cache(); update_option("theme_switched_via_customizer", true); switch_theme(get_stylesheet());'
```

PR #721, opened 2016-12-29
Create ssl_enabled helper var

fullyint:
This is a convenience, not a necessity. Originally part of #630.


swalkinshaw:
:shipit:

PR #722, opened 2016-12-29
Make handler out of task disabling acme challenge sites

fullyint:
Originally part of #630:

In current Trellis, suppose the playbook fails after the "[Enable Nginx sites](https://github.com/roots/trellis/blob/6ad6fcbb6e0c0cfdaaf7ecb734c2e867ff0437a7/roles/letsencrypt/tasks/nginx.yml#L21)" task, but before the "[Disable Nginx site](https://github.com/roots/trellis/blob/6ad6fcbb6e0c0cfdaaf7ecb734c2e867ff0437a7/roles/letsencrypt/tasks/certificates.yml#L34)" task. Some `letsencrypt-example.com.conf` files would remain enabled and their server blocks would have the potential to interfere with future Nginx confs, depending on how the user responds to the failed playbook.

This PR moves the "Disable" task to a handler so it will always run and disable temp confs/sites, regardless whether the playbook fails. I placed the handler in `roles/common` before the `reload nginx` handler so that it would run before Nginx is reloaded. If the handler were in the `letsencrypt` role, it would be run after the Nginx reload handler.

> Notify handlers are always run in the same order they are defined, _not_ in the order listed in the notify-statement. -- _[Ansible handler docs](http://docs.ansible.com/ansible/playbooks_intro.html#handlers-running-operations-on-change)_

### DIFF
--- a/roles/common/handlers/main.yml
+++ b/roles/common/handlers/main.yml
@@ -1,4 +1,11 @@
 ---
+- name: disable temporary challenge sites
+  file:
+    path: "{{ nginx_path }}/sites-enabled/letsencrypt-{{ item.key }}.conf"
+    state: absent
+  with_dict: "{{ wordpress_sites }}"
+  notify: reload nginx
+
 - name: restart memcached
   service:
     name: memcached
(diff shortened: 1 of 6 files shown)


swalkinshaw:
:shipit:

PR #724, opened 2016-12-30
Vagrant: add automatic support for landrush 

swalkinshaw:
Previously we documented these changes that people would need to manually make. This handles it automatically so landrush is supported out of the box if the user installs the plugin.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -59,12 +59,16 @@ Vagrant.configure('2') do |config|
 
   redirects = site_hosts.flat_map { |host| host['redirects'] }.compact
 
-  if Vagrant.has_plugin? 'vagrant-hostmanager'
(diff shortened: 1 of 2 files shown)


Vagrantfile
```diff
@@ -63,8 +63,12 @@ Vagrant.configure('2') do |config|
     config.hostmanager.enabled = true
     config.hostmanager.manage_host = true
     config.hostmanager.aliases = hostnames + redirects
+  elsif Vagrant.has_plugin? 'landrush'
```
kalenjohnson: This probably isn't going to work as expected, many users will have both hostmanager and landrush installed.

kalenjohnson:
Much better, you are my hero wilkinsin

PR #726, opened 2017-01-04
Require Ansible 2.2 for better SSH error output

fullyint:
Ansible 2.2 (released Oct 31, 2016) displays [additional SSH error output](https://github.com/ansible/ansible/commit/9ec69ee). It would be particularly useful for Trellis users to see the previously hidden `WARNING: REMOTE HOST IDENTIFICATION HAS CHANGED!` message, especially because #716 plans to change host keys.


swalkinshaw:
:shipit:

PR #728, opened 2017-01-07
Fix #727 - HSTS: default preload to off

### LINKED ISSUE #727: Do not default to preloading HSTS

The [HSTS defaults](https://github.com/roots/trellis/blob/cbe647a26a95126a3cf7fd8ace7f1f2f0454adf0/roles/nginx/defaults/main.yml#L12) include
`nginx_hsts_preload: true`

Sites that enable HSTS should not send the `preload` directly by default.
I was contacted by someone who unintentionally ended up on the preload list. Removal is slow and painful for such sites: https://hstspreload.org/#removal
(I deal with multiple removal requests daily.)

Preloaded HSTS should never be the default setting. It has no effect unless the site is submitted to https://hstspreload.org , which should be done with the knowledge and consent of the site operator. It's fine to encourage preloaded HSTS, but it should be an explicit opt-in.

I would send a pull request to change the default value, but I don't know how to handle the migration path for existing project that rely on the default value to [stay preloaded](https://hstspreload.org/#continued-requirements). However, the setting should be `false` for new projects.

swalkinshaw:
The HSTS `preload` option defaulted to true. This turns it off by default since the HSTS preload list should be opt-in as it also requires a submission to https://hstspreload.org.

**NOTE**: If you submitted your site to the HSTS preload list and want to keep using the `preload` option, then set the variable:

```yaml
# group_vars/production/wordpress_sites.yml (example)

example.com:
  # rest of site config
  ssl:
    enabled: true
    provider: letsencrypt
    hsts_max_age: 31536000
    hsts_include_subdomains: true
    hsts_preload: true
```

Or set the global default:

```yaml
nginx_hsts_preload: true
```

See https://roots.io/trellis/docs/ssl/ for more information.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* [BREAKING] Fix #727 - HSTS: default preload to off ([#728](https://github.com/roots/trellis/pull/728))
 * `Vagrantfile`: add automatic support for landrush ([#724](https://github.com/roots/trellis/pull/724))
 * Suppress extra output in SSL certificates ([#723](https://github.com/roots/trellis/pull/723))
 * Fix #718 - improve method of updating theme paths ([#720](https://github.com/roots/trellis/pull/720))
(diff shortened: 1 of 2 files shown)


swalkinshaw:
Doc updates with much more information on this: https://github.com/roots/docs/pull/66

PR #730, opened 2017-01-09
Include wp-*.php rewrite rule in nginx config for non-multisite sites.

BrandonMathis:
## Why
`wp-*.php` to `/wp/$1` nginx rewrite was not present for non-multisite sites.

The wp-*.php rewrite condition in the wordpress-site.conf.j2 was never being included in `/etc/sites-enabled/<sitename>.conf` nginx configs. Without this, all requests to /wp-*.php files 404.

### DIFF
--- a/roles/wordpress-setup/templates/wordpress-site.conf.j2
+++ b/roles/wordpress-setup/templates/wordpress-site.conf.j2
@@ -19,11 +19,11 @@ server {
 
   {% if item.value.multisite.enabled | default(false) -%}
     {% if item.value.multisite.subdomains | default(false) -%}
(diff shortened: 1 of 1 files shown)


retlehs:
> Without this, all requests to /wp-.php files 404.

that shouldn't be happening. is there something that you maybe have changed on your trellis setup that's causing this?

swalkinshaw:
I don't really understand the problem here. I'm going to close this unless you have more information @BrandonMathis. Feel free to re-open with more details, or you can post on https://discourse.roots.io/ if it's something specific to your setup.

PR #732, opened 2017-01-10
Begin adding configuration for SSL termination

nathanielks:
This PR adds the `load_balancer_enabled` and `ssl_terminated` helpers.
It checks for the `load_balancer` key on a site. For example:

```yaml
wordpress_sites:
  example.com:
    // blah blah
    ssl:
      enabled: false
    load_balancer:
      enabled: true
      ssl_terminated: true
    // blah blah
```

The pattern it follows is the load balancer will send https traffic to
port 80 and http traffic to port 8080. Because SSL has already been
handled, the server doesn't need to do anything over 443, so it can be
treated like a normal web request (thus leaving everything alone and
communicating over port 80). Port 8080 is then used to receive http
traffic and then force redirect to the https version of the site. To
coerce WordPress into thinking this is an HTTPS request, we set the
`HTTPS` PHP `$_SERVER` variable to make `is_ssl()` work inside
WordPress.

### DIFF
--- a/roles/wordpress-setup/defaults/main.yml
+++ b/roles/wordpress-setup/defaults/main.yml
@@ -3,3 +3,4 @@ nginx_includes_templates_path: nginx-includes
 nginx_includes_deprecated: roles/wordpress-setup/templates/includes.d
 nginx_includes_pattern: "^({{ nginx_includes_templates_path | regex_escape }}|{{ nginx_includes_deprecated | regex_escape }})/(.*)\\.j2$"
 nginx_includes_d_cleanup: true
+ssl_terminated_http_port: 8080
--- a/roles/wordpress-setup/defaults/main.yml
+++ b/roles/wordpress-setup/defaults/main.yml
@@ -3,4 +3,4 @@ nginx_includes_templates_path: nginx-includes
 nginx_includes_deprecated: roles/wordpress-setup/templates/includes.d
 nginx_includes_pattern: "^({{ nginx_includes_templates_path | regex_escape }}|{{ nginx_includes_deprecated | regex_escape }})/(.*)\\.j2$"
 nginx_includes_d_cleanup: true
-ssl_terminated_http_port: 8080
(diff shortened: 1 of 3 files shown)


roles/wordpress-setup/defaults/main.yml
```diff
@@ -3,3 +3,4 @@ nginx_includes_templates_path: nginx-includes
 nginx_includes_deprecated: roles/wordpress-setup/templates/includes.d
 nginx_includes_pattern: "^({{ nginx_includes_templates_path | regex_escape }}|{{ nginx_includes_deprecated | regex_escape }})/(.*)\\.j2$"
 nginx_includes_d_cleanup: true
+ssl_terminated_http_port: 8080
```
swalkinshaw: How about making this per site? 

```yaml
wordpress_sites:
  example.com:
    // blah blah
    ssl:
      enabled: false
    load_balancer:
      enabled: true
      ssl_terminated: true
      port: 8080
    // blah blah
```

nathanielks: @swalkinshaw I do believe the latest resolves this!

roles/wordpress-setup/templates/wordpress-site.conf.j2
```diff
@@ -66,17 +66,20 @@ server {
...
     fastcgi_param DOCUMENT_ROOT $realpath_root;
+    {% if ssl_terminated -%}
+    fastcgi_param HTTPS on;
+    {% endif %}
     fastcgi_pass unix:/var/run/php-fpm-wordpress.sock;
   }
 }
 
-{% if ssl_enabled %}
+{% if ssl_enabled or ssl_terminated %}
```
swalkinshaw: I might rather duplicate this whole block at this point. What do you think? Meaning one for `ssl_enabled` and one for `ssl_terminated`

nathanielks: The DRYist in me is like "nah" because most of it is duplicated, but extra verbosity doesn't usually hurt anyone. @fullyint what's your feeling on this?

PR #734, opened 2017-01-16
Enable SKIP_GALAXY env var to skip galaxy install

nathanielks:
This would add the ability to skip the Ansible Galaxy requirements installation step by prefixing `vagrant up` or `vagrant provision` commands with the env var `SKIP_GALAXY=1`. Any non-empty value will work. For example:

```sh
SKIP_GALAXY=1 vagrant provision
```

Installing the requirements on every provision can get tiresome when running playbooks over and over (say when you have a modified version of Trellis, for instance).


swalkinshaw:
👍

PR #736, opened 2017-01-18
compatibility with capistrano defaults

morrislaptop:


swalkinshaw:
Does Capistrano even use the `bare` option?

https://github.com/capistrano/capistrano/blob/3d9800d977294c173b207cb99dda550c05966daa/lib/capistrano/scm/git.rb#L38-L44

swalkinshaw:
It uses `--mirror` which the Ansible module doesn't seem to support. There is a difference though.

morrislaptop:
It uses the mirror option which implies bare. Is there a way to use the
mirror optional with the Ansible git module?

swalkinshaw:
I don't think `bare` is good enough:

> Set up a mirror of the source repository. This implies --bare. Compared to --bare,
    --mirror not only maps local branches of the source to local branches of the target,
    it maps all refs (including remote-tracking branches, notes etc.) and sets up a
    refspec configuration such that all these refs are overwritten by a git remote
    update in the target repository.

We'd have to switch to using `command` and implementing the raw Git command.

swalkinshaw:
Closing since it's stale. Feel free to re-open/update.

PR #737, opened 2017-01-19
Enable letsencrypt role to respect the "site" var

nathanielks:
When provisioning a machine, if I pass the `site` var via `-e site=example.com`, I expect the letsencrypt role to make changes only to that site. This enables that ability by limited the sites to just the site provided, _if it's passed._ If no site is passed, it behaves like normal.


nathanielks:
Unnecessary. https://discourse.roots.io/t/to-deploy-a-second-site/5229/6 accomplishes what I wanted to do

swalkinshaw:
This should too: https://github.com/roots/trellis/pull/630

PR #739, opened 2017-01-21
Update nginx wiki link

jbicha:
I linked to the paragraph _above_ Virtualbox since linking directly to VirtualBox cut off the top part of the paragraph when I tested.

### DIFF
--- a/roles/wordpress-setup/templates/wordpress-site.conf.j2
+++ b/roles/wordpress-setup/templates/wordpress-site.conf.j2
@@ -13,7 +13,7 @@ server {
   charset utf-8;
 
   {% if env == 'development' -%}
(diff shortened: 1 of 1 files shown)


swalkinshaw:
fwiw https://www.nginx.com/resources/wiki/start/topics/tutorials/config_pitfalls/#virtualbox works just fine for me.

I'd rather link to that since the URL matters a bit in this case.

jbicha:
@swalkinshaw Could you check in Firefox?

It worked for me in Chromium 55 but not on Firefox 51 Beta.

swalkinshaw:
Yeah it's not great in Firefox. But Chrome is way higher in usage and the URL is correct anyway.

swalkinshaw:
Thanks 👍

PR #742, opened 2017-01-24
Fix apt_packages_custom

jac-firemancreative:
Commit #54ab64c0b3f376ac98c377eebdd825be033b4388 from yesterday breaks provisioning with the following error: "the python mysqldb module is required"

This PR fixes it.

### DIFF
--- a/roles/common/defaults/main.yml
+++ b/roles/common/defaults/main.yml
@@ -17,4 +17,4 @@ apt_packages_default:
   - libnss-myhostname
 
 apt_packages_custom: []
-apt_packages_install: "{{ default_apt_packages + custom_apt_packages }}"
+apt_packages_install: "{{ apt_packages_default + apt_packages_custom }}"


nathanielks:
you muh boy, @jac-firemancreative

fullyint:
@jac-firemancreative @nathanielks 
What do you think about dropping the `apt_packages_install` var from `defaults` then making this change to the task?

```diff
 - name: Checking essentials
   apt:
     name: "{{ item }}"
     state: present
     update_cache: true
     cache_valid_time: "{{ apt_cache_valid_time }}"
-  with_items: "{{ apt_packages_install }}"
+  with_items:
+    - "{{ apt_packages_default }}"
+    - "{{ apt_packages_custom }}"
```

This feels a tiny bit more "ansible-like" to me, but either way works.

nathanielks:
My only vote against that would be having `apt_packages_install` allows the user to override _all_ the apt packages if they so choose. It's a case of "I don't know why they would want to, but it's at least nice to have the option"

nathanielks:
Though I suppose they could also override `apt_packages_default` as well... sure! It gets a 👍  from me!

swalkinshaw:
We can change/improve this after but wanted the fix in ASAP.

PR #745, opened 2017-02-01
Change remote-user role to connection role: tests host key, user

fullyint:
This PR adds a task that provides guidance when SSH fails due to changed host keys:
```
TASK [connection : Warn about change in host keys] *****************************
System info:
  Ansible 2.2.1.0; Darwin
  Trellis at "Accommodate template inheritance for nginx confs"
WARNING: REMOTE HOST IDENTIFICATION HAS CHANGED!

If this change in host keys is expected (e.g., if you rebuilt the server
or if the Trellis sshd role made changes recently), then run the following
command to clear the old host key from your known_hosts.

  ssh-keygen -R 45.55.28.29

Then try your Trellis playbook or SSH connection again.

If the change is unexpected, cautiously consider why the host identification
may have changed and whether you may be victim to a man-in-the-middle attack.

The fingerprint for the RSA key sent by the remote host is
SHA256:GMixlA+REf0WwKjCnAlGmS8JPBDdMGzhsCcy8hc3hHo.

Add correct host key in /Users/username/.ssh/known_hosts to get rid of this
message.

Offending ED25519 key in /Users/username/.ssh/known_hosts:100

RSA host key for 45.55.28.29 has changed and you have requested strict
checking.

Host key verification failed.

fatal: [45.55.28.29]: FAILED! => {"changed": false, "failed": true}
```

Ansible's typical message doesn't help identify host key change:
```
TASK [setup] *******************************************************************
System info:
  Ansible 2.2.1.0; Darwin
  Trellis at "Accommodate template inheritance for nginx confs"
SSH Error: data could not be sent to the remote host. Make sure this host can
be reached over ssh
fatal: [45.55.28.29]: UNREACHABLE!
```

------

**First commit.** Changes in `remote-user/tasks/main.yml` are simpler than they appear in the diff:
* adds the task "Warn about change in host keys"
* adds a conditional for first task
* moves the final 3 tasks into a `block` with a new conditional (but the 3 tasks are unchanged)

The first commit also adds the connection test to `deploy.yml` and `rollback.yml`.

**Second commit.** Just renames the `remote-user` role to a more generic name `connection`, now that it is dealing with SSH connection generally, vs. just the `remote_user`.

-----

This only functions with Ansible 2.2+ but doesn't cause problems with 2.0.2.0 or 2.1.1.0


fullyint:
Coming back to this after a couple weeks, with fresh eyes, I still think it is a good idea and I'm comfortable with the implementation.

Don't worry that the pythonic `.replace()` and ansible filter `regex_replace()` risk failures. If the strings they try to match are absent, there is no failure. The task would just print all the output, instead of a selected portion.

In any case, the matching strings have been stable in OpenSSH for roughly 5+ years.
[6.0](https://github.com/openssh/openssh-portable/releases?after=V_6_1_P1) (Apr 19, 2012) -- [7.4](https://github.com/openssh/openssh-portable/releases?after=V_7_5_P1) (Dec 18, 2016)
* [6.0](https://github.com/openssh/openssh-portable/blob/V_6_0_P1/sshconnect.c#L1241) -- [7.4](https://github.com/openssh/openssh-portable/blob/V_7_4_P1/sshconnect.c#L1466) `REMOTE HOST IDENTIFICATION HAS CHANGED`
* [6.0](https://github.com/openssh/openssh-portable/blob/V_6_0_P1/sshconnect.c#L1248) -- [7.4](https://github.com/openssh/openssh-portable/blob/V_7_4_P1/sshconnect.c#L1473) `Please contact your system administrator.`
* [6.0](https://github.com/openssh/openssh-portable/blob/V_6_0_P1/sshconnect.c#L1246) -- [7.4](https://github.com/openssh/openssh-portable/blob/V_7_4_P1/sshconnect.c#L1471) `The fingerprint for the`
* [6.0](https://github.com/openssh/openssh-portable/blob/V_6_0_P1/sshconnect2.c#L101) -- [7.4](https://github.com/openssh/openssh-portable/blob/V_7_4_P1/sshconnect2.c#L99) `Host key verification failed.`

swalkinshaw:
🚀

Issue #749, opened 2017-02-09
Use wordpress_sites.yml.example and vault.yml.example files to make tracking changes with upstream easier.


apmarshall:
## Feature Request

Having read through previous issues, I'm aware that #76 and #120 address similar suggestions to this one, but it's been a while, the file structure has changed, and I think we can accomplish what those proposals were getting at with a much more narrow approach. 

The previous issues/pull requests required removing everything under group_vars/ from version control except for example templates. That would have added a lot of extra work to apply configuration updates pulled down from roots/trellis (copying and pasting those changes from the .example template into the live template), which in a way defeated the point of those changes (to make it easier to keep in sync with roots/trellis. Since then, the configuration options under group_vars have been split out into much more focused files that allows us to be more precise/specific about which files we are moving to example templates:
  - group_vars/environment/wordpress_sites.yml
  - group_vars/environment/vault.yml
  - group_vars/all/vault.yml
  - group_vars/all/mail.yml

These are the files that a user updates with their site-specific variables. They are already populated with dummy/example information that has to be overwritten by the user on set-up, so moving them to file.yml.example doesn't change much about the set-up process for a given user (you need to copy over the sample templates once on set-up). And since these files contain no universally applied logic/constants, just variables set by the user, they are somewhat protected from future updates: ostensibly variable names aren't going to be changed in future versions, and if new site-specific configuration options are added (hypothetically, let's say, there's a variable to enable or disable the wp-api that's added to Trellis someday), sensibly set defaults can ensure that old sites aren't "broken" by not having a variable set in wordpress_sites.yml.

Potential benefits:
1. Users can pull down changes from the roots/trellis repository, including changes to the sample templates (to see new configuration options) without worrying about overriding their local configs.
2. We prevent accidental committing of site configuration data to a public repository. Yes, I know there is git encrypt and Ansible vault, but when its this easy to offer an added layer of safety to users, why not?
3. (optional) Sample templates can be built out more elaborately to provide in-code documentation of all the configuration options available to users.

[...]

PR #750, opened 2017-02-09
Use wordpress_sites.yml.example and vault.yml.example to make tracking changes from roots/trellis easier

apmarshall:
See #749


swalkinshaw:
See https://github.com/roots/trellis/issues/749

PR #744, opened 2017-01-29
Use more secure sshd defaults

fullyint:
### Background

[A discourse thread](https://discourse.roots.io/t/do-we-need-ssh-rc4-cipher-enabled/6438/6) prompted #561, a call to create more secure sshd defaults. @RiFi2k implemented these improvements in #702 in a `ssh-hardening` branch that @fullyint augmented and PR-ed to `master` in #716. Discussion in #716 determined that we should further modify our adaptation of the [dev-sec/ansible-ssh-hardening](https://github.com/dev-sec/ansible-ssh-hardening) role to match Trellis coding conventions. Many thanks to @RiFi2k and @swalkinshaw for their work preparing these modifications.

In the end, whittling and modifying the dev-sec role ended up being more work and code change than if we'd simply adjusted our existing sshd role. In addition, the final modification of the dev-sec role still struck me as somewhat inorganic to the rest of Trellis (example of [nearly final modification here](https://github.com/fullyint/trellis/tree/ssh-hardening/roles/ssh-hardening)).

### This PR

This PR starts over with just our original sshd role, eliminates parts we don't need, then adds the minimum needed to achieve great security and excellent README guidance for users. It strikes me as more streamlined and suitable than #716.

This PR takes one logical step per commit, for easier review. Commits could be squashed on merge.

Sorry @RiFi2k that this didn't directly use and credit all your incredible work. I hope it's ok that I cherry-picked and added your "Remove insecure lines from moduli file" commit (with you as author).

### Comparing with #716

Comments in #716 offered a [diff of configs](https://gist.github.com/fullyint/15a7eb72916a4f9384a5c7afe52549a1) before vs. after #716. Here is how the final config of this PR differs from the final config of #716:

```diff
  # ssh_config
  # ---------------------

  # stribika recommends 'no'
- challengeresponseauthentication yes
+ challengeresponseauthentication no


  # sshd_config
  # ---------------------

  # we already have PasswordAuthentication no
  # keeping 'yes/no' avoids having to change name/handling of sshd_permit_root_login var
- permitrootlogin without-password
+ permitrootlogin yes

  # Applies to RhostsRSAAuthentication and HostbasedAuthentication which are both 'no'
  # cleaner to omit IgnoreUserKnownHosts option (system default is 'no')
- ignoreuserknownhosts yes
+ ignoreuserknownhosts no

  # KerberosAuthentication is 'no'
  # cleaner to omit KerberosAuthentication option (system default is 'yes')
- kerberosorlocalpasswd no
[...]

PR #751, opened 2017-02-10
Add basic git repo host keys to known_hosts

fullyint:
https://github.com/roots/trellis/pull/475#issuecomment-217995557 describes how to populate a remote server's `known_hosts` via deploy hook. The purpose is to avoid the `host key verification failed` error during the composer install task.

This PR makes the `known_hosts` handling part of Trellis core:
* avoids the error for the Vagrant VM (`wordpress-install` role) too
* users no longer need the deploy hook

I included both `rsa` and `ed25519` key types for GitLab. These work with current master and with the `HostKeyAlgorithms` that #744 will apply to the SSH client. I avoided the `ecdsa` type for GitLab because it will be disallowed after #744. 

GitHub and Bitbucket seem to only offer the `rsa` type, before and after #744.

Docs in roots/docs#78

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Add basic git repo host keys to `known_hosts` ([#751](https://github.com/roots/trellis/pull/751))
 * Accommodate template inheritance for nginx confs ([#740](https://github.com/roots/trellis/pull/740))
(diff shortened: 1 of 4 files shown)


swalkinshaw:
Good stuff 🚀

swalkinshaw:
changelog and :shipit:

PR #752, opened 2017-02-11
Use ipify_facts module instead of lookup plugin for IP lookup

fullyint:
>Lookup plugins are lazy evaluated and this may occur at multiple points in time. _[--reference](https://github.com/ansible/ansible/issues/9623#issuecomment-64452130)_

This PR changes the IP lookup to use the command module instead of a lookup.
* This ensures only a single connection to api.ipify.org (e.g., because no connection should be made during deploys, as unfortunately occurred in [this example](https://discourse.roots.io/t/deploys-on-a-slow-connection/8785)). This should reduce how often users have connectivity issues with api.ipify.org.
* It also avoids confusion of potential failures at random points during the playbook ([example](https://discourse.roots.io/t/ansible-fails-at-ensure-fail2ban-is-configured/8698)).

Users can disable the api.ipify.org connection by adding a new variable to group_vars:
`ssh_client_ip_lookup: false`

____

This PR accounts for how a `vagrant ssh` connection to the VM uses the IP in `ansible_default_ipv4.gateway` (e.g., `10.0.2.2`) instead of the public IP from api.ipify.org.

____

I was disappointed that the [`ansible_env["SSH_CLIENT"]`](http://stackoverflow.com/a/33961318/6847025) fact is not available when the play parameter `become: yes` is used. Otherwise, we could have avoided the need to query api.ipify.org and used the IP retrieved from `ansible_env` instead. The `ansible_env` var  is populated during the setup task. The ad hoc commands below demonstrate how `ansible_env` differs when `become: yes`.

```
# SSH_CLIENT is present in output
ansible 138.68.53.198 -m setup -u admin

# `-b` is become: yes, SSH_CLIENT is absent from output
ansible 138.68.53.198 -m setup -b -u admin -K
```

### DIFF
--- a/group_vars/all/main.yml
+++ b/group_vars/all/main.yml
@@ -6,7 +6,7 @@ ntp_timezone: Etc/UTC
 ntp_manage_config: true
 www_root: /srv/www
 ip_whitelist:
-  - "{{ (env == 'development') | ternary(ansible_default_ipv4.gateway, ssh_client_ip.stdout | default('')) }}"
+  - "{{ (env == 'development') | ternary(ansible_default_ipv4.gateway, ipify_public_ip | default('')) }}"
 
 # Values of raw_vars will be wrapped in `{% raw %}` to avoid templating problems if values include `{%` and `{{`.
 # Will recurse dicts/lists. `*` is wildcard for one or more dict keys, list indices, or strings. Example:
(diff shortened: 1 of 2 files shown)


nathanielks:
Good work, @fullyint 👍

fullyint:
@swalkinshaw prompted me to take a second look at the `ipify_facts` module.

I added a commit changing from the `command` module with `curl` to the `ipify_facts` module.  I also added `connection: local` to ensure it retrieves the IP for the local machine IP, not the remote host.

swalkinshaw:
Much better 👍

PR #754, opened 2017-02-13
Disable multisite constants while checking wp core is-installed

### LINKED ISSUE #554: Multisite is-installed deploy check still failing
I belive a releated fix was provided in [#482](https://github.com/roots/trellis/pull/543/commits)

However, running `./deploy.sh staging site1` with the latest version of Trellis fails at `TASK [deploy : WordPress Installed?]` resulting in this database error:

```
WordPress database error Table 'site1_staging.wp_blogs' doesn't exist for query SELECT * FROM wp_blogs WHERE domain = 'staging.site1.com' AND path = '/' made by include('phar:///usr/bin/wp/php/boot-phar.php'), include('phar:///usr/bin/wp/php/wp-cli.php'), WP_CLI\Runner->start, WP_CLI\Runner->load_wordpress, require('phar:///usr/bin/wp/php/wp-settings- cli.php'), require('wp-includes/ms-settings.php'), get_site_by_path WordPress database error Table 'site1_staging.wp_sitemeta' doesn't exist for query SELECT meta_value FROM wp_sitemeta WHERE meta_key = 'site_name' AND site_id = 1 made by include('phar:///usr/bin/wp/php/boot-phar.php'), include('phar:///usr/bin/wp/php/wp-cli.php'), WP_CLI\Runner->start, WP_CLI\Runner->load_wordpress, require('phar:///usr/bin/wp/php/wp-settings- cli.php'), require('wp-includes/ms-settings.php'), WP_Network->__construct, WP_Network->_set_site_name, get_network_option WordPress database error Table 'site1_staging.wp_sitemeta' doesn't exist for query SELECT meta_value FROM wp_sitemeta WHERE meta_key = 'siteurl' AND site_id = 1 made by include('phar:///usr/bin/wp/php/boot-phar.php'), include('phar:///usr/bin/wp/php/wp-cli.php'), WP_CLI\Runner->start, WP_CLI\Runner->load_wordpress, require('phar:///usr/bin/wp/php/wp-settings- cli.php'), wp_cookie_constants, get_site_option, get_network_option fatal: [site1]: FAILED! => {"changed": false, "cmd": ["wp", "core", "is-installed", "--network"], "delta": "0:00:00.483683", "end": "2016-04-06 19:47:19.801462", "failed": true, "failed_when_result": true, "rc": 1, "start": "2016-04-06 19:47:19.317779", "stderr": "WordPress database error Table 'site1_staging.wp_blogs' doesn't exist for query SELECT * FROM wp_blogs WHERE domain = 'staging.site1.com' AND path = '/' made by include('phar:///usr/bin/wp/php/boot-phar.php'), include('phar:///usr/bin/wp/php/wp-cli.php'), WP_CLI\\Runner->start, WP_CLI\\Runner->load_wordpress, require('phar:///usr/bin/wp/php/wp-settings-cli.php'), require('wp-includes/ms-settings.php'), get_site_by_path\nWordPress database error Table 'site1_staging.wp_sitemeta' doesn't exist for query SELECT meta_value FROM wp_sitemeta WHERE meta_key = 'site_name' AND site_id = 1 made by include('phar:///usr/bin/wp/php/boot-phar.php'), include('phar:///usr/bin/wp/php/wp-cli.php'), WP_CLI\\Runner->start, WP_CLI\\Runner->load_wordpress, require('phar:///usr/bin/wp/php/wp-settings-cli.php'), require('wp-includes/ms-settings.php'), WP_Network->__construct, WP_Network->_set_site_name, get_network_option\nWordPress database error Table 'site1_staging.wp_sitemeta' doesn't exist for query SELECT meta_value FROM wp_sitemeta WHERE meta_key = 'siteurl' AND site_id = 1 made by include('phar:///usr/bin/wp/php/boot-phar.php'), include('phar:///usr/bin/wp/php/wp-cli.php'), WP_CLI\\Runner->start, WP_CLI\\Runner->load_wordpress, require('phar:///usr/bin/wp/php/wp-settings-cli.php'), wp_cookie_constants, get_site_option, get_network_option", "stdout": "", "stdout_lines": [], "warnings": []}
```

**group_vars/staging/wordpress_sites.yml**:

```
wordpress_sites:
  site1:
    cache:
      enabled: false
    env:
      domain_current_site: staging.site1.com
      wp_home: https://staging.site1.com
      wp_siteurl: https://staging.site1.com/wp
    local_path: ../site1
    multisite:
      enabled: true
      subdomain: false
    repo: git@github.com:redacted/site1.git
    site_hosts:
      - staging.site1.com
    ssl:
      enabled: false
      provider: letsencrypt
```

A temporary workaround has been to ssh into the server, navigate to `/srv/www/site1/current/` and install Wordpress using WP-CLI:

```
wp core multisite-install --allow-root --title="Site 1" --admin_user="admin" --admin_password="password" --admin_email="admin@site1.com"
```

fullyint:
Fixes #554: `WordPress database error Table 'name' doesn't exist` 
The error happens on a first deploy of multisite -- before multisite is installed.

>When `MULTISITE=>true` and the database tables don't yet exist, WordPress will error before WP-CLI can execute  _[--reference](https://github.com/wp-cli/wp-cli/issues/3641#issuecomment-265172773)_

The [`mysql_db` module](http://docs.ansible.com/ansible/mysql_db_module.html) doesn't allow checking presence of tables as far as I could tell. I didn't manage to come up with any satisfactory CLI commands to inline a mysql query or php command to check the DB tables. Suggestions welcome.

This PR dynamically comments out the `MULTISITE` and `SUBDOMAIN_INSTALL` constants, allows WP-CLI to check `wp core is-installed`, then immediately re-enables the constants.

I removed the `--network` option from the `is-installed` command because it yields `RC: 1` when the multisite constants are absent. (This PR makes the constants absent during the `is-installed` check.)

This PR assumes the location of multisite constants in `config/application.php`, a bedrock directory structure. If that is not a safe assumption, I suppose we could add a task that checks whether the file exists then make conditional the tasks that disable the multisite constants.

### DIFF
--- a/roles/deploy/hooks/finalize-after.yml
+++ b/roles/deploy/hooks/finalize-after.yml
@@ -1,11 +1,32 @@
 ---
+- name: Temporarily disable multisite constants
+  lineinfile:
+    dest: "{{ deploy_helper.current_path }}/config/application.php"
+    line: //\1
+    regexp: ([^/]+['"]{{ item }}["'].+)
+    backrefs: yes
+  with_items:
+    - MULTISITE
+    - SUBDOMAIN_INSTALL
+  register: multisite_constants
+  when: project.multisite.enabled
+
 - name: WordPress Installed?
-  command: wp core is-installed {{ project.multisite.enabled | default(false) | ternary('--network', '') }}
+  command: wp core is-installed
   args:
     chdir: "{{ deploy_helper.current_path }}"
   register: wp_installed
   changed_when: false
   failed_when: wp_installed.stderr != ""
+
+- name: Re-enable multisite constants
+  lineinfile:
+    dest: "{{ deploy_helper.current_path }}/config/application.php"
+    line: \1
+    regexp: ^//{{ item.invocation.module_args.regexp }}
+    backrefs: yes
+  with_items: "{{ multisite_constants.results }}"
+  when: project.multisite.enabled and item | changed
 
 - block:
   - name: Update WP database
--- a/roles/deploy/hooks/finalize-after.yml
+++ b/roles/deploy/hooks/finalize-after.yml
@@ -1,6 +1,11 @@
 ---
+- name: Create file with multisite constants defined as false
+  template:
+    src: "tmp_multisite_constants.php.j2"
+    dest: "{{ deploy_helper.shared_path }}/tmp_multisite_constants.php"
+
 - name: WordPress Installed?
-  command: wp core is-installed {{ project.multisite.enabled | default(false) | ternary('--network', '') }}
+  command: wp core is-installed --require={{ deploy_helper.shared_path }}/tmp_multisite_constants.php
   args:
     chdir: "{{ deploy_helper.current_path }}"
   register: wp_installed
--- a/roles/deploy/hooks/finalize-after.yml
+++ b/roles/deploy/hooks/finalize-after.yml
@@ -1,6 +1,6 @@
 ---
 - name: WordPress Installed?
-  shell: MULTISITE=false SUBDOMAIN_INSTALL=false wp core is-installed
+  command: wp core is-installed {{ project.multisite.enabled | default(false) | ternary('--network', '') }}
   args:
     chdir: "{{ deploy_helper.current_path }}"
   register: wp_installed
--- a/roles/deploy/hooks/finalize-after.yml
+++ b/roles/deploy/hooks/finalize-after.yml
@@ -1,6 +1,6 @@
 ---
 - name: WordPress Installed?
-  command: wp core is-installed {{ project.multisite.enabled | default(false) | ternary('--network', '') }}
+  shell: MULTISITE=false SUBDOMAIN_INSTALL=false wp core is-installed
   args:
     chdir: "{{ deploy_helper.current_path }}"
   register: wp_installed
--- a/roles/deploy/hooks/finalize-after.yml
+++ b/roles/deploy/hooks/finalize-after.yml
@@ -1,32 +1,11 @@
 ---
-- name: Temporarily disable multisite constants
-  lineinfile:
-    dest: "{{ deploy_helper.current_path }}/config/application.php"
-    line: //\1
-    regexp: ([^/]+['"]{{ item }}["'].+)
-    backrefs: yes
-  with_items:
-    - MULTISITE
-    - SUBDOMAIN_INSTALL
-  register: multisite_constants
-  when: project.multisite.enabled
-
 - name: WordPress Installed?
-  command: wp core is-installed
+  command: wp core is-installed {{ project.multisite.enabled | default(false) | ternary('--network', '') }}
   args:
     chdir: "{{ deploy_helper.current_path }}"
   register: wp_installed
   changed_when: false
   failed_when: wp_installed.stderr != ""
-
-- name: Re-enable multisite constants
-  lineinfile:
-    dest: "{{ deploy_helper.current_path }}/config/application.php"
-    line: \1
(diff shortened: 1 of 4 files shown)


swalkinshaw:
Would this help? http://wp-cli.org/commands/db/query/

fullyint:
The problem is that we can't use wp-cli because it can't even load. Under these conditions WordPress fails before wp-cli can do anything.

QWp6t:
if those were environment variables (i.e., in `.env`), you could easily override by setting the env var before loading wp-cli since phpdotenv doesn't override existing environment variables.

then the user's directory structure wouldn't matter, but the user wouldn't be able to hard-code those values.

:thinking:

nathanielks:
I've also run into this. My hack was to install WP in the `finalize-before` step

fullyint:
I've updated this PR but also posted an alternative in #765. Both follow @QWp6t's suggestion:
>if those were environment variables (i.e., in `.env`), you could easily override by setting the env var before loading wp-cli since phpdotenv doesn't override existing environment variables.
>
>then the user's directory structure wouldn't matter, but the user wouldn't be able to hard-code those values.

Both PRs expect that users have added these lines to `site/config/application.php`:
```
define('MULTISITE', env('MULTISITE') ?? true);
define('SUBDOMAIN_INSTALL', env('SUBDOMAIN_INSTALL') ?? true); // Set to false if using subdirectories
```

This PR retains the regular `multisite.enabled` and `multisite.subdomains` settings in `wordpress_sites`, using those values to create corresponding default env vars. #765 drops the former settings, consolidating all multisite config in the env vars only.

This PR added a default [`path_current_site` env var](https://github.com/wp-cli/wp-cli/blob/b8c50b8742c3febc27f252bcdffbd0a7f01c7594/php/commands/core.php#L617) based on [`multisite.base_path`](https://github.com/roots/trellis/blob/019acbcba163322ffe1b05ba17581bc921e9b0e5/roles/wordpress-install/tasks/main.yml#L40) because it looked like those two values should be aligned.

roles/deploy/hooks/finalize-after.yml
```diff
@@ -1,6 +1,6 @@
 ---
 - name: WordPress Installed?
-  command: wp core is-installed {{ project.multisite.enabled | default(false) | ternary('--network', '') }}
+  shell: MULTISITE=false SUBDOMAIN_INSTALL=false wp core is-installed
```
swalkinshaw: is `shell` needed for ENV vars?

Ansible does support `environment`: http://docs.ansible.com/ansible/playbooks_environment.html

nathanielks:
I do believe so as `command` is not run in a shell environment. You have to
pass them into the task itself

fullyint:
Thank you @swalkinshaw and others for the review, and @swalkinshaw for the idea that wp-cli could `--require` a file with temporary overrides for the multisite constants. This `require` loads before WP loads, so it prevents the problem that WP won't load when multisite constants are `true`.

I've reverted again 😅  and added a commit with just this `--require` strategy. Of course, we'll need to squash before merging. 

I added `error_reporting(E_ALL & ~E_NOTICE);` to the `--require` file to avoid having the `is-installed` task show in red `PHP Notice:  Constant MULTISITE already defined` (even though task status is `ok`).

roles/deploy/hooks/finalize-after.yml
```diff
@@ -1,6 +1,11 @@
 ---
+- name: Create file with multisite constants defined as false
+  template:
```
swalkinshaw: This can just be a plain file with a `copy` right?

fullyint:
Latest edits submitted in fresh uncluttered PR: #766

PR #757, opened 2017-02-15
Parallels tools MUST BE installed first time that vagrant is run.

jawngee:
If you are using Parallels with vagrant, the initial vagrant up on a new trellis install will fail because Parallels Guest Tools aren't installed, which means that the NFS mount will fail.


swalkinshaw:
👍  thanks

PR #761, opened 2017-02-16
Enable admin_user to invoke sudo with xdebug-tunnel.yml

fullyint:
When `sshd_permit_root_login: false` the `admin_user` will connect to run the `xdebug-tunnel.yml` playbook. The `admin_user` needs to invoke `sudo` in order to install and configure Xdebug.


nathanielks:
yup, forgot to add that 👍

swalkinshaw:
👍

PR #762, opened 2017-02-16
Allow for per-project packagist.com authentication

nathanielks:
This enables the use of a per-site packagist.com token in `wordpress_sites.yml`:
```
wordpress_sites:
  example.com:
    packagist_token: "{{ vault_example_com_packagist_token }}"
  foobar.com:
    packagist_token: "{{ vault_foobar_com_packagist_token }}"
```

If the token is present, it will configure that site on each deploy with the token before installing composer dependencies.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Allow for per-project packagist.com authentication ([#762](https://github.com/roots/trellis/pull/762))
 * Forward extra bin/deploy.sh parameters to ansible-playbook ([#748](https://github.com/roots/trellis/pull/748))
 * Update WP-CLI to 1.1.0 ([#759](https://github.com/roots/trellis/pull/759))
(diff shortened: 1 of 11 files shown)


swalkinshaw:
Shouldn't this use a vault var?

nathanielks:
The idea would be:
```
wordpress_sites:
  example.com:
    packagist_token: "{{ vault_example_com_packagist_token }}"
  foobar.com:
    packagist_token: "{{ vault_foobar_com_packagist_token }}"
```

swalkinshaw:
Oops, correct. Was just going by your example code which I assumed was from `wordpress_sites.yml` and not from `vault.yml`.

I'd just update that to expand the example.

nathanielks:
Description updated!

swalkinshaw:
This looks good but can we do it without the two "helper" vars. They aren't needed and we try and avoid them unless it really helps.

swalkinshaw:
👏

PR #764, opened 2017-02-17
Permit TCP port forwarding for 127.0.0.1:3306

fullyint:
#744 changed `AllowTcpForwarding` from `yes` to `no`. This blocks tools like Sequel Pro from being able to access the DB on port 3306 (example [discourse thread](https://discourse.roots.io/t/error-connecting-to-vagrant-db-with-sequel-pro/8835)).

The PR allows local port forwarding and limits it to the host `127.0.0.1` and port `3306`, re-enabling DB access for tools like Sequel Pro.


swalkinshaw:
:shipit:

PR #767, opened 2017-02-20
Enable ability to change users role `update_password` value

nathanielks:
This is helpful in the event there are multiple users on the server instead of `web` and `admin`. I don't want to know what a another user's password is, so being able to change `update_password` to `on_create` gives me the ability to bootstrap a user with a known password and then that person the ability to go in and change it with the confidence it won't be overwritten the next time the playbook is applied.

Perhaps we even allow an `update_password` field in the `users` dictionary that specifies what each user's setting is?
```yaml
users:
  - name: "{{ web_user }}"
    groups:
      - "{{ web_group }}"
    keys:
      - "{{ lookup('file', '~/.ssh/id_rsa.pub') }}"
      # - https://github.com/username.keys
  - name: "nathanielks"
    update_password: on_create
    groups:
      - sudo
    keys:
      - "{{ lookup('file', '~/.ssh/id_rsa.pub') }}"

```

Thoughts @fullyint @swalkinshaw?


nathanielks:
I updated it so the setting can change per-user

roles/users/tasks/main.yml
```diff
@@ -36,7 +36,7 @@
     password: '{% for user in vault_users | default([]) if user.name == item.name and user.password is defined %}{{ user.password | password_hash("sha512", user.salt | default("") | truncate(16, true, "") | regex_replace("[^\.\/a-zA-Z0-9]", "x")) }}{% else %}{{ None }}{% endfor %}'
     state: present
     shell: /bin/bash
-    update_password: always
+    update_password: "{{ item.value.update_password | default('always') }}"
   with_items: "{{ users }}"
```
fullyint: Given the `with_items` loop (not `with_dict`), I believe it should just be
`item.update_password`
(vs. `item.value.update_password`)

nathanielks:
Hmm, just realized an issue. If we set the user's passwords to `on_create`, then https://github.com/roots/trellis/blob/master/roles/remote-user/tasks/main.yml#L25-L29 will cause problems when the user's password is different and ansible will fail. Should I add a provision in that conditional if the user's `update_password != 'always'`?

fullyint:
When people run `server.yml` with an out-of-date `admin_user` password in `group_vars/production/vault.yml`, I believe they will be forced to use `-K, --ask-become-pass` and manually enter a password at the prompt, causing [`cli_ask_become_pass`](https://github.com/roots/trellis/blob/be80f822f831a28028d44ff91474f338ff665b4f/lib/trellis/plugins/callback/vars.py#L94) to be `true`. The conditional for the task you pointed out already includes `not cli_ask_become_pass`, so I believe that task will skip in this scenario.

nathanielks:
I'll test to verify 👍

group_vars/development/wordpress_sites.yml
```diff
@@ -5,7 +5,7 @@
 wordpress_sites:
   example.com:
     site_hosts:
-      - canonical: example.dev
+      - canonical: exampletrellis.dev
```
JulienMelissas: did we mean for this to be changed?

kalenjohnson: https://github.com/roots/trellis/pull/767/commits/061b8fb92cbf04394f9f758327a40d26980a4cfb

nathanielks: good eye though @JulienMelissas!

JulienMelissas: durp, sorry guys I'll shut up

nathanielks: nah!

swalkinshaw:
Is this good to go (other than conflict)?

partounian:
I would love to have this is in

nathanielks:
Conflicts resolved!

PR #770, opened 2017-02-22
Secure cookies default settings

samkarpluk:
I recently ran a site based on the standard roots/bedrock/trellis setup against the [Mozilla Observatory](https://observatory.mozilla.org/) security review tool and received a C-. One of the reasons was that httponly and secure cookies weren't set in the php.ini file as follows:

```
session.cookie_httponly = On
session.cookie_secure = On
```
I canvassed the roots discourse about updating the php.ini file and [was encouraged](https://discourse.roots.io/t/php-ini-settings/8855/4) to create a PR here.

After putting these settings in place the score in the Observatory scan went up to a B.


swalkinshaw:
Thanks for contributing @samkarpluk :)

I realized that we can't default to secure though. This only works if someone is using HTTPS which isn't guaranteed. `HttpOnly` we can definitely keep as a default like this though.

samkarpluk:
@swalkinshaw right yeah, https is an option so that makes sense. It could still be set with as `php_session_secure_cookie: 'Off'` as a default, which still gives folks the ability to set it to secure if they are https.

swalkinshaw:
Yeah, agreed. Still useful to be there. Want to change it to `Off`? We can maybe try and figure out a smart way of setting it automatically later.

roles/php/templates/php.ini.j2
```diff
@@ -11,6 +11,8 @@ memory_limit = {{ php_memory_limit }}
 post_max_size = {{ php_post_max_size }}
 sendmail_path = {{ php_sendmail_path }}
 session.save_path = {{ php_session_save_path }}
+session.cookie_httponly = {{php_session_cookie_httponly}}
+session.cookie_secure = {{php_session_cookie_secure}}
```
swalkinshaw: Can both of these get spaces surrounding them like the others?

swalkinshaw:
Thanks 🎉

PR #782, opened 2017-03-01
Always set_fact for ansible_user, even if CLI --check or --user

### LINKED ISSUE #779: Improve error handling with "Set remote user for each host"

- [x] This is a bug report

---

When ansible logs an error (stdout), task "Set remote user for each host" tries to use the field "root_status.stdout_lines" which isn't set in this situation (root_status was in this case `{u'msg': u'remote module (command) does not support check mode',
u'skipped': True, u'changed': False}`). resulting in a confusing error that requires debugging for finding out the actual underlying issue:
````
the field 'args' has an invalid value, which appears to include a variable
that is undefined. The error was: Unable to look up a name or access an
attribute in template string ({{ ('root' in root_status.stdout_lines) |
ternary('root', admin_user) }}).
Make sure your variable name does not contain invalid characters like '-':
argument of type 'StrictUndefined' is not iterable

The error appears to have been in
'[...]/trellis/roles/remote-user/tasks/main.yml':
line 16, column 3, but may

- name: Set remote user for each host
  ^ here

fatal: [web]: FAILED! => {"failed": true}
````

The task doesn't cause an error but allows to proceed so the underlying issue is logged instead.

---

## Bug report

(delete this section if not applicable)

**Please provide steps to reproduce, including full log output:**

Run the server trellis playbook in a way that causes ansible to complain e.g. about the passed arguments (passing --check when using the remote module (command)).

**Please describe your local environment:**

Ansible version: 2.2.1.0

OS: Ubuntu 16.04.2 LTS

Vagrant version: Vagrant 1.9.1 (though not applicable in this case as baremetal should be provisioned)

**Where did the bug happen? Development or remote servers?**

Remote (though ansible aborted because of the error)

## Further information:

Maybe it would be a good idea adding an appropriate check_mode field to 
the remote commands so a check run is possible with the trellis playbook.

### PR DESCRIPTION
**Pull request by fullyint**
Fixes #779 -- thank you for the report, @strarsis

In check mode, remote user test and assignment skips. No `ansible_user` is set despite being needed.

This PR adds [`check_mode: no`](http://docs.ansible.com/ansible/playbooks_checkmode.html#enabling-or-disabling-check-mode-for-tasks), forcing the relevant `remote-user` tasks to run in normal mode even when the playbook is running in check mode (e.g., via `--check`).

Note that [check mode usually doesn't make sense for the first playbook run](http://blog.nordeus.com/dev-ops/ansible-check-mode-tips.htm) of `server.yml`.

----

Testing this PR revealed another problem addressed in ansible/ansible#22156. Ansible's magic var for `ansible_user` is apparently not available for consumption in explicit redefinition of the `ansible_become_pass` magic var.

For example, if you pass `ansible-playbook -u admin`, the Trellis `set_fact` task for `ansible_user` skips because `ansible_user` **is already defined**, as a magic var in this case. Trellis explicitly defines  `ansible_become_pass` (normally a magic var) using `ansible_user`, but an error occurs because `ansible_user` **is somehow `undefined` in this unique context**:
```
TASK [Install Python 2.x] ******************************************************
System info:
  Ansible 2.2.0.0; Darwin
  Trellis at "Allow for per-project packagist.com authentication"
the field 'become_pass' has an invalid value, which appears to include a
variable that is undefined. The error was: 'ansible_user' is undefined
fatal: [test]: FAILED! => {"failed": true}
```

The second part of this PR prevents the error above by causing the `set_fact` for `ansible_user` to always run, ensuring that `ansible_user` is always **explicitly defined** (not just a magic var).

### DIFF
--- a/roles/remote-user/tasks/main.yml
+++ b/roles/remote-user/tasks/main.yml
@@ -10,13 +10,14 @@
   local_action: command ansible {{ inventory_hostname }} -m raw -a whoami -u root {{ cli_options | default('') }}
   failed_when: false
   changed_when: false
+  check_mode: no
   register: root_status
   tags: [connection-tests]
 
 - name: Set remote user for each host
   set_fact:
-    ansible_user: "{{ ('root' in root_status.stdout_lines) | ternary('root', admin_user) }}"
-  when: ansible_user is not defined
+    ansible_user: "{{ ansible_user | default(('root' in root_status.stdout_lines) | ternary('root', admin_user)) }}"
+  check_mode: no
 
 - name: Announce which user was selected
   debug:


swalkinshaw:
👍

PR #785, opened 2017-03-10
Integrate Content Security Policy

corbinhesse:
[Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) is a large part of a security workflow for websites, and it is only becoming more popular with CSP3 coming soon.

I've added a very simple way to create this nginx header for each site host. This can be easily enabled and configured for each WordPress site (`item.key`). Only a single directive was defined in `group_vars/development/wordpress_sites.yml` file to show how it works.

This would output the following nginx header in `/etc/nginx/sites-available/example.com.conf`

```nginx
add_header Content-Security-Policy "default-src 'none'; ";
```

### DIFF
--- a/group_vars/development/wordpress_sites.yml
+++ b/group_vars/development/wordpress_sites.yml
@@ -17,3 +17,7 @@ wordpress_sites:
       provider: self-signed
     cache:
       enabled: false
+    csp:
+      enabled: false
+      directives:
+        - "default-src 'none'"
--- a/group_vars/production/wordpress_sites.yml
+++ b/group_vars/production/wordpress_sites.yml
@@ -19,3 +19,5 @@ wordpress_sites:
       provider: letsencrypt
     cache:
       enabled: false
+    csp:
+      enabled: false
--- a/group_vars/staging/wordpress_sites.yml
+++ b/group_vars/staging/wordpress_sites.yml
@@ -19,3 +19,5 @@ wordpress_sites:
       provider: letsencrypt
     cache:
       enabled: false
+    csp:
+      enabled: false
--- a/roles/wordpress-setup/templates/wordpress-site.conf.j2
+++ b/roles/wordpress-setup/templates/wordpress-site.conf.j2
@@ -45,7 +45,7 @@ server {
 
   {% endif -%}
   {% endblock -%}
-  
+
   {% block multisite_rewrites -%}
   {% if item.value.multisite.enabled | default(false) -%}
   # Multisite rewrites
@@ -100,6 +100,13 @@ server {
   {% block includes_d -%}
   include includes.d/{{ item.key }}/*.conf;
 
+  {% endblock -%}
+
+  {% block csp -%}
+  {% if item.value.csp is defined and item.value.csp.enabled | default(false) -%}
+  add_header Content-Security-Policy "{% for source in item.value.csp.directives %}{{ source }}; {% endfor %}";
+
+  {% endif -%}
   {% endblock -%}
 
   {% block location_uploads_php -%}


QWp6t:
Does this do anything that can't be done with our current [nginx includes](https://roots.io/trellis/docs/nginx-includes/) solution?

swalkinshaw:
CSP is awesome but I'm not sure Trellis is the right place for it. It's more of an application level concern. So if you're running a WP site, you'd control these headers through a WP plugin for example.

corbinhesse:
The only difference this does compared to the nginx-includes directory is allow for setting different directives within each environment. Allowing for reporting to only be active on production environments or have no directives set on the development environment if it will be a simple policy.

Only reason I submitted a request is to help the generation of these policies. Trellis has made SSL automated, and thinking CSP is another security measure it would be nice to automate as well.

On complex policies they can get really difficult to read as an nginx header; therefore, each directive can be separated into a single line like the following:

```yml
csp:
  enabled: true
  directives:
    - "default-src 'none'"
    - "script-src 'self' https://www.googletagmanager.com/gtm.js https://www.google-analytics.com 'sha256-lxi5DtuADDE7N15mfVx3ZWBrGUn/HD+MUoBiUE0cKxw='"
    - "style-src 'self' 'unsafe-inline'"
    - "block-all-mixed-content"
    - "require-sri-for script style"
```

And with that example you cannot do with WordPress plugins as they do not allow some of those directives. These plugins are usually set with `<meta>` tags in which IE doesn't support.

No skin on my back if it's declined. Just wanted to help if it would be useful to the group. Thanks for taking the time to look at it!

swalkinshaw:
> And with that example you cannot do with WordPress plugins as they do not allow some of those directives. These plugins are usually set with <meta> tags in which IE doesn't support.

Are you talking about existing plugins? When I said plugins, I even just meant a custom plugin someone writes to set specific CSP directives they want.

I agree the existing plugins are kind of crappy.

I'm not necessarily against this as a way to quickly set some CSP rules. I just think once you start getting into more complex ones they belong in the application because you'll need some logic around it (or state that exists in the app).

For example: in a typical WP site you'll have embeds in posts/pages, with CSP in the app you could add some meta field for a single post only to allow it (instead of site wide).

corbinhesse:
> I agree the existing plugins are kind of crappy.

Yes, I was only referring to existing plugins.

> I'm not necessarily against this as a way to quickly set some CSP rules. I just think once you start getting into more complex ones they belong in the application because you'll need some logic around it (or state that exists in the app).

Yep, I agree. This proposal **or** including them in `nginx-includes` would not be satisfactory for complex page level directives.

swalkinshaw:
@corbinhesse thanks for the work on this but we're going to pass on it. When we're really on the fence about something, we tend to default to not adding it.

This can easily be done through Nginx includes which we like to encourage. This might be something we'd do if we really thought it would encourage people to set CSP directives, but I don't think that will happen. Especially because we wouldn't want to have the `csp` dict in the configs by default anyway so they'd be discoverable by some docs.

So, if you're interested, we could definitely add a section to our docs explaining a little about CSP and how to set directives.

swalkinshaw:
Oh and simple Nginx includes confs can't vary by environment like this feature can, but you could use child templates to accomplish it. It's definitely not as simple or nice though.

corbinhesse:
Not a problem @swalkinshaw! No explanation necessary so I appreciate the response. I just want more sites to be secure.

I also didn't intend on those `csp` dict in the `wordpress_sites.yml` file by default. It was more of a presentational purpose. It could be removed similar and act as false by default.

Regardless, I'll look into the docs repository and see if I can write some quick notes on CSP in the `security` or `nginx-includes` file.

Thanks!

PR #735, opened 2017-01-16
Customize installed Apt packages

nathanielks:
This adds a few new variables:

- `apt_packages_default`
- `apt_packages_custom `
- `apt_packges_install`

If a user wishes to install additional apt packages, they simply need to add items to the `custom_apt_packages` variable:

```yaml
apt_packages_custom:
  - nfs-utils
```

PR #787, opened 2017-03-11
Allow customization of PHP extensions

swalkinshaw:
Adds a `php_extensions_custom` variable to allow users to specific custom extensions to be installed.

`php_extensions_default` can also be defined to change the default extensions.

This follows the pattern set in https://github.com/roots/trellis/pull/735 but updates the naming.


fullyint:
Looks good to me. That will definitely be a helpful option.

PR #789, opened 2017-03-15
Add additional nginx sites-available templates support

### LINKED ISSUE #786: Allow customizing the default site's Nginx configuration

- [ ] This is a bug report

---

---

## Feature Request

Thanks for your work !!

medfreeman:
closes #786 

Adds a `nginx_sites_available_templates_path` variable
to allow users to specify the custom templates folder
from which all *.conf.j2 files
will be copied to the remote hosts,
default is `nginx-sites-available`.

`nginx_sites_enabled` array can also be defined
to change the additionally enabled sites,
default is
```
nginx_sites_enabled:
  - no-default
```
User can keep or remove trellis' default site configuration.

### DIFF
--- a/roles/nginx/defaults/main.yml
+++ b/roles/nginx/defaults/main.yml
@@ -1,5 +1,12 @@
 ---
 nginx_conf: nginx.conf.j2
+
+nginx_sites_available_templates_path: nginx-sites-available
+nginx_sites_available_pattern: "^({{ nginx_sites_available_templates_path | regex_escape }})/(.*)\\.j2$"
+nginx_sites_available_cleanup: true
+nginx_sites_enabled:
+  - no-default
+
 nginx_path: /etc/nginx
 nginx_logs_root: /var/log/nginx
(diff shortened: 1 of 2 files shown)


medfreeman:
It seems that the `Remove unmanaged files from sites-available` (https://github.com/medfreeman/trellis/blob/b0e2712d934f3b9223067889105accfd2d285bbd/roles/nginx/tasks/main.yml#L122) task will also prune let's encrypt nginx configurations (i've done my testing on a site with a freshly disabled let's encrypt config).

To avoid over-complicating the code, should i instead template the additional sites configurations to a subfolder of sites-available on the hosts ? So i don't need to manage every single exception (wordpress sites templates, let'sencrypt, etc...)

medfreeman:
Aww, i just realized i can create an 'extras' folder in sites-available and sites-enabled, add a directive to nginx.conf template to load files in sites-enabled/extras and manage things in a simple way..
Restarting..

PR #790, opened 2017-03-15
Add additional nginx sites-available templates support

### LINKED ISSUE #786: Allow customizing the default site's Nginx configuration

medfreeman:
closes #786 

Adds a `nginx_sites_available_templates_path` variable
to allow users to specify the custom templates folder
from which all *.conf.j2 files
will be copied to the remote hosts
nginx/sites-available/extras folder,
default is `nginx-sites-available`.

`nginx_sites_enabled` array can also be defined
to change the additionally enabled sites,
default is
```
nginx_sites_enabled:
  - no-default
```
User can keep or remove trellis' default site configuration.

### DIFF
--- a/roles/nginx/defaults/main.yml
+++ b/roles/nginx/defaults/main.yml
@@ -1,5 +1,12 @@
 ---
 nginx_conf: nginx.conf.j2
+
+nginx_sites_available_templates_path: nginx-sites-available
+nginx_sites_available_pattern: "^({{ nginx_sites_available_templates_path | regex_escape }})/(.*)\\.j2$"
+nginx_sites_available_cleanup: true
+nginx_sites_enabled:
+  - no-default
+
 nginx_path: /etc/nginx
 nginx_logs_root: /var/log/nginx
 nginx_user: www-data
--- a/roles/nginx/tasks/main.yml
+++ b/roles/nginx/tasks/main.yml
@@ -50,8 +50,100 @@
     state: absent
   notify: reload nginx
 
-- name: Enable better default site to drop unknown requests
-  command: cp {{ nginx_path }}/h5bp-server-configs/sites-available/no-default {{ nginx_path }}/sites-enabled/no-default.conf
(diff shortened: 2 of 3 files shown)


medfreeman:
This time i simply : 
  - remove h5bp's no-default.conf from sites-enabled
  - create remote sites-available/extras and sites-enabled/extras folders
  - template *.conf.j2 files from local ```nginx_sites_available_templates_path``` to remote sites-available/extras folder
  - prune non-enabled (not present in ```nginx_sites_enabled``` array) sites from sites-enabled/extras
  - prune non-managed (not present in ```nginx_sites_available_templates_path``` except no-default.conf) files from sites-available/extras
  - enable sites present in ```nginx_sites_enabled``` array by creating symlinks from their sites-available/extras counterparts

swalkinshaw:
@medfreeman thanks for this.

I'm really wary of adding this much overhead. Cleaning up the templates is a nice feature, but I don't know if we want to keep up that pattern in multiple places.

Here's the minimal feature I had in mind: https://github.com/roots/trellis/compare/nginx-sites-available-templates?expand=1

note: I'm perfectly fine having the conf template in the repo. It hasn't changed since 2013.

What do you think?

medfreeman:
I completely understand, i also think that it is a bit heavy, as said before.

If doing this the simpler way, i think my [original solution](https://github.com/roots/trellis/issues/786#issuecomment-286078624) is eventually better.

Because not cleaning templates will fare better if the user has to override the no-default template ;
with your solution, you'd have to ssh onto the server to remove the no-default template while adding one with another name in certain cases, and i think that wouldn't be a good thing (in regards to automation).
And if reverting, ssh would have to be used again.

Typing this while thinking, the general one is also more useful..
I'm trying to find a good middle ground.
So supplying a new PR.

I hope it's simple yet flexible enough.

PR #791, opened 2017-03-16
Add additional nginx sites-enabled templates support

### LINKED ISSUE #786: Allow customizing the default site's Nginx configuration

medfreeman:
closes #786 

Add a `nginx_sites_enabled_templates` array
to allow users to specify custom files to be templated to the server
sites-enabled/extras folder, that are automatically included by nginx.
Default empty.

Add a `nginx_sites_enabled_cleanup` boolean variable that enforces
removing all files in sites-enabled/extras before copying new ones.
Default true.

Add a `nginx_default_site_enabled` boolean variable,
which defines whether to enable or disable trellis' default site
configuration.
Default true.

### DIFF
--- a/roles/nginx/defaults/main.yml
+++ b/roles/nginx/defaults/main.yml
@@ -5,6 +5,9 @@ nginx_logs_root: /var/log/nginx
 nginx_user: www-data
 nginx_fastcgi_buffers: 8 8k
 nginx_fastcgi_buffer_size: 8k
+nginx_default_site_enabled: true
+nginx_sites_enabled_cleanup: true
+nginx_sites_enabled_templates: []
 
 # Fastcgi cache params
 nginx_cache_path: /var/cache/nginx
--- a/roles/nginx/tasks/main.yml
+++ b/roles/nginx/tasks/main.yml
@@ -55,3 +55,36 @@
   args:
     creates: "{{ nginx_path }}/sites-enabled/no-default.conf"
   notify: reload nginx
+  tags: nginx-sites
+  when: nginx_default_site_enabled
+
+- name: Disable better default site
+  file:
+    path: "{{ nginx_path }}/sites-enabled/no-default.conf"
+    state: absent
+  notify: reload nginx
+  tags: nginx-sites
+  when: not nginx_default_site_enabled
+
+- name: Create sites-enabled/extras directory
+  file:
+    mode: 0755
+    path: "{{ nginx_path }}/sites-enabled/extras"
+    state: directory
+  tags: nginx-sites
+
+- name: Remove nginx extra enabled sites
+  file:
+    state: absent
+    path: "{{ nginx_path }}/sites-enabled/extras/"
+  notify: reload nginx
+  tags: nginx-sites
+  when: nginx_sites_enabled_cleanup
+
+- name: Create Nginx extra enabled sites
+  template:
+    src: "{{ item }}"
+    dest: "{{ nginx_path }}/sites-enabled/extras/{{ item | basename | regex_replace('.j2$', '') }}"
+  with_items: "{{ nginx_sites_enabled_templates }}"
+  notify: reload nginx
+  tags: nginx-sites
--- a/roles/nginx/tasks/main.yml
+++ b/roles/nginx/tasks/main.yml
@@ -74,9 +74,9 @@
   tags: nginx-sites
 
 - name: Remove Nginx extra enabled sites
-  file:
-    state: absent
-    path: "{{ nginx_path }}/sites-enabled/extras/"
+  command: rm -f {{ nginx_path }}/sites-enabled/extras/*
+  args:
+    warn: false
   notify: reload nginx
   tags: nginx-sites
   when: nginx_sites_enabled_cleanup
--- a/roles/nginx/tasks/main.yml
+++ b/roles/nginx/tasks/main.yml
@@ -73,7 +73,7 @@
     state: directory
   tags: nginx-sites
 
-- name: Remove nginx extra enabled sites
+- name: Remove Nginx extra enabled sites
   file:
     state: absent
     path: "{{ nginx_path }}/sites-enabled/extras/"
--- a/roles/nginx/templates/nginx.conf.j2
+++ b/roles/nginx/templates/nginx.conf.j2
@@ -169,6 +169,7 @@ http {
   {% endblock %}
 
   {% block sites_enabled -%}
-  include sites-enabled/*;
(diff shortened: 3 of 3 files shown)


medfreeman:
The overhead is very limited.
Added benefit: not changing where the default site has been since 2013.
In certain cases you'd want additional sites without disabling the default site also.
I hope that keepss things simple yet flexible.

fullyint:
@medfreeman You've done a ton of work and great thinking on this. Thank you! I know it can be really tedious to think through all the implementation and potentialities. You and @swalkinshaw have done the vast majority of the work here, and I'll just add a small percentage of brainstorming in the form of [another example branch](https://github.com/roots/trellis/compare/nginx-sites-confs).

Your response to @swalkinshaw's [example branch](https://github.com/roots/trellis/compare/nginx-sites-available-templates) was to seek an option that removed out-of-date confs from the server, to avoid the requirement for manual SSH connections to clean them up later. [My example](https://github.com/roots/trellis/compare/nginx-sites-confs) augments @swalkinshaw's example to pursue your goal.

- renamed `nginx_sites_available_templates` list to `nginx_sites_confs`, where each conf...
  - has a `src` attribute (template local path)
  - has an optional `enabled` attribute (default `true`)
- switched to templating the confs to `sites-available` instead of `sites-enabled`
- added task that links or unlinks confs based on `enabled` attribute
- added task that removes from `sites-available` the confs that are `enabled=false` (I don't think this task is necessary given that confs are unlinked, but maybe you desire to remove confs altogether?)

Consider this example a user could place in `group_vars/all/main.yml`:
```
nginx_sites_confs:
  - src: no-default.conf.j2
  - src: nginx-includes/some-outdated-site.conf.j2
    enabled: false
  - src: nginx-includes/some-site.conf.j2
```
Given that `enabled` defaults to `true`, `no-default.conf` and `some-site.conf` will be templated to `sites-available` and linked in `sites-enabled`. Given that `some-outdated-site.conf` is `enabled=false` (imagine a conf used previously, but now we want it gone), it will be unlinked and removed. This feature would probably need a paragraph in the [Nginx includes docs](https://roots.io/trellis/docs/nginx-includes/).

I think my example has a few advantages.
- The example avoids attempting the magic detection and cleanup of confs on the server that have no counterpart on the local control machine. This "cleanup" in the regular nginx-includes feature made sense to me given that users never have to manually list out the templates to be processed. But this proposed nginx-sites feature requires users to manually list out of templates. So, my example branch takes advantage of that manual listing, giving users the chance to specify the `enabled` status, and allowing us to avoid the magical find and remove.
- The example requires just one config in `defaults/main.yml` (the templates list), avoiding the need for extra configs like `nginx_default_site_enabled` and `nginx_sites_enabled_cleanup`. Of course, my proposed `nginx_sites_confs` has the downside of being more complex than just a regular list.
- The example gets around the scenario of having a task to "Enable better default site" followed by a task to "Disable better default site".
- The example avoids what looks like non-idempotence in having the "Remove Nginx extra enabled sites" task apparently remove all the `sites-enabled/extras/*` every time then having the next task add some back, every time.

**Your work is really good. What I've written here is just to share ideas and brainstorming, not negativity. My sentiment is 100% positive. However, I'm going to avoid the potential for burnout by not taking the extra time to finess the tone above to avoid what could come off as overbearing and critical.**

----

The four options, in my view.

**1. Minimal.** There's a big part of me that leans toward not adding anything other than the option to prevent the `no-default.conf` from being enabled. The rationale would be that all this extra nginx-sites stuff falls in the category of "extra server setup that a small minority of users might want," and thus could/should be handled by users' own custom playbooks.

**2. Lean and consistent.** However, if we _do_ add something, I'm very sympathetic with @swalkinshaw's [statement](https://github.com/roots/trellis/pull/790#issuecomment-286928219)
> Cleaning up the templates is a nice feature, but I don't know if we want to keep up that pattern in multiple places.

Aside from cleaning up nginx-includes, Trellis doesn't clean up site-specific confs or dirs after a site is removed from `wordpress_sites`, or remove old ssl certs, etc.

**3. Include cleanup.** If we do want to enable cleanup, it seems fine to me to leave confs in `sites-availble` after they are unlinked from `sites-enabled`. That would allow removing the "Remove disabled confs from sites-available" task from my example.  Part of my rationale is that servers built using Trellis are quite disposable because they're so easy to rebuild, so I doubt there would be a ton of old outdated confs that would build up in `sites-available` before the server is just destroyed/rebuilt.

**4. Extra-thorough cleanup.** But it's only one extra task to clean out `sites-available` and maybe there's an argument for it.

medfreeman:
Thanks for your work and comments ! I can assure you nothing comes up as negativity !

As you could perhaps foresee, i'm leaning for option 3.
But i'll share my opinions about each option anyway.

1) As @swalkinshaw [stated](https://github.com/roots/trellis/issues/786#issuecomment-285898921)

> Seems useful since we'd like to encourage people to have better setups with load balancers (even though in this case it seems ELB is kind of broken).

And given that:
  - this feature will save time, even for a small minority
  - the overhead is *really* small
  - there is already an nginx-includes feature that is complimentary to this one
  - in specific cases (like mine), i'd need to child template / disable / overwrite the default configuration, and that involves editing trellis files (not only adding a playbook because overwriting a file in a subsequent playbook doesn't seem indepotent), that then complicates trellis' updates on client repositories
  - we'll then can hopefully add simple [guides](https://github.com/roots/trellis/pull/650#issuecomment-287147470) for these kind of time consuming tasks (when you have to do the research and implementation the first time), also imo helping trellis' adoption by having clear ways of doing specific but possibly needed features

I think this feature would be a great thing to add.

2) I concur, but not cleaning *site specific confs or dirs* wouldn't create side effects, whereas this feature without cleaning could.
Also you could even template directly to sites-enabled the sites that are enabled, and remove the non-enabled sites just after, but the proposed approach feels better, having only one nginx reload statement.

3) Seems good to me. And that removes a step that is not frankly useful. As long as the sites are unlinked, this achieves the same result.

4) I share the opinion that this isn't needed, and even if it was, it would make more sense to template directly to sites-enabled in this case, because linking in the first place while removing the link and the linked file while enabled=false doesn't really make sense imo.

fullyint:
Thanks @medfreeman. Good arguments. This convinces me regarding adding cleanup:
>not cleaning _site specific confs or dirs_ wouldn't create side effects, whereas this feature without cleaning could

Option 3 looks good to me. :+1:

medfreeman:
Thanks !

PR #792, opened 2017-03-17
sshd: move AddressFamily above ListenAddress

kenbird:
Provisioning a server with Ansible 2.2.0.0 fails with:

failed to validate /root/.ansible/tmp/ansible-tmp-1489759318.18-198304737737752/source line 8: address family must be specified before ListenAddress.

This pull request simply moves the AddressFamily entry above the ListenAddress block

### DIFF
--- a/roles/sshd/templates/sshd_config.j2
+++ b/roles/sshd/templates/sshd_config.j2
@@ -5,6 +5,8 @@
 Port {{ port }}
 {% endfor -%}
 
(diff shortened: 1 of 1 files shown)


kenbird:
I should maybe mention the OS was Ubuntu 14.04.5 x64

swalkinshaw:
We'd have to verify it's the same on Ubuntu 16.04 as well.

swalkinshaw:
Thanks 👍

PR #793, opened 2017-03-21
Add additional nginx sites configurations support

### LINKED ISSUE #786: Allow customizing the default site's Nginx configuration

medfreeman:
closes #786

Add a nginx_sites_conf array
to allow users to specify custom nginx confs to be templated to the server's
nginx/sites-available folder, that are then symlinked to nginx/sites-enabled.
Array element properties:
  - src: jinja template path
  - enabled: whether to enable / disable the site conf 
by symlinking it to nginx/sites-enabled / removing it (default true)

Default:
```
nginx_sites_confs:
  - src: no-default.conf.j2
```

Add nginx-sites playbook tag.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Add additional Nginx sites configurations support ([#793](https://github.com/roots/trellis/pull/793))
 * Change `remote-user` role to `connection` role: tests host key, user ([#745](https://github.com/roots/trellis/pull/745))
 * Allow customization of PHP extensions ([#787](https://github.com/roots/trellis/pull/787))
 * Allow for per-project packagist.com authentication ([#762](https://github.com/roots/trellis/pull/762))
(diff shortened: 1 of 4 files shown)


swalkinshaw:
Looks great 👍 

Thanks for the work and patience on this

fullyint:
Works in my tests (and it unearthed #795 :+1:).
Thanks for this!

For the docs, yeah, I think `some-name.conf.site.j2` is fine to suggest as a convention. You [understood it all](https://github.com/roots/trellis/issues/786#issuecomment-288054646), that the child template convention of `*.child` is just to avoid the default templating to `includes.d`. People can use the convention or not. It is harmless to template extra files to `includes.d` that aren't part of the sitename subdirs.

medfreeman:
Thanks to you both ! @fullyint i finally commited your code, i hope you don't mind !
Taking my time writing the doc, hope to be the clearest possible.

Next proposition in a week or two, have something like an 'external' or 'proxy' ssl setting / proxy conf to handle ssl termination.
Rationale: it needs changes in a lot of places, is hard to maintain, but again not sure if it warrants a PR or something more like a recipe

fullyint:
Regarding SSL termination, the only prior discussion I recall was partial accommodation in #732. We ultimately figured child templates could handle the particular changes of #732, as implemented in "[Child template example – complex](https://roots.io/trellis/docs/nginx-includes/#child-template-example-complex)."

swalkinshaw:
Thanks 🚀

PR #796, opened 2017-03-22
Run WordPress installed check before finalizing deploy

fullyint:
The "WordPress installed?" deploy task occasionally detects
errors. It would be helpful to detect such errors and fail before
finalizing the deploy (before creating symlink to release dir).

This PR moves the two tasks related to "WordPress installed?"
from `finalize-after.yml` to a new `finalize-before.yml`, with one change:
```diff
- chdir: "{{ deploy_helper.current_path }}"
+ chdir: "{{ deploy_helper.new_release_path }}"
```

```
# Deploy steps

initialize.yml

update.yml

prepare.yml

build.yml

share.yml
                <-- tasks now here, in finalize-before.yml
finalize.yml
                <-- tasks formerly here, in finalize-after.yml
```

**Edit.** Moved the two new tasks from `build-after.yml` to a new `finalize-before.yml`.

### DIFF
--- a/deploy.yml
+++ b/deploy.yml
@@ -18,6 +18,7 @@
   vars:
     deploy_build_before: "{{ playbook_dir }}/deploy-hooks/build-before.yml"
     deploy_build_after: "{{ playbook_dir }}/roles/deploy/hooks/build-after.yml"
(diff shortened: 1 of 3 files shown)


swalkinshaw:
🚀 good improvement

PR #797, opened 2017-03-23
Check Ansible version before Ansible validates task attributes

fullyint:
Avoids `ERROR! 'check_mode' is not a valid attribute for a Task` for Ansible < 2.2 ([example](https://discourse.roots.io/t/error-check-mode-is-not-a-valid-attribute-for-a-task/8984)).

The challenge is that Ansible validates task attributes before running a playbook, before the [current version check](https://github.com/roots/trellis/blob/acfb5fd78df73977b735f0021736f0f8fd8f5401/roles/common/tasks/main.yml#L2-L11) can run, and even before the [callback plugin](https://github.com/roots/trellis/blob/acfb5fd78df73977b735f0021736f0f8fd8f5401/lib/trellis/plugins/callback/output.py#L11-L12) is processed.

A vars plugin will load before Ansible validates task attributes, so it appeared to be the only option for performing a version check and helpful "please upgrade" message.

This PR adds back a vars plugin as a bit of a hack for the sake of the Ansible version check. Trellis formerly had a vars plugin (removed in #684 and #707).
**Edit:** Modeled after Ansible's [`version_compare`](https://github.com/ansible/ansible/blob/0c449598009fb440e150e9578ab3a18958cfca41/lib/ansible/plugins/test/core.py#L112-L113) test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Check Ansible version before Ansible validates task attributes ([#797](https://github.com/roots/trellis/pull/797))
 * Add additional Nginx sites configurations support ([#793](https://github.com/roots/trellis/pull/793))
 * Change `remote-user` role to `connection` role: tests host key, user ([#745](https://github.com/roots/trellis/pull/745))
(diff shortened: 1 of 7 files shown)


swalkinshaw:
👍

PR #799, opened 2017-03-24
Added home_root to group_vars/main with /home/ as default

slackday:
Fixes problem when deploying to servers not provisioned with Trellis. I run a mix of sites on the same Trellis box. Some servers I deploy to use different users than Trellis default. 

When a server has a different directory for /home/ this causes problems when Ansible wants to modify `known_hosts` as in latest version of Trellis. 

For example one server uses `/mnt/persist/home/` and that gave me the problem discussed here https://discourse.roots.io/t/failed-to-connect-to-the-host-via-ssh-openssh-6-2p2-osslshim-0-9-8r-8-dec-2011-command-line-line-0-bad-protocol-2-host-key-algorithms-ssh-ed25519-cert-v01-openssh-com-ssh-rsa-cert-v01-openssh-com-ssh-ed25519-ssh-rsa/9142/5?u=stefanlindberg

I suggest adding a variable to `group_vars` called home_root (similar to web_root) that can be overwritted on a per site basis with `host_vars`. 

The default is `/home/` which so out of the box it works as before.

### DIFF
--- a/roles/deploy/tasks/update.yml
+++ b/roles/deploy/tasks/update.yml
@@ -6,7 +6,7 @@
   known_hosts:
     name: "{{ item.name }}"
     key: "{{ item.key | default(omit) }}"
-    path: "{{ item.path | default(home_root + ansible_user + '/.ssh/known_hosts') }}"
+    path: "{{ item.path | default(omit) }}"
     state: "{{ item.state | default('present') }}"
   with_items: "{{ known_hosts | default([]) }}"
 
--- a/roles/deploy/tasks/update.yml
+++ b/roles/deploy/tasks/update.yml
@@ -6,7 +6,7 @@
   known_hosts:
     name: "{{ item.name }}"
     key: "{{ item.key | default(omit) }}"
-    path: "{{ item.path | default('/home/' + ansible_user + '/.ssh/known_hosts') }}"
+    path: "{{ item.path | default(home_root + ansible_user + '/.ssh/known_hosts') }}"
     state: "{{ item.state | default('present') }}"
   with_items: "{{ known_hosts | default([]) }}"
 
--- a/roles/wordpress-install/tasks/main.yml
+++ b/roles/wordpress-install/tasks/main.yml
@@ -20,7 +20,7 @@
   known_hosts:
     name: "{{ item.name }}"
     key: "{{ item.key | default(omit) }}"
-    path: "{{ item.path | default('/home/' + ansible_user + '/.ssh/known_hosts') }}"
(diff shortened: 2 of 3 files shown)


roles/deploy/tasks/update.yml
```diff
@@ -6,7 +6,7 @@
   known_hosts:
     name: "{{ item.name }}"
     key: "{{ item.key | default(omit) }}"
-    path: "{{ item.path | default('/home/' + ansible_user + '/.ssh/known_hosts') }}"
+    path: "{{ item.path | default(home_root + ansible_user + '/.ssh/known_hosts') }}"
```
fullyint: How about using [`default(omit)`](http://docs.ansible.com/ansible/playbooks_filters.html#omitting-parameters) instead?
`path: "{{ item.path | default(omit) }}"`

It should default to `~/.ssh/known_hosts` ([docs](http://docs.ansible.com/ansible/known_hosts_module.html) and [src](https://github.com/ansible/ansible-modules-extras/blob/stable-2.2/system/known_hosts.py#L274)). This works in my tests (but I didn't test custom home dir). I probably should have used `omit` in #751.

If this works for your user's custom home directory, we could use `omit` for both tasks' `path` and no longer need a new `home_root` variable.

roles/wordpress-install/tasks/main.yml
```diff
@@ -20,7 +20,7 @@
   known_hosts:
     name: "{{ item.name }}"
     key: "{{ item.key | default(omit) }}"
-    path: "{{ item.path | default('/home/' + ansible_user + '/.ssh/known_hosts') }}"
+    path: "{{ item.path | default(home_root + ansible_user + '/.ssh/known_hosts') }}"
     state: "{{ item.state | default('present') }}"
   with_items: "{{ known_hosts | default([]) }}"
```
fullyint: If you go with `default(omit)` for `path`, let's add `become: no` right above `with_items` (this task only). This will make the task run as `vagrant` (vs. `root`), adding to `known_hosts` in $HOME for `vagrant`. This is good because the subsequent composer install task that needs the `known_hosts` will run as `vagrant` (also uses [`become: no`](https://github.com/roots/trellis/blob/a7506ed3fcf4152da0a01d0e34efa100aedd6033/roles/wordpress-install/tasks/main.yml#L41)).

slackday:
I just tried `default(omit)` with custom home dir and it works the same as my `home_root` group_var.

I removed it and update the PR to use `default(omit)` instead.

Thanks!

fullyint:
This is a good improvement. Thank you @slackday! :+1:

slackday:
Thank you :) I just got my deploys working again! Happy weekend 🎉

PR #802, opened 2017-03-26
Add cron task for multisite

QWp6t:


swalkinshaw:
One concern with this was installs with a large number of sites

QWp6t:
For those sites, they can modify it accordingly.

This would probably be good for _most_ multisite installations.

We could put it behind a flag, something like `enable_multisite_cron` and have it true by default. Then when their site grows, they can switch it to false ezpz.

fullyint:
>This would probably be good for _most_ multisite installations.
>
>We could put it behind a flag, something like `enable_multisite_cron` and have it true by default. Then when their site grows, they can switch it to false ezpz.

I don't have experience on the topic, but the default of `true` sounds fine to me. If `true`, then the `cron: false` that this PR adds to the `group_vars` files should be `true`.

But actually, my inclination would be to not add this new `cron` setting to the `group_vars` files at all, in keeping with the trend toward streamlining the configs in these files. Instead, the existing `item.value.multisite.cron | default(true)` in the new `cron_enabled` helper would handle the default and the [multisite docs](https://roots.io/trellis/docs/multisite/) could mention the `cron` option.

PR #807, opened 2017-03-28
Fix nginx server name wildcard entry in multisite subdomain installs

### LINKED ISSUE #806: Wrong root domain used as base for nginx multisite config server_name wildcard entry when the main site domain starts by www.

- [X] This is a bug report

---

Given this wordpress_sites.yml

```
wordpress_sites:
  example.com:
    site_hosts:
      - canonical: www.example.com
        redirects:
          - example.com
    local_path: ../site
    repo: git@github.com:nickname/example_com.git
    repo_subtree_path: site
    branch: production
    multisite:
      enabled: true
      subdomains: true
    env:
      domain_current_site: example.com
      wp_home: http://www.example.com
      wp_siteurl: http://www.example.com/wp
    ssl:
      enabled: false
    cache:
      enabled: false
```

The nginx configuration of the corresponding 'example.com' site contains : 
```
server_name    www.example.com *.www.example.com ;
```

As seen in [roles/wordpress-setup/templates/wordpress-site.conf.j2 l.8](https://github.com/roots/trellis/blob/32a078ab09ea1fbe25f4daf1319b2a234e233242/roles/wordpress-setup/templates/wordpress-site.conf.j2#L8), the wildcard entry is generated from the 'canonical' host entry.

The nginx configuration of the corresponding 'example.com' should contain : 
```
server_name    www.example.com *.example.com ;
```

The multisite root domain is example.com, and it subsites are available on subdomains of example.com, not www.example.com.
www.example.com is only the canonical url of the main site.

The wildcard entry should be generated from the 'domain_current_site' host entry.

medfreeman:
fixes #806 

Use `domain_current_site` host variable instead of `canonical` variable

### DIFF
--- a/roles/wordpress-setup/templates/wordpress-site.conf.j2
+++ b/roles/wordpress-setup/templates/wordpress-site.conf.j2
@@ -5,7 +5,7 @@
 server {
   {% block server_id -%}
   listen {{ ssl_enabled | ternary('443 ssl http2', '80') }};
-  server_name  {% for host in site_hosts_canonical %}{{ host }} {% if item.value.multisite.subdomains | default(false) %}*.{{ host }} {% endif %}{% endfor %};
+  server_name  {% for host in site_hosts_canonical %}{{ host }}{% endfor %} {% if item.value.multisite.subdomains | default(false) %}*.{{ item.value.env.domain_current_site }} {% endif %};
   {% endblock %}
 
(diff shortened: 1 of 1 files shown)


medfreeman:
This solution is incorrect, due to a misunderstanding on wordpress multisite subdomains inner workings.

PR #810, opened 2017-03-30
Ansible 2.2.2 compatibility

fullyint:
Fixes https://discourse.roots.io/t/deploy-fail-if-repo-subtree-path-is-set-incorrectly/9235

Prior to Ansible 2.2.2, a registered var from a non-looping skipped
task was undefined. As of Ansible 2.2.2, such a var is defined.
This commit changes a subsequent task's conditional to no longer rely
on the registered var's inconsistent status as defined/undefined.

I found no other instances of this type of problem, nor of any other Ansible 2.2.2 incompatibility.


swalkinshaw:
🚀

PR #811, opened 2017-03-30
Remove unused registered var generate_account_key

fullyint:
The registered var `generate_account_key ` is [not used](https://github.com/roots/trellis/search?utf8=%E2%9C%93&q=generate_account_key&type=). It appears to be an unused leftover from the [original role](https://github.com/andreaswolf/ansible-role-letsencrypt/blob/0e292997640c8bd002e6abed3b727a6b473c2b79/tasks/main.yml#L91-L106).


swalkinshaw:
🚀

PR #812, opened 2017-03-30
Add self signed certificates subdomains

### LINKED ISSUE #809: Add subdomains and all canonical names to self-signed certificates

---

Self-signed certificates are only generated for the main domain canonical host name in a multisite subdomain installation.

Self-signed certificates are generated for the main domain canonical host name, all other canonical host names, and wildcard entries.

---

## Feature Request

A few reasons:
  - consistency with letsencrypt
  - simplifying redirection testing in complex scenarios (by avoiding adding exceptions for each subdomain)
  - simplifying development for teams with a lot of subsites (in a client's doc i have a few steps guide, ~ 4
terminal commands, to add the certs to the trust store in ubuntu and macOS, and avoid further exceptions, sites show as valid)
  - using browsersync in theme development (it needs to have the self-signed cert for the refresh to happen, i've added a dev.yml task that changes the permission of certs, allowing browsersync to read them)

Previous positions:
[Comment on Allow auto-generation of self signed SSL certificate. PR](https://github.com/roots/trellis/pull/279#issuecomment-122674423)

**Implementation**

Move [Get list of hosts in current Nginx conf](https://github.com/roots/trellis/blob/32a078ab09ea1fbe25f4daf1319b2a234e233242/roles/letsencrypt/tasks/nginx.yml#L7-L14) letsencrypt task to the `common` role, in a specific task.

Include it in letsencrypt and self-signed certs tasks.

Include the main domain canonical host name, the other canonical host names, and the wildcard entry as DNS subjectAltNames entries, as explained [here](http://blog.endpoint.com/2014/10/openssl-csr-with-alternative-names-one.html) like @swalkinshaw stated previously.
Down in this article's comments, there's a [link to an online certificate request generator](https://certificatetools.com/) that provides a clean one-liner (in `OpenSSL command (Linux)` after submitting the form).
Though it misses the point of also having ([as stated here](http://blog.endpoint.com/2014/10/openssl-csr-with-alternative-names-one.html?showComment=1414727230063#c5347857091415121519)) the main domain name as first DNS subjectAltName.

i.e. resulting command
```
openssl req -new -newkey rsa:2048 -nodes -sha256 -subj "/CN=www.test.com" -config <(
cat <<-EOF
[req]
req_extensions = req_ext
[ req_ext ]
subjectAltName = @alt_names
[alt_names]
DNS.1 = www.test.com
DNS.2 = test.com
DNS.3 = *.test.com
EOF
)
```

Side note: i understand that some issues can be viewed as (and are sometimes) corner cases, but i've stumbled on an important number of them on a single project. Since i know i'll encounter part of the same problems in subsequent projects (and obviously different ones), imho the most *sensible* approach for me is to try including them in trellis (especially with this one that is transparent to most users, and matches what i thought trellis would do by itself). Or facing maintenance hell.

medfreeman:
closes #809 

Add support for all alternate hosts and subdomains in self-signed certificates
The `subjectAltName` certificate field now contains the DNS entries of all canonical and redirect hosts of a site, along with all wildcard domain names (for a site with multisite and subdomains enabled).

Add `multisite_subdomains_wildcards` helper variable
Replace server_name expression in wordpress sites template by `multisite_subdomains_wildcards` helper variable for clearer code

### DIFF
--- a/roles/wordpress-setup/tasks/self-signed-certificate.yml
+++ b/roles/wordpress-setup/tasks/self-signed-certificate.yml
@@ -14,7 +14,7 @@ EOF\n
     -keyout {{ item.key | quote }}.key -out {{ item.key | quote }}.cert"
   args:
     executable: "/bin/bash"
-    chdir: "{{ nginx_path }}/ssl"
+    chdir: "{{ nginx_ssl_path }}"
     creates: "{{ item.key }}.*"
   with_dict: "{{ wordpress_sites }}"
   when: ssl_enabled and item.value.ssl.provider | default('manual') == 'self-signed'
--- a/roles/wordpress-setup/tasks/self-signed-certificate.yml
+++ b/roles/wordpress-setup/tasks/self-signed-certificate.yml
@@ -1,10 +1,19 @@
 ---
 - name: Generate self-signed certificates
-  shell: >
-    openssl req -subj "/CN={{ item.value.site_hosts[0].canonical }}" -new
-    -newkey rsa:2048 -days 3650 -nodes -x509 -sha256
-    -keyout {{ item.key }}.key -out {{ item.key }}.cert
+  shell: "openssl req -subj \"/CN={{ item.value.site_hosts[0].canonical }}\" -new \
+    -newkey rsa:2048 -days 3650 -nodes -x509 -sha256 \
+    -extensions req_ext -config <( \
+cat <<\" EOF\"\n
+[req]\n
+distinguished_name = dn\n
+[ dn ]\n
+[ req_ext ]\n
+subjectAltName = {{ site_hosts | union(multisite_subdomains_wildcards) | map('regex_replace', '(.*)', 'DNS:\\1') | join(',') }}\n
+EOF\n
+    ) \
+    -keyout {{ item.key | quote }}.key -out {{ item.key | quote }}.cert"
   args:
+    executable: "/bin/bash"
     chdir: "{{ nginx_path }}/ssl"
     creates: "{{ item.key }}.*"
   with_dict: "{{ wordpress_sites }}"
--- a/roles/wordpress-setup/tasks/self-signed-certificate.yml
+++ b/roles/wordpress-setup/tasks/self-signed-certificate.yml
@@ -1,13 +1,15 @@
 ---
 - name: Generate self-signed certificates
-  shell: "openssl req -subj \"/CN={{ item.value.site_hosts[0].canonical }}\" -new \
-    -newkey rsa:2048 -days 3650 -nodes -x509 -sha256 \
+  shell: "openssl req -new -newkey rsa:2048 \
+    -days 3650 -nodes -x509 -sha256 \
     -extensions req_ext -config <( \
-cat <<\" EOF\"\n
+cat <<' EOF'\n
 [req]\n
-distinguished_name = dn\n
-[ dn ]\n
-[ req_ext ]\n
+prompt = no\n
+distinguished_name = req_dn\n
+[req_dn]\n
+commonName = {{ item.value.site_hosts[0].canonical }}\n
+[req_ext]\n
 subjectAltName = {{ site_hosts | union(multisite_subdomains_wildcards) | map('regex_replace', '(.*)', 'DNS:\\1') | join(',') }}\n
 EOF\n
     ) \
--- a/group_vars/all/helpers.yml
+++ b/group_vars/all/helpers.yml
@@ -13,4 +13,5 @@ site_env: "{{ wordpress_env_defaults | combine(item.value.env | default({}), vau
 site_hosts_canonical: "{{ item.value.site_hosts | map(attribute='canonical') | list }}"
 site_hosts_redirects: "{{ item.value.site_hosts | selectattr('redirects', 'defined') | sum(attribute='redirects', start=[]) | list }}"
 site_hosts: "{{ site_hosts_canonical | union(site_hosts_redirects) }}"
+multisite_subdomains_wildcards: "{{ item.value.multisite.subdomains | default(false) | ternary( site_hosts_canonical | map('regex_replace', '^(www\\.)?(.*)$', '*.\\2') | list, [] ) }}"
 ssl_enabled: "{{ item.value.ssl is defined and item.value.ssl.enabled | default(false) }}"
--- a/roles/wordpress-setup/templates/wordpress-site.conf.j2
+++ b/roles/wordpress-setup/templates/wordpress-site.conf.j2
@@ -5,7 +5,7 @@
 server {
   {% block server_id -%}
   listen {{ ssl_enabled | ternary('443 ssl http2', '80') }};
-  server_name {{ site_hosts_canonical | join(' ') }}{% if item.value.multisite.subdomains | default(false) %} *.{{ site_hosts_canonical | map('regex_replace', '^www\.', '') | join(' *.') }}{% endif %};
+  server_name {{ site_hosts_canonical | union(multisite_subdomains_wildcards) | join(' ') }};
   {% endblock %}
 
   {% block logs -%}
@@ -45,7 +45,7 @@ server {
 
   {% endif -%}
   {% endblock -%}
-  
+
   {% block multisite_rewrites -%}
   {% if item.value.multisite.enabled | default(false) -%}
   # Multisite rewrites
@@ -157,7 +157,7 @@ server {
 # Redirect to https
 server {
   listen 80;
-  server_name {{ site_hosts | join(' ') }}{% if item.value.multisite.subdomains | default(false) %} *.{{ site_hosts_canonical | map('regex_replace', '^www\.', '') | join(' *.') }}{% endif %};
+  server_name {{ site_hosts | union(multisite_subdomains_wildcards) | join(' ') }};
 
   {{ self.acme_challenge() -}}
 


roles/wordpress-setup/tasks/self-signed-certificate.yml
```diff
@@ -1,10 +1,19 @@
 ---
 - name: Generate self-signed certificates
-  shell: >
-    openssl req -subj "/CN={{ item.value.site_hosts[0].canonical }}" -new
-    -newkey rsa:2048 -days 3650 -nodes -x509 -sha256
-    -keyout {{ item.key }}.key -out {{ item.key }}.cert
+  shell: "openssl req -subj \"/CN={{ item.value.site_hosts[0].canonical }}\" -new \
```
medfreeman: The change of command style was mandatory for a few reasons, first one is that the brackets characters (l.7-10) have to be escaped or ansible throws a syntax error.

The original `Block` scalar style using a `>` removes the newlines within the string, thus not allowing the `config` openssl section to work properly.

The choice of the `Flow` scalar style using a `"` (double-quote) is the only style in yaml multi-line string allowing the use of `inline newlines` (`\n`) and `spaceless newlines` (trailing `\`) (c.f. http://stackoverflow.com/questions/3790454/in-yaml-how-do-i-break-a-string-over-multiple-lines#answer-21699210).

I could have also used the `Flow` literal style using a `|`, but that would have forced all the dashed parameters of the 
 openssl command to be written on the same line since a spaceless newline (trailing `\`) can't be added in this mode.

In the chosen style, the double quotes `"` and backslash `\` have to be escaped though.

roles/wordpress-setup/tasks/self-signed-certificate.yml
```diff
@@ -1,10 +1,19 @@
...
+[req]\n
+distinguished_name = dn\n
+[ dn ]\n
+[ req_ext ]\n
+subjectAltName = {{ site_hosts | union(multisite_subdomains_wildcards) | map('regex_replace', '(.*)', 'DNS:\\1') | join(',') }}\n
+EOF\n
+    ) \
+    -keyout {{ item.key | quote }}.key -out {{ item.key | quote }}.cert"
   args:
+    executable: "/bin/bash"
```
medfreeman: The bash shell has to be used, since virtual files use the subshell operator `()` (l.5), which produces a syntax error using the default `sh` shell.
Perhaps there is a way to make this work with sh, but i didn't find it.

roles/wordpress-setup/tasks/self-signed-certificate.yml
```diff
@@ -1,10 +1,19 @@
...
-    openssl req -subj "/CN={{ item.value.site_hosts[0].canonical }}" -new
-    -newkey rsa:2048 -days 3650 -nodes -x509 -sha256
-    -keyout {{ item.key }}.key -out {{ item.key }}.cert
+  shell: "openssl req -subj \"/CN={{ item.value.site_hosts[0].canonical }}\" -new \
+    -newkey rsa:2048 -days 3650 -nodes -x509 -sha256 \
+    -extensions req_ext -config <( \
+cat <<\" EOF\"\n
+[req]\n
+distinguished_name = dn\n
+[ dn ]\n
```
medfreeman: Providing an empty `dn` section is mandatory, even with the subject provided in command-line (l.3), while using a config file, or openssl throws an error.

roles/wordpress-setup/tasks/self-signed-certificate.yml
```diff
@@ -1,10 +1,19 @@
 ---
 - name: Generate self-signed certificates
-  shell: >
-    openssl req -subj "/CN={{ item.value.site_hosts[0].canonical }}" -new
-    -newkey rsa:2048 -days 3650 -nodes -x509 -sha256
-    -keyout {{ item.key }}.key -out {{ item.key }}.cert
+  shell: "openssl req -subj \"/CN={{ item.value.site_hosts[0].canonical }}\" -new \
+    -newkey rsa:2048 -days 3650 -nodes -x509 -sha256 \
+    -extensions req_ext -config <( \
+cat <<\" EOF\"\n
```
medfreeman: Ansible issues ansible/ansible#12856 and ansible/ansible#8512 prevent the use of `EOF` or `-EOF` as file delimiter, since a space character is inserted by ansible at the start of each line while parsing the shell string.
If found the solution [here in the second issue](https://github.com/ansible/ansible/issues/8512#issuecomment-102382070).

I don't know if you have a better understanding of the problem than me, but i couldn't find a better solution here or elsewhere.

medfreeman: Using the `cmd` shell parameter instead of providing the command directly to the `shell` module yields the same error while using `EOF` instead of `\" EOF\"`.

roles/wordpress-setup/tasks/self-signed-certificate.yml
```diff
@@ -1,10 +1,19 @@
 ---
 - name: Generate self-signed certificates
-  shell: >
-    openssl req -subj "/CN={{ item.value.site_hosts[0].canonical }}" -new
-    -newkey rsa:2048 -days 3650 -nodes -x509 -sha256
-    -keyout {{ item.key }}.key -out {{ item.key }}.cert
+  shell: "openssl req -subj \"/CN={{ item.value.site_hosts[0].canonical }}\" -new \
+    -newkey rsa:2048 -days 3650 -nodes -x509 -sha256 \
+    -extensions req_ext -config <( \
```
medfreeman: The `extensions` parameter is mandatory to have openssl x509 key generation take into account the subjectAltName parameter (it references the `req_ext` section l.10).

roles/wordpress-setup/tasks/self-signed-certificate.yml
medfreeman: Perhaps that can work with backticks, i didn't realise it until now. I'll try this during the afternoon.

swalkinshaw:
@medfreeman can you take a look at https://github.com/roots/trellis/pull/818 as well? Solves some (all?) of the same problem.

The method of using an Openssl config might be a much nicer option than the shell command.

roles/wordpress-setup/tasks/self-signed-certificate.yml
medfreeman: https://github.com/ansible/ansible/issues/12034

medfreeman:
@swalkinshaw sure.

While #818 can solve the *syntax* problem, it only adds a second task for each site to achieve the same result (config file vs command-line virtual config via here-now document, since subjectAltName cannot directly passed as a command-line argument).

As is this pr won't also work with multiple site keys, because the last site will have overwritten previous openssl configurations.

To follow this path, two options are possible:
  - Generate a specific openssl config file for each site
    -> As stated that only duplicates the cert task and will produce additional files, is this really warranted ? Moreover the original config works already, and i don't like duplicating a common config with args that can be passed in the command-line

  - Generate a single openssl config for all sites in one go
    -> I personally prefer having them separate for consistency, and keeps working as-is with one per-site (and subdomains) certificate pair

I personaly am in favor of writing a simpler syntax of my PR, but perhaps it is because i have written it.
Please tell me what you think.

medfreeman:
Ok, i made the syntax cleaner by avoiding unecessary quoting, including commonName in the configuration, and using single quotes for ` EOF`.
Imo the only thing that goes in favor of #818 is that this PR will break when ansible/ansible/issues/12034 is resolved.

heyfletch:
I got this to work. Didn't try #818.  This has become critical now that [Chrome 58 will not load the local https site without Subject Alternative Name](https://medium.com/carwow-product-engineering/chrome-58-and-self-signed-ssl-certificate-c28a874d80fa). 

I had a rough go of this, so reiterating some points here for others:

1. Start with @medfreeman's testing procedure and be sure the certs and keys are gone (those commands didn't clear them for me, so the reprovisioning was not working)
2. The ansible-playbook provision line didn't work either, so did: `vagrant reload --provision` instead
3. The `sudo security add-trusted-cert` on macOS didn't work for me, so I just did it manually by opening dev site, viewing certificate, dragging to desktop, dragging into keychain. 

Not sure why some of that didn't work, but in either case, it will be helpful to update the self-signed section of the SSL docs on how to trust ssl certificates with Keychain, etc. to avoid browser errors.

heyfletch:
My cheatsheet for new site with https/SSL with Subject Alternative Name on a local Mac:

1. Incorporate [this pull request's changes](https://github.com/roots/trellis/pull/812/files): 
2. Create and trust the local certificate:

```bash
openssl s_client -connect example.localdev:443 < /dev/null > example.raw
openssl x509 -inform PEM -in example.raw -text -out example.localdev.crt
sudo security add-trusted-cert -d -k /Library/Keychains/System.keychain example.localdev.crt
rm -f example.raw example.localdev.crt
```

Haven't tried this on multisite.

Curious, anything else we need to do to merge this?  Any way I can help?

fullyint:
I'm only a novice on the topic but tried to thoroughly review. @medfreeman I really appreciate your exceptionally thorough research and explanation. This works in my tests and I don't have any requested changes.

**heredoc vs. config file.** I slightly prefer the _here document_ approach here in #812 over the templated openssl config file approach in #818. I find @medfreeman's concern about #818's approach to be convincing (especially for projects with 10–20+ sites): 
>[by creating openssl config files, #818] duplicates the cert task and will produce additional files... and i don't like duplicating a common config with args that can be passed in the command-line  _[--ref](https://github.com/roots/trellis/pull/812#issuecomment-292708297)_

**extensions.** The other main difference I noticed was that #818 has additional config file sections:
- `req_extensions = v3_req`
- `x509_extensions = v3_ca`

As best I could tell these extensions are considered optional for webservers and probably of no additional benefit to non-CA self-signed certs aimed at local development. @scherii perhaps you can correct me if I've misunderstood.

**#812 appears preferable and ready.** I'm not opposed to #818 but I lean toward #812 for the reasons above and because I think #818 needs more discussion:
- I think #818 should include multisite subdomains (wildcard) in SANs like #812 does (and probably also include `redirects` hosts, as discussed in #818 comments).
- I'm uncertain about the value of some of the extensions in #818's config file, as discussed above.
- #818's `commonName` should perhaps be the first `canonical` host instead of `{{ item.key }}`, just because site keys in `wordpress_sites` aren't required to be host names. But I'm not sure how much `commonName` matters anymore or if I'm understanding it.

Huge thanks to @scherii and @medfreeman for such awesome work. Having to discuss which of two excellent PRs to pick is a great problem to have.

fullyint:
Thank you again @medfreeman @scherii @heyfletch 
This is a great improvement to Trellis!

PR #813, opened 2017-03-30
Ansible 2.3 compatibility

fullyint:
Broken out into separate commits, each with a commit message explaining. One commit prevents Ansible's new [`dense.py`](https://github.com/ansible/ansible/blob/v2.3.0.0-0.2.rc2/lib/ansible/plugins/callback/dense.py#L89) callback from causing Trellis ssh connection tests to always show `root` as failing to connect. The other commits just prevent Ansible's [new warning message](https://github.com/ansible/ansible/commit/ff20ab7d4439ad6e5ca099c428ac6d3f25a154e7) about jinja delimiters.

```
[WARNING]: when statements should not include jinja2 templating delimiters such as {{ }} or {% %}.
```

----

The commit `Avoid single var containing jinja delimiters in when parameter` deserves some explanation but don't expect this explanation to stand on its own without examining the commit's diff.

Consider a `when` parameter with  a single var, e.g., [`when: site_uses_letsencrypt`](https://github.com/roots/trellis/blob/41b7f641d650af26d51800943045b5a412763168/roles/letsencrypt/tasks/nginx.yml#L13). The var [will be expanded](https://github.com/ansible/ansible/blob/v2.3.0.0-0.2.rc2/lib/ansible/playbook/conditional.py#L138-L139) and the var's jinja delimiters [will trigger](https://github.com/ansible/ansible/blob/v2.3.0.0-0.2.rc2/lib/ansible/playbook/conditional.py#L141) related warnings. This PR's commit avoids the warnings by removing jinja delimiters from the var value/definition:

```diff
# roles/letsencrypt/defaults.main.yml
- site_uses_letsencrypt: "{{ ssl_enabled and item.value.ssl.provider | default('manual') == 'letsencrypt' }}"
+ site_uses_letsencrypt: ssl_enabled and item.value.ssl.provider | default('manual') == 'letsencrypt'
```

That change prevents the warnings, but makes the `site_uses_letsencrypt` var not function in contexts that aren't part of the `when` task parameter. This accounts for why the commit in question removes `site_uses_letsencrypt` from the definition of `missing_hosts`
```diff
# roles/letsencrypt/defaults.main.yml
- missing_hosts: "{{ site_uses_letsencrypt | ternary(site_hosts, []) | difference(... bunch of stuff
+ missing_hosts: "{{ site_hosts | difference(... bunch of stuff
```

This change to `missing_hosts` means its two appearances in the form
`when: missing_hosts | count` must be adjusted. However, it turns out that the lack of jinja delimiters in `site_uses_letsencrypt` causes this adjustment to be ineffective:

```diff
- when: missing_hosts | count
+ when: missing_hosts | count and site_uses_letsencrypt
```

This accounts for why the commit in question uses the strategy below, leaving the `site_uses_letsencrypt` alone in a single item (vs. the compound item in the diff above).

```diff
- when: missing_hosts | count
+ when:
+   - site_uses_letsencrypt
+   - missing_hosts | count
```

----

Tested on Ansible [branch 2.3-stable](https://github.com/ansible/ansible/commits/stable-2.3) as of `07ea6a6` and on Ansible 2.2.0.0. Tests included default configs and alternative configs such as Let's Encrypt enabled, nginx-includes, and `root` login disabled, on DigitalOcean and AWS EC2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Ansible 2.3 compatibility ([#813](https://github.com/roots/trellis/pull/813))
 * Remove potentially dangerous `db_import` option ([#825](https://github.com/roots/trellis/pull/825))
 
 ### 1.0.0-rc.1: April 7th, 2017
--- a/roles/connection/tasks/main.yml
+++ b/roles/connection/tasks/main.yml
@@ -52,7 +52,7 @@
 - block:
   - name: Set remote user for each host
     set_fact:
-      ansible_user: "{{ ansible_user | default(('root' in connection_status.stdout_lines) | ternary('root', admin_user)) }}"
+      ansible_user: "{{ ansible_user | default((connection_status.stdout_lines | intersect(['root', '\e[0;32mroot']) | count) | ternary('root', admin_user)) }}"
     check_mode: no
 
   - name: Announce which user was selected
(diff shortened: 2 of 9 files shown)


fullyint:
This PR works with today's official release of Ansible 2.3.

PR #817, opened 2017-04-05
Accommodate Ansible 2.1's new loop_var in output formatting

fullyint:
Usage of Ansible's new [`loop_var`](http://docs.ansible.com/ansible/playbooks_loops.html#loop-control) causes the following warning about failure in custom output plugin:
```
[WARNING]: Failure using method (v2_runner_item_on_skipped) in callback plugin (<ansible.plugins.callback.output.CallbackModule object at 0x108c3b8d0>): 'item'
```

This PR prevents the error by no longer hard-coding the 'item' `loop_var`.

A playbook to demonstrate the error on current master, vs. no error on PR branch:
```
- gather_facts: false
  hosts: localhost
  vars:
    some_dict:
      example.com:
        foo: bar
      site.com:
        foo: bar
  tasks:
    - command: echo {{ some_loop_var }}
      register: some
      with_dict: "{{ some_dict }}"
      loop_control:
        loop_var: some_loop_var
    - command: echo {{ item }}
      with_items: "{{ some.results }}"
```

The purpose of the `replace_item_with_key()` feature that this PR adjusts is to truncate output. Ansible 2.2's new new loop_control `label` parameter seems targeted at the same purpose, so this PR makes Trellis only use its feature if the `label` parameter is absent (conditional on `'label' not in result._task._ds.get('loop_control', {})`).

### DIFF
--- a/lib/trellis/utils/output.py
+++ b/lib/trellis/utils/output.py
@@ -50,11 +50,16 @@ def reset_task_info(obj, task=None):
 
 # Display dict key only, instead of full json dump
 def replace_item_with_key(obj, result):
-    if not obj._display.verbosity:
-        if 'key' in result._result['item']:
-            result._result['item'] = result._result['item']['key']
-        elif 'item' in result._result['item'] and 'key' in result._result['item']['item']:
-            result._result['item'] = result._result['item']['item']['key']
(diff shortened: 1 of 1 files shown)


swalkinshaw:
🚀

PR #818, opened 2017-04-05
Generate multi-domain self-signed certificate with Subject Alternative Names (SAN)

scherii:
Generate a self-signed certificate with SAN (Subject Alternative Name) based on wordpress_sites dict.
All canonical domains will be included in the snakeoil certificate.

`group_vars/development/wordpress_sites.yml`
```
wordpress_sites:
  example.com:
    site_hosts:
      - canonical: example-a.test
      - canonical: example-b.test
      - canonical: example-c.test

…
```

### DIFF
--- a/roles/wordpress-setup/tasks/self-signed-certificate.yml
+++ b/roles/wordpress-setup/tasks/self-signed-certificate.yml
@@ -11,7 +11,7 @@
     -newkey rsa:2048 -days 3650 -nodes -x509 -sha256
     -keyout {{ item.key }}.key -out {{ item.key }}.cert
   args:
-    chdir: "{{ nginx_path }}/ssl"
+    chdir: "{{ nginx_ssl_path }}"
     creates: "{{ item.key }}.*"
   with_dict: "{{ wordpress_sites }}"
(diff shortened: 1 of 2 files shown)


scherii:
Could close #809

TangRufus:
Should we add redirect domains as well?

scherii:
@TangRufus With this method you'd specify the canonical domains in the wordpress_sites dict. And if I understood your question correctly you'd put the redirects in there as well. But this depends on the use case and has nothing to do with generating self-signed certificates. 🙂

TangRufus:
Consider this setup from the docs:

```yaml
# multiple hosts and redirects are possible
example.com:
  site_hosts:
    - canonical: example.com
      redirects:
        - www.example.com
        - site.com
```

Chrome stops `www.example.com` and `site.com` at the very beginning because of HTTPS error and would not redirect to `example.com`

scherii:
In this case I'd propose to add both `site.com` and `www.example.com` to the canonical domains.
I might be wrong – any other opinions?

fullyint:
closed by #812
Thank you @scherii for your excellent work on this PR and improving the self-signed certs for Trellis!

PR #823, opened 2017-04-08
Add vault_wordpress_sites validations

swalkinshaw:
This will catch a common error of forgetting to update site names/keys in both `wordpress_sites` and `vault_wordpress_sites`.

Example output:

```plain
Invalid WordPress sites configuration: site names in `wordpress_sites` must
have matching entry in `vault_wordpress_sites`.

Sites without a matching vault entry:
* `example.com`

Update `group_vars/development/vault_wordpress_sites.yml` to continue.

Docs: https://roots.io/trellis/docs/wordpress-sites/#passwordssecrets
```


fullyint:
Works in my tests!

🏆  So many will be spared the frustration.

PR #824, opened 2017-04-08
Fix #439 - Add ansible_local support for non-Windows

### LINKED ISSUE #439: Use ansible_local provisioner on all platforms
Vagrant's latest 1.8 release adds an [ansible_local](https://docs.vagrantup.com/v2/provisioning/ansible_local.html) provisioner which should be able to replace the functionality of the windows provisioner shell script.

swalkinshaw:
Previously only Windows would use Vagrant's `ansible_local` provisioner. But it's a much easier experience to get started in development since it skips the Ansible requirement.

This automatically uses the `ansible_local` provisioner if `ansible-playbook` does not exist on the host machine.

The env var `FORCE_ANSIBLE_LOCAL` can be set to force it as well.

Docs: https://github.com/roots/docs/pull/90

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -91,11 +91,11 @@ Vagrant.configure('2') do |config|
     end
   end
 
-  provisioner = Vagrant::Util::Platform.windows? ? :ansible_local : :ansible
-  provisioning_path = Vagrant::Util::Platform.windows? ? ANSIBLE_PATH_ON_VM : ANSIBLE_PATH
+  provisioner = local_provisioning? ? :ansible_local : :ansible
+  provisioning_path = local_provisioning? ? ANSIBLE_PATH_ON_VM : ANSIBLE_PATH
 
   config.vm.provision provisioner do |ansible|
-    if Vagrant::Util::Platform.windows?
+    if local_provisioning?
       ansible.install_mode = 'pip'
       ansible.provisioning_path = provisioning_path
       ansible.version = vconfig.fetch('vagrant_ansible_version')
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Add ansible_local support for non-Windows ([#824](https://github.com/roots/trellis/pull/824))
 * Add `fastcgi_read_timeout` to Nginx config ([#860](https://github.com/roots/trellis/pull/860))
 * Accommodate child themes: Update WP `stylesheet_root` separately ([#850](https://github.com/roots/trellis/pull/850))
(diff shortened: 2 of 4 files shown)


swalkinshaw:
Right now this creates a needless extra synced folder. It probably needs some changes which were done specifically for Windows.

```plain
    default: /vagrant => /Users/scott/dev/trellis
    default: /home/vagrant/trellis => /Users/scott/dev/trellis
    default: /home/vagrant/trellis/bin => /Users/scott/dev/trellis/bin
```

Vagrantfile
```diff
@@ -88,35 +89,33 @@ Vagrant.configure('2') do |config|
...
 
     ansible.groups = {
       'web' => ['default'],
       'development' => ['default']
     }
 
-    if tags = ENV['ANSIBLE_TAGS']
-      ansible.tags = tags
-    end
+    ansible.tags = tags if ENV['ANSIBLE_TAGS']
```
fullyint: `ansible.tags = ENV['ANSIBLE_TAGS']`

Vagrantfile
```diff
@@ -171,12 +174,24 @@ def post_up_message
...
+
+  ENV['PATH'].split(File::PATH_SEPARATOR).each do |path|
+    exts.each do |ext|
+      exe = File.join(path, "#{cmd}#{ext}")
+      return exe if File.executable?(exe) && !File.directory?(exe)
+    end
+  end
+
+  nil
+end
```
fullyint: Here's an issue that you may figure should be addressed by the user, not by Trellis.

I use pyenv to manage python versions on my machine. Even after uninstalling Ansible, a shim remains at `/usr/local/var/pyenv/shims/ansible-playbook`, causing an incorrect return value from this `which(cmd)`.

Here's a potential change that resolves it on my machine (dev/null handling from [here](http://stackoverflow.com/a/35142176/6847025)):
```diff
- return exe if File.executable?(exe) && !File.directory?(exe)
+ return system("#{exe} --help", [:out, :err] => File::NULL) if File.executable?(exe) && !File.directory?(exe)
```
This appeals to me because it tests whether the `ansible-playbook` command actually works rather than just whether it should work, the latter being assumed based on the presence of the `exe`. I don't know whether this revised approach would be good practice or acceptable ruby.

It the change above is applied, it returns `true`, `false`, or `nil`, if I recall correctly (instead of `exe` path). So, maybe the name `which()` should change to something like `cmd_valid(cmd)`.

fullyint:
I just tested this latest 3ca2d19 and it works great. Gets around that [pyenv issue](https://github.com/roots/trellis/pull/824#discussion_r111023178) I had with a previous iteration.

The only other thing I can think of is that the README "Requirements" section could say that Ansible is optional.

PR #825, opened 2017-04-09
Remove db_import option

swalkinshaw:
This option is used to import a local SQL dump file into the database.

It was seldom used and could be dangerous. This option would cause the DB to be imported *every* provision which could overwrite data.

As an alternative for most DB operations, we always recommend using WP CLI.

Example: `wp db import dbname.sql`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+### HEAD
+* Remove potentially dangerous `db_import` option ([#825](https://github.com/roots/trellis/pull/825))
+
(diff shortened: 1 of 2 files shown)


QWp6t:
For those who were using this to populate production server with local data, check out this post and discussion instead:
- post: https://roots.io/leveraging-wp-cli-aliases-in-your-wordpress-development-workflow/
- discussion: https://discourse.roots.io/t/leveraging-wp-cli-aliases-in-your-wordpress-development-workflow/8414

PR #827, opened 2017-04-12
Deploys: Make wp core update-db optional

fullyint:
On deploys, users with large network databases may want to disable the `wp core update-db` option to avoid the long processing time.

This PR adds the `update_db_on_deploy` option variable (default `true`) to control whether the `update-db` runs. This var may be defined per site in `wordpress_site` or just generally in `group_vars`.

This PR also consolidates the two "Update WP database" tasks, where the only difference was the presence of the `--network` flag for multisite.


swalkinshaw:
👍

PR #828, opened 2017-04-14
Vagrant config

swalkinshaw:
Inspired by https://github.com/geerlingguy/drupal-vm

* Move methods into separate file
* Intro Vagrant config

Todo:
- [x] Better name for `vagrant.yml`. drupal-vm uses `default.config.yml`?
- [x] Support for untracked config file?

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,43 +1,28 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
-require 'yaml'
-
-ip = '192.168.50.5' # pick any local IP
-cpus = 1
-memory = 1024 # in MB
-
 ANSIBLE_PATH = __dir__ # absolute path to Ansible directory on host machine
 ANSIBLE_PATH_ON_VM = '/home/vagrant/trellis' # absolute path to Ansible directory on virtual machine
 
-# Set Ansible paths relative to Ansible directory
-ENV['ANSIBLE_CONFIG'] = ANSIBLE_PATH
-ENV['ANSIBLE_CALLBACK_PLUGINS'] = "~/.ansible/plugins/callback_plugins/:/usr/share/ansible_plugins/callback_plugins:#{File.join(ANSIBLE_PATH, 'lib/trellis/plugins/callback')}"
-ENV['ANSIBLE_FILTER_PLUGINS'] = "~/.ansible/plugins/filter_plugins/:/usr/share/ansible_plugins/filter_plugins:#{File.join(ANSIBLE_PATH, 'lib/trellis/plugins/filter')}"
-ENV['ANSIBLE_LIBRARY'] = "/usr/share/ansible:#{File.join(ANSIBLE_PATH, 'lib/trellis/modules')}"
-ENV['ANSIBLE_ROLES_PATH'] = File.join(ANSIBLE_PATH, 'vendor', 'roles')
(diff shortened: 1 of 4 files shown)


mAAdhaTTah:
Can I make a suggestion: I edit my local Vagrantfile to mount WP plugins I'm working on into the VM. Can the configuration be edited to allow one to include a src/dest array:

```
mount_dirs:
    - src: 'path/to/plugin'
      dest: 'path/to/wp-content/plugins'
```

or something like this?

swalkinshaw:
@mAAdhaTTah sure, makes sense 👍

swalkinshaw:
@mAAdhaTTah added, try it out 👼

lib/trellis/vagrant.rb
```diff
@@ -0,0 +1,64 @@
...
+end
+
+def multisite_subdomains?(wordpress_sites)
+  wordpress_sites.any? do |(_name, site)|
+    site['multisite'].fetch('enabled', false) && site['multisite'].fetch('subdomains', false)
+  end
+end
+
+def nfs_path(site_name)
+  "/vagrant-nfs-#{site_name}"
```
fullyint: This could be used for nfs path names in the `vagrant_synced_folders` loop:

```diff
- def nfs_path(site_name)
-   "/vagrant-nfs-#{site_name}"
+ def nfs_path(path)
+   "/vagrant-nfs-#{File.basename(path)}"
  end
```

Vagrantfile
```diff
@@ -88,37 +66,49 @@ Vagrant.configure('2') do |config|
...
+    options = {
+      type: folder.fetch('type', 'nfs'),
+      create: folder.fetch('create', false),
+      mount_options: folder.fetch('mount_options', [])
+    }
+
+    config.vm.synced_folder folder['local_path'], folder['destination'], options
+
+    if folder.fetch('bindfs', true)
+      config.bindfs.bind_folder folder['local_path'], folder['destination'], options
```
fullyint: When using vagrant-bindfs, the convention appears to be...
1. do a regular vagrant folder sync from `somedir` to `/vagrant/somedir-nfs`
2. do a bindfs sync from `/vagrant/somedir-nfs` to `/final_path_on_vm/somedir`

Combined with the `File.basename()` adjustment to `nfs_path()` proposed in a separate comment, the following seems to handle it.
```diff
+ destination_folder = folder.fetch('bindfs', true) ? nfs_path(folder['destination']) : folder['destination']

- config.vm.synced_folder folder['local_path'], folder['destination'], options
+ config.vm.synced_folder folder['local_path'], destination_folder, options

  if folder.fetch('bindfs', true)
-   config.bindfs.bind_folder folder['local_path'], folder['destination'], options
+   config.bindfs.bind_folder destination_folder, folder['destination'], folder.fetch('bindfs_options', {})
```
The other piece to this is the `bindfs_options`, which are different from a regular vagrant sync dir `options` and must be treated differently. The above appears to successfully handle the following example 

```
# vagrant.local.yml
vagrant_synced_folders:
  - local_path: ../db
    destination: /home/vagrant/db
    create: false
    type: nfs
    bindfs: true
    bindfs_options:
      o: 'nonempty'
      p: '0644,a+D'
  - local_path: ../test
    destination: /home/vagrant/test
    create: false
    type: virtualbox
    bindfs: false
    mount_options:
      - 'dmode=755'
      - 'fmode=644'
```

swalkinshaw: 👌 updated

fullyint:
:+1: Looks great to me!

PR #832, opened 2017-05-01
skip plugins with WP-CLI, to avoid crappy plugins breaking deploys

alkymst:


retlehs:
:+1: ran into this with the woocommerce subscriptions plugin activated on a site

PR #837, opened 2017-06-09
Install WP-CLI packages

QWp6t:


fullyint:
Looks good to me!

roles/wp-cli/tasks/main.yml
```diff
@@ -23,3 +23,10 @@
     warn: false
   register: wp_cli_completion
   changed_when: wp_cli_completion.stdout == 'wp-completion-' + wp_cli_version + '.bash'
+
+- name: Install WP-CLI packages
+  command: "wp package install {{ item }}"
+  become_user: "{{ web_user }}"
+  register: wp_cli_packages_installed
+  changed_when: "'Nothing to install or update' not in wp_cli_packages_installed.stdout"
```
fullyint: While testing, I noticed that in some cases `stdout` for already-installed packages did not include [`'Nothing to install or update'`](https://github.com/composer/composer/blob/9aadbe27283bda0ed0f149fa4304c245f517c369/src/Composer/Installer.php#L490) but instead included [`'Package operations: 0 installs, 0 updates, 0 removals'`](https://github.com/composer/composer/blob/9aadbe27283bda0ed0f149fa4304c245f517c369/src/Composer/Installer.php#L520). This meant that sometimes Ansible inaccurately reported `changed` for packages that were already installed.

How does this modification look to you?
```diff
- changed_when: "'Nothing to install or update' not in wp_cli_packages_installed.stdout"
+ changed_when:
+   - "'Nothing to install or update' not in wp_cli_packages_installed.stdout"
+   - "'Package operations: 0 installs, 0 updates, 0 removals' not in wp_cli_packages_installed.stdout"
```
I would have preferred to be more concise by checking for a string that is unique to when packages are freshly installed but I didn't observe any such string.

fullyint:
thank you!

PR #841, opened 2017-06-17
Remove unnecessary output about running WP-CLI and Composer

partounian:
I removed this line (post vagrant up) as it originally confused me because both of these commands run automatically.
`==> default: * Composer and WP-CLI commands need to be run on the virtual machine.`

### DIFF
--- a/lib/trellis/vagrant.rb
+++ b/lib/trellis/vagrant.rb
@@ -78,7 +78,8 @@ end
 
 def post_up_message
   msg = 'Your Trellis Vagrant box is ready to use!'
-  msg << "\n* Composer and WP-CLI commands need to be run on the virtual machine."
+  msg << "\n* Composer and WP-CLI commands need to be run on the virtual machine"
+  msg << "\n  for any post-provision modifications."
   msg << "\n* You can SSH into the machine with `vagrant ssh`."
   msg << "\n* Then navigate to your WordPress sites at `/srv/www`"
   msg << "\n  or to your Trellis files at `#{ANSIBLE_PATH_ON_VM}`."


swalkinshaw:
That line is telling that any *future* manually-run commands need to be done on the VM. It was a common mistake that people would try and run them on their host.

I'm fine with improving the wording, but I think some form of it should stay.

partounian:
@swalkinshaw How does that look?

lib/trellis/vagrant.rb
```diff
@@ -78,7 +78,8 @@ def nfs_path(path)
 
 def post_up_message
   msg = 'Your Trellis Vagrant box is ready to use!'
-  msg << "\n* Composer and WP-CLI commands need to be run on the virtual machine."
+  msg << "\n* Composer and WP-CLI commands need to be run on the virtual machine,"
```
QWp6t: Comma isn't necessary here.

lib/trellis/vagrant.rb
```diff
@@ -78,7 +78,8 @@ def nfs_path(path)
...
+  msg << "\n* for any post-provision modifications."
```
QWp6t: This shouldn't be a new bullet point. (replace asterisk `*` with space ` `)

CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 ### HEAD
+* Remove unnecessary post vagrant up output ([#841](https://github.com/roots/trellis/pull/841))
```
swalkinshaw: Don't need this changelog entry either anymore.

partounian:
Thank you both for the quick feedback. I have updated the commit accordingly.

swalkinshaw:
thanks 👍

PR #842, opened 2017-06-19
Add support for SMTP without authentication

perifer:
I admit this is an edge case. I have a client that provided their own internal smtp-server that had no authentication (it was IP-blocked). By not providing user and password I got it to work.


swalkinshaw:
Edge case but I think it makes sense 👍  Thanks

partounian:
@perifer Just to confirm the only changes you made were having an empty username in `group_vars/all/mail.yml` and empty password in `group_vars/all/vault.yml`?  Or did you also change `ssmtp_auth_method:` in `roles/ssmtp/defaults/main.yml` etc?

perifer:
@partounian only empty password and empty username!

PR #846, opened 2017-07-08
Fix incorrect file path in vault error template

greatislander:
When a vault entry is missing for a site in `wordpress_sites.yml`, [this template](https://github.com/roots/trellis/blob/master/roles/common/templates/wordpress_sites.j2) is used to display an error message. However, the error message incorrectly instructs the user to update the nonexistent `group_vars/<environment>/vault_wordpress_sites.yml` file instead of `group_vars/<environment>/vault.yml`. This PR fixes this issue.


swalkinshaw:
Good catch, thanks 🚀

PR #848, opened 2017-07-14
Update WP template_root for multisite subsites when applicable

fullyint:
Followup to #840 and its history. Fixes https://discourse.roots.io/t/multisite-subsite-using-twentyfifteen-theme/9871/5

This PR enables Trellis deploys to update `template_root` and `stylesheet_root` WP options also for multisite subsites, not just the primary site.

Multisite operators that wish to disable all this (e.g., in order to avoid the processing time) may add `update_wp_theme_paths: false` somewhere in group_vars, or disable it on the fly:
```
ansible-playbook deploy.yml -e env=production -e site=example.com -e update_wp_theme_paths=false
```

Re: `xargs` etc, see also https://danielbachhuber.com/tip/run-wp-cli-command-wordpress-multisite/

Re: `bool` filter, it is necessary for proper interpretation of cli `-e update_wp_theme_paths=false` (otherwise the var remains a string "false" interpreted as boolean true).

No longer assumes that `template_root` and `stylesheet_root` are within a `.../web/wp/wp-content/themes` subdirectory of the `releases_directory` (although I don't know where else they'd be).

### DIFF
--- a/roles/deploy/hooks/finalize-after.yml
+++ b/roles/deploy/hooks/finalize-after.yml
@@ -16,7 +16,7 @@
       {% if not project.multisite.enabled | default(false) %}
         wp option get template_root --skip-plugins
       {% else %}
-        wp site list --field=url | xargs -I {} bash -c 'export url="{}"; echo -n "$url " && wp option get template_root --skip-plugins --url=$url'
+        wp site list --field=url | xargs -I {} bash -c 'export url="{}"; echo -n "$url " && wp option get template_root --skip-plugins --url=$url || echo'
       {% endif %}
     args:
       chdir: "{{ deploy_helper.current_path }}"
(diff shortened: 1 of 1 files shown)


fullyint:
The third commit `Accommodate child themes: Update WP stylesheet_root separately` addresses https://discourse.roots.io/t/template-root-not-updated-by-env/9724

When a child theme is in Bedrock `web/app/themes` and its parent theme is in different directory (e.g., `web/wp/wp-content/themes`), `stylesheet_root` must be `/themes` while `template_root` should be updated with the latest deploy releases_path. Prior to this commit, `stylesheet_root` was always updated with `template_root`, causing problems for some child themes.

fullyint:
The fourth commit `Deploys: --skip-themes when updating WP template_root` addresses https://discourse.roots.io/t/deploy-failing-at-task-deploy-get-wp-theme-template-root/10017

Some WP themes may throw errors if loaded in this context, e.g., some PHP classes may not be available with `--skip-plugins`. It is not necessary to load themes when updating WP options for `template_root` and `stylesheet_root`, so it is safer to `--skip-themes`.

Trellis already uses `--skip-themes` in conjunction with `--skip-plugins` in the [`WordPress Installed?`](https://github.com/roots/trellis/blob/2ae7cf189ff6305aa22226bfa1825881862d879e/roles/deploy/hooks/finalize-before.yml#L7-L8) task.

fullyint:
Closing. Will split into small chunks for easier testing and review.

PR #849, opened 2017-07-20
Deploys: --skip-themes when updating WP template_root

fullyint:
This PR is a piece of the former #848 and addresses https://discourse.roots.io/t/deploy-failing-at-task-deploy-get-wp-theme-template-root/10017

Some WP themes may throw errors if loaded in this context, e.g., some PHP classes may not be available with `--skip-plugins`. It is not necessary to load themes when updating WP options for `template_root` and `stylesheet_root`, so it is safer to `--skip-themes`.

Trellis already uses `--skip-themes` in conjunction with `--skip-plugins` in the [`WordPress Installed?`](https://github.com/roots/trellis/blob/2ae7cf189ff6305aa22226bfa1825881862d879e/roles/deploy/hooks/finalize-before.yml#L7-L8) task.


swalkinshaw:
:shipit:

PR #850, opened 2017-07-21
Accommodate child themes: Update WP stylesheet_root separately

fullyint:
This PR is a piece of the former #848 and addresses https://discourse.roots.io/t/template-root-not-updated-by-env/9724

When a child theme is in Bedrock `web/app/themes` and its parent theme is in different directory (e.g., `web/wp/wp-content/themes`), `stylesheet_root` must be `/themes` while `template_root` should be updated with the latest deploy releases_path. Prior to this commit, `stylesheet_root` was always updated with `template_root`, causing problems for some child themes.

----

Regarding the `regex_replace`: Till now Trellis has assumed that if a `template_root`includes the releases path (e.g., /srv/www/example.com/releases) then we can replace the entire `template_root` using the new release path + `/web/wp/wp-content/themes`. Maybe that's a safe assumption, but at my level of familiarity, the only thing I feel safe saying is that "I know the release path portion can be replaced with the new release path, but I don't know about the rest of the path." 

That's why you'll notice that I made the following replacement (pseudo code):
```diff
- wp option set {{ <option> }} {{ deploy_helper.new_release_path }}/web/wp/wp-content/themes
+ wp option set {{ <option> }} <old_full_path> | regex_replace(releases_path + '/[^/]+(.*)', new_release_path + '\1')
```

The only weakness I can think of for this change is that it is less readable. Maybe someone who knows more will say that the original assumption was safe, implying that this change is unnecessary.

### DIFF
--- a/roles/deploy/hooks/finalize-after.yml
+++ b/roles/deploy/hooks/finalize-after.yml
@@ -11,22 +11,25 @@
       chdir: "{{ deploy_helper.current_path }}"
     when: project.update_db_on_deploy | default(update_db_on_deploy)
 
-  - name: Get WP theme template root
-    command: wp option get template_root --skip-plugins --skip-themes
+  - name: Get WP theme template and stylesheet roots
+    command: wp option get {{ item }} --skip-plugins --skip-themes
     args:
       chdir: "{{ deploy_helper.current_path }}"
     register: wp_template_root
     changed_when: false
     failed_when: wp_template_root.stderr != ""
+    with_items:
+      - template_root
+      - stylesheet_root
 
   - name: Update WP theme paths
-    command: wp option set {{ item }} {{ deploy_helper.new_release_path }}/web/wp/wp-content/themes
+    command: >
+      wp option set {{ item.item }}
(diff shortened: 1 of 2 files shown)


roles/deploy/hooks/finalize-after.yml
```diff
@@ -11,22 +11,25 @@
...
     failed_when: wp_template_root.stderr != ""
+    with_items:
+      - template_root
+      - stylesheet_root
 
   - name: Update WP theme paths
-    command: wp option set {{ item }} {{ deploy_helper.new_release_path }}/web/wp/wp-content/themes
+    command: >
+      wp option set {{ item.item }}
+      {{ item.stdout | regex_replace(deploy_helper.releases_path + '/[^/]+(.*)', deploy_helper.new_release_path + '\1') }}
```
swalkinshaw: What about using `deploy_helper.previous_release_path` and just replace with `deploy_helper.new_release_path`?

fullyint: I figured users could have `template_root` values with release paths different from `previous_release_path` but that still needed to be replaced. One example might be if a user imports a DB from staging where the `previous_release_path` was different than here in production. Or, perhaps the prior release was failed and `template_root` never updated but deploy helper has that failed path as `previous_release_path`? I didn't test if this latter could happen.

PR #855, opened 2017-07-31
Nginx: Skip caching WP API

TangRufus:

- [ ] This is a bug report

---

Nginx caches WP API responses and serving them to everyone without checking authorizations.

Nginx should skip caching WP API responses


swalkinshaw:
Makes sense 👍

mAAdhaTTah:
It is possible to change the prefix programmatically with a filter in WordPress. Does that need to be taken into consideration here?

TangRufus:
@mAAdhaTTah 
When changing wp api prefix with filters, they need to follow https://roots.io/trellis/docs/fastcgi-caching/

```yaml
cache:
  skip_cache_uri: <Change here>
```

Maybe we need to update the docs?

PR #856, opened 2017-08-01
Allow overriding php-fpm params by ansible variables

TangRufus:
Allow overriding php-fpm params by ansible variables.

Related: https://discourse.roots.io/t/overriding-php-fpm-config-best-practice/10101


TangRufus:
When adding this to the docs, I think we should add links to tutorials about finding these php fpm pm values. Setting them too high would raise memory issues.

swalkinshaw:
These defaults might be better defined in `roles/php/defaults/main.yml`. This template already uses `php_display_errors` which is from there.

@fullyint do you have an opinion on this?

fullyint:
I tend to think that these variable defaults are appropriate here in the `wordpress-setup` role given that the vars are used in `wordpress-setup` role and not in the `php` role. In contrast, it appears that `php_display_errors` is used in both roles. 

Although users could intuit that `roles/php/defaults/main.yml` holds all the php configs, I think the real function of a role's `default/main.yml` is to hold defaults for variables used in the role, not to hold configs for a service generally (e.g., php generally).

For comparison, I added some nginx-related vars to `roles/wordpress-setup/defaults` in the past with the justification that the vars were only used in the `wordpress-setup` role.

swalkinshaw:
Thanks @TangRufus 👍

partounian:
This is very helpful as I was going to modify the values directly myself also. Quick question, what is the reason we use a 3rd party PHP 7.1 repo instead of the official one?

PR #858, opened 2017-08-07
Allow customization of the nginx package name

reidab:
The nginx/development PPA includes several variants of nginx each built
with different collections of modules: nginx-core (default), nginx-light,
nginx-full, and nginx-extras.

This small change allows the nginx package installed by Trellis to be
customized via the `nginx_package` variable.

### DIFF
--- a/roles/nginx/defaults/main.yml
+++ b/roles/nginx/defaults/main.yml
@@ -1,4 +1,5 @@
 ---
+nginx_package: nginx
 nginx_conf: nginx.conf.j2
 nginx_path: /etc/nginx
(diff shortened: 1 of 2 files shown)


TangRufus:
Should we make apt repo customisable as well?

```yaml
- name: Add Nginx PPA
  apt_repository:
    repo: "ppa:nginx/development" <------ this one
    update_cache: yes
```

reidab:
@TangRufus I think so. I almost did this initially — I'll update shortly.

swalkinshaw:
Thanks 🚀

PR #859, opened 2017-08-07
Load modules-enabled config files

reidab:
The nginx package was updated earlier this year to include
`modules-available` and `modules-enabled` directories in `/etc/nginx`.

This adds loading of conf files from modules-enabled to the top of
the trellis nginx.conf (in a new `modules_enabled` block).


swalkinshaw:
Didn't know about this. Thanks 👍

PR #861, opened 2017-08-12
Update vagrant box version

mockey:
update `vagrant_box_version` to current bento/ubuntu-16.04 version v2.3.8


swalkinshaw:
Will Vagrant even prompt to update the box with this logic?

`2.2.9` satisfies `<= 2.3.8`

mockey:
Yes, I got the update message on `vagrant up` here. I tested with 2.3.7 and then with 2.3.8.
It does not work with `2.3.8` straight. Even update doesn't work then.
That's why I went for `<= 2.3.8`. It' a bit strange, though, I agree.

swalkinshaw:
Tried it out and works great. Thanks 👍

PR #862, opened 2017-08-12
Only include *.conf files in sites-enabled/

partounian:
If for any reason you need to backup a host (i.e `mv website.com.conf website.com.conf.bak`) then this allow you to name it anything that doesn't end in `.conf`.


swalkinshaw:
Think this makes sense.

We do `include includes.d/{{ item.key }}/*.conf;` elsewhere.

PR #866, opened 2017-08-15
Nginx: Compile dynamic modules

TangRufus:
Note:
- it doesn't compile or install Nginx
- it assumes dynamic module sources are git repos
- it recompiles if Nginx version or dynamic module repos changed
- it skips when `nginx_dynamic_modules` is empty

Usage (see #867): 
```yaml
# group_vars/all/main.yml
nginx_dynamic_modules:
  ngx_brotli:
    repo: https://github.com/google/ngx_brotli.git
    objs:
      - ngx_http_brotli_filter_module.so
      - ngx_http_brotli_static_module.so
```

Related: #863

Issue #863, opened 2017-08-12
Feature Request: Add Nginx Brotli


TangRufus:
## Submit a feature request or bug report

- [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/guidelines/blob/master/CONTRIBUTING.md)
- [x] This is a feature request
- [ ] This is a bug report
- [x] This request isn't a duplicate of an [existing issue](https://github.com/roots/trellis/issues)
- [x] I've read the [docs](https://roots.io/trellis/docs) and followed them (if applicable)
- [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/c/trellis) forums

Replace any `X` with your information.

---

**What is the current behavior?**

No Brotli support

**What is the expected or desired behavior?**

Add [Brotli module](https://github.com/google/ngx_brotli) to Nginx

---

## Feature Request

What I have in mind is more less the same as [this article](https://certsimple.com/blog/nginx-brotli):

1. Install ngx_brotli 
2. Load ngx_brotli filter module to nginx
3. Load ngx_brotli static module to nginx
3. Add brotli directives to nginx.conf
4. Compress *.br files during deploy

swalkinshaw:
This isn't really possible right now unfortunately. We don't compile Nginx from source so we can't add compiled modules like these.

There's been some threads on our forum about various modules. Here's an example: https://discourse.roots.io/t/trellis-build-nginx-from-source-ngx-pagespeed-enjoy/6391

The more basic feature request would be: support compiling Nginx from source. Then anyone could specify other modules they want.

partounian:
This guide has some great tips and on top #2 it links a PPA you can use and dynamically link the brotli module which seems like a better idea than compiling. Note: I did not verify the PPA so there may be better alternatives.
https://certsimple.com/blog/nginx-brotli

swalkinshaw:
Nginx just got support for customizing the ppa and package name: https://github.com/roots/trellis/pull/858/files

So you can easily try this yourself now :)

partounian:
Yup I wrote that comment with that PR in mind but I never thought to link it, oops.

strarsis:
@TangRufus: Wouldn't this be included in mod_pagespeed?

TangRufus:
Closed in 2021.

It is 2021. This should be handled by a CDN or similar.

PR #867, opened 2017-08-15
Nginx: Add ngx_brotli

TangRufus:
Depends on: #866 
Related: #863

### DIFF
--- a/roles/nginx/defaults/main.yml
+++ b/roles/nginx/defaults/main.yml
@@ -17,6 +17,18 @@ nginx_cache_key_storage_size: 10m
 nginx_cache_size: 250m
 nginx_cache_inactive: 1h
 
+# Brotli
+nginx_brotli_static: 'off'
+nginx_brotli: 'on'
+nginx_brotli_comp_level: 4
+nginx_brotli_window: 512k
+nginx_brotli_min_length: 20
+
 # Dynamic modules
 nginx_modules_path: /usr/share/nginx/modules
-nginx_dynamic_modules: {}
+nginx_dynamic_modules:
+  ngx_brotli:
+    repo: https://github.com/google/ngx_brotli.git
+    objs:
+      - ngx_http_brotli_filter_module.so
+      - ngx_http_brotli_static_module.so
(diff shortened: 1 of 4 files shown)


partounian:
I suggest removing the dynamic module from this PR but just leaving your statement that is depends on #866. Also, if we are making gzip static why not brotli?

TangRufus:
> Also, if we are making gzip static why not brotli?

Compressing .br takes much longer than gzip. Better wait for #865 and think of a better way to do so.

TangRufus:
Will this be consider to be merged, or should I extract it to be a galaxy role?

Thanks!

swalkinshaw:
I wish this didn't require all the dynamic module support since Brotli is a nice improvement. So since it depends on the other, still thinking about it.

TangRufus:
Dynamic module seems less intrusive because it doesn't require a custom Nginx package.

Another benefit is that it enable developers to use more Nginx modules if needed while keeping the official Nginx package intact. (Yes, i am talking about third party roles again...)

Let me know how the Roots team think about it. This pull request needs rebase.

wilmardo:
Any update on this or #866 @swalkinshaw? In my humble opinion Brotli support would be an awesome addition to Trellis since it reduces load time by a bunch :)

swalkinshaw:
Not yet sorry. Have other priorities in Trellis to finish first.

eshimischi:
Instead of all these way better to install PageSpeed and make a right configuration..

partounian:
Just wanted to say Cloudflare has brotli on by default.

TangRufus:
This pull request was made some time ago.
Cloudflare now uses gzip level 8 and brotli by default.

If you are using Cloudflare, you don't need this pull request.

eshimischi:
Have you ever heard of SDCH ? [Read here](https://mikandi.com/blog/tips-and-tricks/press-me-a-sdch-implementing-shared-dictionary-compression/)

wilmardo:
@eshimischi SDCH has been deprecated some time ago: https://www.chromestatus.com/feature/5763176272494592

eshimischi:
> @eshimischi SDCH has been deprecated some time ago: https://www.chromestatus.com/feature/5763176272494592

Oh, missed it. thanks, let's forget about it than

PR #869, opened 2017-08-16
WordPress Setup: Add Nginx ssl_client_certificate

TangRufus:
Usage:
```yaml
wordpress_sites:
  example.com:
    site_hosts:
      - canonical: example.com
    ssl:
      enabled: true
      provider: self-signed
      client_cert_url: 'https://support.cloudflare.com/hc/en-us/article_attachments/201243967/origin-pull-ca.pem'
```

See: 
- https://support.cloudflare.com/hc/en-us/articles/204494148-Setting-up-NGINX-to-use-TLS-Authenticated-Origin-Pulls
- https://blog.cloudflare.com/protecting-the-origin-with-tls-authenticated-origin-pulls/

### DIFF
--- a/roles/wordpress-setup/tasks/nginx-client-cert.yml
+++ b/roles/wordpress-setup/tasks/nginx-client-cert.yml
@@ -1,0 +1,8 @@
+---
+- name: Download client cert
+  get_url:
+    url: "{{ item.value.ssl.client_cert_url }}"
+    dest: "{{ nginx_ssl_path }}/client-{{ (item.value.ssl.client_cert_url | hash('md5'))[:7] }}.crt"
+    mode: 0640
+  with_dict: "{{ wordpress_sites }}"
+  when: ssl_enabled and item.value.ssl.client_cert_url is defined
--- a/roles/wordpress-setup/templates/wordpress-site.conf.j2
+++ b/roles/wordpress-setup/templates/wordpress-site.conf.j2
@@ -80,6 +80,11 @@ server {
 
   add_header Strict-Transport-Security "max-age={{ [hsts_max_age, hsts_include_subdomains, hsts_preload] | reject('none') | join('; ') }}";
 
+  {% if item.value.ssl.client_cert_url is defined -%}
+  ssl_verify_client       on;
+  ssl_client_certificate  {{ nginx_ssl_path }}/client-{{ (item.value.ssl.client_cert_url | hash('md5'))[:7] }}.crt;
+  {% endif -%}
+
   {% if item.value.ssl.provider | default('manual') == 'manual' and item.value.ssl.cert is defined and item.value.ssl.key is defined -%}
   ssl_certificate         {{ nginx_path }}/ssl/{{ item.value.ssl.cert | basename }};
   ssl_certificate_key     {{ nginx_path }}/ssl/{{ item.value.ssl.key | basename }};
--- a/roles/wordpress-setup/tasks/main.yml
+++ b/roles/wordpress-setup/tasks/main.yml
@@ -3,6 +3,8 @@
   tags: wordpress-setup-database
 - include: self-signed-certificate.yml
   tags: wordpress-setup-self-signed-certificate
+- include: nginx-client-cert.yml
+  tags: wordpress-setup-nginx-client-cert
 
 - name: Create web root
   file:


swalkinshaw:
Seems fairly simple and it's a generic Nginx setting.

My only concern is the downloading with `force` being set.

Ansible docs say
> Generally should be yes only for small local files. 

Not exactly sure what the alternative is though or how much it matters. Obviously Trellis "downloads" many things during a provision but this won't be idempotent.

fullyint:
**Idempotent reporting.** I was delightfully surprised that even with `force`, the `get_url` task reporting shows `ok` if the file is unchanged, versus always showing `changed`. I believe that to `force` download of the temp file every time enables calculating and comparing checksums.

**Copy module more flexible?** Could users want a `ssl_client_certificate` that is not accessible via URL? Using the `copy` module may accommodate more scenarios. We could let users specify a local path to their own `origin-pull-ca.pem` they've previously downloaded to their machine, as is done with the `provider: manual` certs and keys. Using the copy module would avoid downloading every time and the related potential for playbook failure due to network connectivity issues.

**Capturing updates via partial hash in dest filename.** If you stick with `get_url`, I assume the intent in using `force` is to detect and incorporate changes in the content at the URL. I think we could still capture updates **even when `force: no`** by appending to the `dest` filename a few characters from a hash of the URL. Ansible would detect that `dest` does not exist and proceed to `get_url`.

This assumes that the update would be at a new URL, NOT the same URL. The source URL doesn't strike me as official/canonical being in Cloudflare's support `article_attachments`: `https://support.cloudflare.com/hc/en-us/article_attachments/201243967/origin-pull-ca.pem`. I couldn't find any other URL. I suspect any updates to `origin-pull-ca.pem` could likely appear at a new URL.

TangRufus:
Updated:
- Using md5 hash as file name
- Not using `force` anymore, only download at the first provision
- Using `ssl_client_cert_url` as global default

Usage 1:
```yaml
wordpress_sites:
  example.com:
    ssl:
      enabled: true
      client_cert_url: 'https://example.com/origin-pull-ca.pem'
```

Usage 2:
```yaml
ssl_client_cert_url: 'https://another-example.com/origin-pull-ca.pem'

wordpress_sites:
  example.com:
    ssl:
      enabled: true
```

Usage 3:
```yaml
ssl_client_cert_url: 'https://another-example.com/origin-pull-ca.pem'

wordpress_sites:
  example.com:
    ssl:
      enabled: true
      client_cert_url: false
```

Usage 4:
```yaml
ssl_client_cert_url: 'https://another-example.com/origin-pull-ca.pem'

wordpress_sites:
  example.com:
    ssl:
      enabled: true
      client_cert_url: 'https://example.com/origin-pull-ca.pem'
```

About using `ansible copy module`: maybe another pull request for another time

Thanks!

roles/wordpress-setup/templates/wordpress-site.conf.j2
```diff
@@ -75,6 +75,15 @@ server {
 
   add_header Strict-Transport-Security "max-age={{ [hsts_max_age, hsts_include_subdomains, hsts_preload] | reject('none') | join('; ') }}";
 
+  {% if item.value.ssl.client_cert_url is defined and item.value.ssl.client_cert_url -%}
```
swalkinshaw: Is this supposed to be `ssl_client_cert_url` in the first one? Just confused why it's checking the same variable

roles/wordpress-setup/templates/wordpress-site.conf.j2
```diff
@@ -75,6 +75,15 @@ server {
...
+  ssl_client_certificate  {{ nginx_ssl_path }}/{{ item.value.ssl.client_cert_url | hash('md5') }};
```
swalkinshaw: There's two spaces before `{` 👾

roles/wordpress-setup/templates/wordpress-site.conf.j2
```diff
@@ -75,6 +75,15 @@ server {
...
+  ssl_verify_client on;
+
+  {% elif ssl_client_cert_url is defined and not item.value.ssl.client_cert_url is defined -%}
+  ssl_client_certificate  {{ nginx_ssl_path }}/{{ ssl_client_cert_url | hash('md5') }};
```
swalkinshaw: ditto here too

TangRufus:
Removed the global option.
2 spaces before `{` because of alignment:
trellis-nginx-ssl

roles/wordpress-setup/templates/wordpress-site.conf.j2
TangRufus: `ssl_client_cert_url` was the global option, removed now.

TangRufus:
Will this be consider to be merged, or should I extract it to be a galaxy role?

Thanks!

roles/wordpress-setup/tasks/nginx.yml
```diff
@@ -1,4 +1,8 @@
 ---
+- include: nginx-client-cert.yml
+  tags: wordpress-setup-nginx-client-cert
+  notify: reload nginx
+
```
fullyint: Two minor requests:
* Could you move this `include` task right after the other [similar include tasks](https://github.com/roots/trellis/blob/f2b81074c83475837e544a8aa5c3e909e760aa8a/roles/wordpress-setup/tasks/main.yml#L2-L5) in `tasks/main.yml`? That placement strikes me as more consistent. I like the idea of avoiding [nested includes](https://www.google.com/search?q=ansible+"nested+includes") when it's easy.
* Let's omit the `notify` here. Technically, Nginx only needs to reload if the conf file produced by our `wordpress-site.conf.j2` changes. A change in the "Download client cert" task will trigger a `changed` [template task](https://github.com/roots/trellis/blob/f2b81074c83475837e544a8aa5c3e909e760aa8a/roles/wordpress-setup/tasks/nginx.yml#L33) which already has `notify: reload nginx`.

roles/wordpress-setup/tasks/nginx-client-cert.yml
```diff
@@ -0,0 +1,8 @@
+---
+- name: Download client cert
+  get_url:
+    url: "{{ item.value.ssl.client_cert_url }}"
+    dest: "{{ nginx_ssl_path }}/{{ item.value.ssl.client_cert_url | hash('md5') }}"
```
fullyint: I think a user who happens upon the server's `nginx_ssl_path` contents would find the filename more intuitive as `client-4c1d4cc.crt` instead of `4c1d4ccad291c1ac1f230ad263c4196c`.

Could we make this change?
```diff
- dest: "{{ nginx_ssl_path }}/{{ item.value.ssl.client_cert_url | hash('md5') }}"
+ dest: "{{ nginx_ssl_path }}/client-{{ (item.value.ssl.client_cert_url | hash('md5'))[:7] }}.crt"
```

A similar filename change would be needed in `wordpress-site.conf.j2`.

roles/wordpress-setup/templates/wordpress-site.conf.j2
fullyint: With the global option removed, I presume users would no longer disable client verification by setting `client_cert_url: false`. Rather, they simply wouldn't define `client_cert_url` for the relevant site.

If we don't need to check the `bool` interpretation of `client_cert_url`, could we make this change?

```diff
- {% if item.value.ssl.client_cert_url is defined and item.value.ssl.client_cert_url -%}
+ {% if item.value.ssl.client_cert_url is defined -%}
```

TangRufus:
All set. Thanks!

fullyint:
Wonderful! Thank you @TangRufus!

Issue #830, opened 2017-04-17
add provisioning-hooks directory to operate like deploy-hooks


nbyloff:
## Submit a feature request or bug report

- [X] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/guidelines/blob/master/CONTRIBUTING.md)
- [X] This is a feature request
- [X] This request isn't a duplicate of an [existing issue](https://github.com/roots/trellis/issues)
- [X] I've read the [docs](https://roots.io/trellis/docs) and followed them (if applicable)
- [X] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/c/trellis) forums

Replace any `X` with your information.

---

**What is the current behavior?**

Any custom roles/tasks needed during provisioning must be added to the main server.yml, causing conflicts between projects and the trellis base when trying to keep it in sync locally.

**What is the expected or desired behavior?**

I think it would be a nice feature to just add custom.yml files to a directory that get automatically run during provisioning. So when merging the main trellis base into local projects, custom provisioning additions won't be lost when updating things like the main server.yml file.

---

## Feature Request

**Please provide use cases for changing the current behavior:**

In my example, I always add a custom backup job that zips and pushes the DB and uploads directory to an Amazon s3 bucket nightly.

**Other relevant information:**

Added this task to the end of the server.yml

````
---
- name: read custom provisioning hooks
  find:
    paths: "custom-roles"
    patterns: "*.yml"
  register: custom_tasks

- name: execute custom provisioning hooks
  include: "{{ item.path }}"
  with_items: "{{ custom_tasks.files }}"
````

Creating a `custom-roles` folder and dropped two YAML files in there. They executed with no errors. Something like that built in would help guys like me from shooting themselves in the foot.

swalkinshaw:
I'm not sure about this. For one thing, roles depend on order. The ugly solution to that is people naming roles like `1-whatever` which is not cool 😔 

One goal of Trellis is to still expose Ansible. I think it's important for people to understand the basic playbooks and this abstracts that away a little bit too.

nbyloff:
[...]

PR #870, opened 2017-08-17
New ssl provider: cloudflare-origin-ca

### LINKED ISSUE #868: Feature Request: New SSL provider CloudFlare Origin CA
Generate a cert by cli for `example.com, *.example.com, *.another.com` when this example config is given:

```yaml
wordpress_sites:
  example.com:
    site_hosts:
      - canonical: example.com
        redirects:
          - www.example.com
          - ww2.example.com
          - www.another.com
    ssl:
      enabled: true
      provider: cloudflare
```

It should:
- [ ] add cloudflare package repository https://pkg.cloudflare.com/
- [ ] `root@server# apt-get install cfca`
- [ ] calculate a list of hostnames, e.g.: `example.com, *.example.com, *.another.com`
- [ ] `root@server# CF_API_KEY=XXXXX cfca getcert -hostnames example.com,*.example.com,*.another.com -key-out /etc/nginx/ssl/cloudflare.key -certificate-out /etc/nginx/ssl/cloudflare.pem -overwrite`
- [ ] Update `/etc/nginx/nginx.conf` template

See:
- https://blog.cloudflare.com/cloudflare-ca-encryption-origin/
- https://support.cloudflare.com/hc/en-us/articles/217471977

TangRufus:
Usage:
```yaml
vault_cloudflare_origin_ca_api_key: 'my_api_key'

wordpress_sites:
  example.com:
    site_hosts:
      - canonical: example.com
        redirects:
          - hi.example.com
          - bye.example.com
    ssl:
      enabled: true
      provider: cloudflare-origin-ca
```

This will generate and upload (to Cloudflare) a cert for `example.com,hi.example.com,bye.example.com`

This is [Origin CA](https://blog.cloudflare.com/cloudflare-ca-encryption-origin/) not [Universal SSL](https://blog.cloudflare.com/introducing-universal-ssl/)

close #868

### DIFF
--- a/group_vars/all/main.yml
+++ b/group_vars/all/main.yml
@@ -7,6 +7,7 @@ ntp_manage_config: true
 www_root: /srv/www
 ip_whitelist:
   - "{{ (env == 'development') | ternary(ansible_default_ipv4.gateway, ipify_public_ip | default('')) }}"
+cloudflare_origin_ca_api_key: "{{ vault_cloudflare_origin_ca_api_key }}" # Define this variable in group_vars/all/vault.yml
 
 # Values of raw_vars will be wrapped in `{% raw %}` to avoid templating problems if values include `{%` and `{{`.
 # Will recurse dicts/lists. `*` is wildcard for one or more dict keys, list indices, or strings. Example:
@@ -18,3 +19,4 @@ raw_vars:
   - vault_users.*.password
   - vault_users.*.salt
   - vault_wordpress_sites
+  - vault_cloudflare_origin_ca_api_key
--- a/group_vars/all/vault.yml
+++ b/group_vars/all/vault.yml
@@ -1,2 +1,3 @@
 # Documentation: https://roots.io/trellis/docs/vault/
 vault_mail_password: smtp_password
+vault_cloudflare_origin_ca_api_key: example_api_key
--- a/roles/wordpress-setup/tasks/cloudflare-origin-ca.yml
+++ b/roles/wordpress-setup/tasks/cloudflare-origin-ca.yml
@@ -1,0 +1,37 @@
+---
+- name: Add Cloudflare key
+  apt_key:
+    url: 'https://pkg.cloudflare.com/pubkey.gpg'
+
+- name: Add Cloudflare PPA
+  apt_repository:
+    repo: 'deb http://pkg.cloudflare.com/ xenial main'
+    update_cache: yes
+
+- name: Install CFCA
+  apt:
+    name: cfca
+    state: present
+    force: yes
+
+- name: Create directoriy and set permission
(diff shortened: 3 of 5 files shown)


partounian:
Any update @swalkinshaw ?

swalkinshaw:
No, I'm undecided on this one. I don't know if Trellis should have a Cloudflare specific option like this built-in.

partounian:
No worries, really appreciate your updates through!

TangRufus:
perhaps publish it to ansible galaxy is a better idea

swalkinshaw:
Good question, I think it would use [Child templates](https://roots.io/trellis/docs/nginx-includes/#child-templates).

Right now there's an `https` block. The new template would use that block and call `{{ super() }}` then just add the new certificate stuff after that.

TangRufus:
Extracted to https://github.com/TypistTech/trellis-cloudflare-origin-ca

swalkinshaw:
@TangRufus thanks, took a quick look at that and it looks great.

This is one of the first 3rd party roles like this so it was interesting to see how the implementation went. Most of it looks fairly straightforward. Other than the playbook hack 😔 .

Something like https://github.com/roots/trellis/issues/830 might be able to help that.

TangRufus:
Other than #830, we need better way for galaxy roles to override Nginx templates.

Telling users use templats in vendor doesn't feel right:
```yaml
nginx_wordpress_site_conf: vendor/roles/TypistTech.trellis-cloudflare-origin-ca/templates/wordpress-site.conf.child
```

But it the only way I could think of.

Adding to `includes.d/{{ item.key }}/` directory:
It doesn't get included for redirect domains..... :(

swalkinshaw:
Yeah I did notice that and thought it wasn't ideal either but honestly I'm not so sure how much it matters.

In any language with packages, it would be equivalent to `include vendor/roles/TypistTech.trellis-cloudflare-origin-ca/templates/wordpress-site.conf.child` which most languages would allow `include TypistTech.trellis-cloudflare-origin-ca/templates/wordpress-site.conf.child`. It's still coming from the same place /shrug

TangRufus:
What to do when another galaxy role need to extend `wordpress-site.conf.j2` as well?

swalkinshaw:
Good question. I'm just not sure we can reasonably expect a solution for that.

TangRufus:
For current state of Trellis, would it be better to use regexp replace instead of jinja2 template override?

https://github.com/TypistTech/trellis-cloudflare-origin-ca/pull/8

swalkinshaw:
Up to you ultimately. Usually I wouldn't like regexp for things like this, but at least the template output is predictable since it's not a manually created file.

PR #872, opened 2017-08-30
Update ansible galaxy roles

TangRufus:
`geerlingguy.daemonize` is removed because:
- trellis doesn't use it directly
- `geerlingguy.mailhog` marked it as a dependency

https://github.com/geerlingguy/ansible-role-mailhog/blob/f7c4603eee04803b9cda7c0ba911c2a49ed2d75f/meta/main.yml#L3

### DIFF
--- a/requirements.yml
+++ b/requirements.yml
@@ -1,10 +1,10 @@
 - name: composer
   src: geerlingguy.composer
-  version: 1.5.0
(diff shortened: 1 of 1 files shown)


swalkinshaw:
Looks good, thanks 🎉

PR #876, opened 2017-09-01
Update h5bp

TangRufus:
Note: I can't get [ssl.no-default](https://github.com/h5bp/server-configs-nginx/blob/master/sites-available/ssl.no-default) to work. Thus, I omitted it.

### DIFF
--- a/roles/nginx/templates/nginx.conf.j2
+++ b/roles/nginx/templates/nginx.conf.j2
@@ -1,49 +1,51 @@
 # {{ ansible_managed }}
 
-# nginx Configuration File
-# http://wiki.nginx.org/Configuration
+# Configuration File - Nginx Server Configs
+# http://nginx.org/en/docs/dirindex.html
 
 {% block modules_enabled -%}
 include modules-enabled/*.conf;
 {% endblock %}
 
 {% block user %}
-# Run as a less privileged user for security reasons.
+# Run as a unique, less privileged user for security reasons.
+# Default: nobody nobody
 user {{ nginx_user }};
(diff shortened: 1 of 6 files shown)


roles/nginx/templates/nginx.conf.j2
```diff
@@ -1,57 +1,60 @@
...
+# This is only used when you don't override it on a server{} level
+# Default: logs/error.log error
 error_log  {{ nginx_logs_root }}/error.log warn;
 {% endblock %}
 
 {% block pid %}
 # The file storing the process ID of the main process
-pid        /run/nginx.pid;
+# Default: nginx.pid
+pid        /var/run/nginx.pid;
```
swalkinshaw: Was this change from h5bp? I believe it should stay as `/run`. It's the "newer" location and `/var/run` is symlinked to it.

TangRufus: it's from h5bp:
https://github.com/h5bp/server-configs-nginx/blob/c5c6602232e0976d9e69d69874aa84d2a2698265/nginx.conf#L34

TangRufus: So `/run` is the correct one?

swalkinshaw: From what I can tell it is. So let's keep it as `/run`

TangRufus: reverted

swalkinshaw:
Rest looks good, thanks 👍

PR #879, opened 2017-09-07
Nginx: Add include.d files to redirect domains

TangRufus:


roles/wordpress-setup/templates/wordpress-site.conf.j2
```diff
@@ -216,6 +218,8 @@ server {
   {{ self.https() -}}
 
   {{ self.acme_challenge() -}}
+  
+  {{ self.includes_d() -%}
```
fullyint: Should these each have this change?
```diff
- {{ self.includes_d() -%}
+ {{ self.includes_d() -}}
```

TangRufus: fixed

fullyint:
Thanks! I see only benefits from this addition.

If people find that this addition causes trouble for their use case, they could use a [child template](https://roots.io/trellis/docs/nginx-includes/#child-templates) to remove the `{{ self.includes_d() -}}` from the `redirects_https ` and `redirects_domains` blocks.

PR #880, opened 2017-09-08
Nginx: Include includes.d/http/*.conf

TangRufus:
Allow third party galaxy roles to inject configuration to http block


swalkinshaw:
Makes sense 👍

PR #882, opened 2017-09-08
Add Provisioning Hooks that Allow Custom Tasks 

### LINKED ISSUE #830: add provisioning-hooks directory to operate like deploy-hooks

nbyloff:
Closes #830 

I added this to both `dev.yml` and `server.yml` so the user can keep environments similar. But since these are custom tasks, you can organize them so tasks are executed based on the environment. For example, I only need to backup the database or add performance monitoring tools in production. So adding a `when env == "production" ` statement to the `include` for `backup-amazon.yml` and `performance-monitor.yml` works great. 

These were my first two tests:
Test 1: Created a single YAML file that had custom tasks to be executed. *SUCCESS*
Test 2: Created two files `1-custom-tasks.yml` and `2-custom-tasks.yml` to make sure they execute in that order. *SUCCESS*

Please keep in mind, those aren't the only two options for adding custom provisioning hooks. I might add a single file that references a bunch of 3rd party ansible task libraries, but this simple addition will do wonders for allowing people to customize environments to match their requirements.

It will also make it much easier to merge trellis updates into local development environments without worrying about overwriting developers custom work.

### DIFF
--- a/dev.yml
+++ b/dev.yml
@@ -21,3 +21,13 @@
     - { role: wp-cli, tags: [wp-cli] }
     - { role: wordpress-setup, tags: [wordpress, wordpress-setup] }
     - { role: wordpress-install, tags: [wordpress, wordpress-install] }
+
+- name: Run Custom Provisioning Tasks
+  hosts: web:&development
+  become: yes
+  remote_user: vagrant
+  tasks:
+    - name: Include Custom Provisioning YAML Files
+      include: "{{ item }}"
+      with_fileglob: 
+        - "provision-hooks/*.yml"
--- a/server.yml
+++ b/server.yml
@@ -40,3 +40,12 @@
     - { role: wp-cli, tags: [wp-cli] }
     - { role: letsencrypt, tags: [letsencrypt], when: sites_using_letsencrypt | count }
     - { role: wordpress-setup, tags: [wordpress, wordpress-setup, letsencrypt] }
+
+- name: Run Custom Provisioning Tasks
+  hosts: web:&{{ env }}
+  become: yes
+  tasks:
+    - name: Include Custom Provisioning YAML Files
+      include: "{{ item }}"
+      with_fileglob: 
+        - "provision-hooks/*.yml"


nbyloff:
I can't help but notice this is the only PR with no activity. Is it safe to assume this, or anything similar won't make it?

It seems harmless to me; if you are new or not comfortable with Ansible yet, this update would go unnoticed. But adds a lot of flexibility. We have `deploy-hooks`, why not `provision-hooks`?

fullyint:
@nbyloff Thanks for your time and effort on this. You've provided valuable engagement with Trellis on GitHub and discourse over the longterm.

### TL;DR
* Instead of merging this PR, I suspect users could achieve greater benefit by running 
`ansible-playbook custom.yml -e env=production`
where `custom.yml` uses [`import_playbook: server.yml`](https://docs.ansible.com/ansible/devel/import_playbook_module.html) and any desired custom tasks, includes, etc.
* I think this PR's feature needs to be preceded by a broader discussion of whether and how to enable users to extend Trellis (esp `server.yml`).

----

### Delayed response

>I can't help but notice this is the only PR with no activity. Is it safe to assume this, or anything similar won't make it?

Although this PR is simple in terms of changed LOC, I think this PR is complex in its implications for big picture strategy. I'm afraid complex PRs often sit open longer because of the time investment they will require. That's no excuse for neglecting to offer feedback, of course. It's just part of the challenge of volunteer OSS. We need to do better.

### How could it hurt?

> It seems harmless to me; if you are new or not comfortable with Ansible yet, this update would go unnoticed.

Although this PR would be harmless for basic users, I'm concerned with the affect on users who do use this PR's functionality. If Trellis begins providing means to extend the provisioning of `server.yml`, I think the effort should explore a more general and versatile implementation. If such were implemented, it would likely mean removing this PR's specific code, breaking functionality for users who have already begun relying on it.


### An alternative
It may seem that my hope for [perfect is the enemy of the good](https://en.wikipedia.org/wiki/Perfect_is_the_enemy_of_good), that any continued delay robs users of this PR's better-than-nothing option that is ready to use now. However, I think users have a superior option already available:

>I suspect users could achieve greater benefit by running 
>`ansible-playbook custom.yml -e env=production`
>where `custom.yml` uses [`import_playbook: server.yml`](https://docs.ansible.com/ansible/devel/import_playbook_module.html) and any desired custom tasks, includes, etc. 

I believe this alternative would offer users even more flexibility than trying to work within the specific implementation of this PR. For example:
* It enables users to also add tasks before server.yml.
* Suppose users decide they don't want the "always include everything" approach of `with_fileglob` which could require filenames like `1-name-x` and `2-name-y`. Instead, uses could choose an `include_tasks` approach of looping over a variable's list of include files.
  * This is more explicit on which files will be included and in what sequence. The list order determines include sequence, obviating the need for numeric prefixes to filenames.
  * In addition, disabling an include is a matter or merely omitting a filename from the list (or commenting it out), whereas the `with_fileglob` method requires removing the file altogether (or perhaps renaming the file). 

I think this alternative strategy of a `custom.yml` playbook that uses `import_playbook: server.yml` achieves the [desired benefits you listed](https://github.com/roots/trellis/issues/830#issuecomment-323887371):
>* With the example code, a user can introduce new code in multiple ways. They could (1) have a single YAML file that references other YAML files (or complete roles) in the order they should be executed, (2) number the files like suggested in order of execution.
>* I think this would actually encourage people to understand ansible and trellis more. I would imagine in most scenarios, once you're to the point you want to add custom functionality to provisioning, you have basic understanding of how things are working.
>* Merging into your project from trellis master become a lot less painful (this cannot be understated how nice it will be)

### Trellis: Positioning and scope

The above is part of why "I think this PR is complex in its implications for big picture strategy." Another part is the discussion of whether this PR would be a step in the direction of Trellis becoming more of a framework or at least able to provision and deploy a wider range of web apps (maybe related: #914). Such could be a shift in emphasis and possibly a dramatic change of scope. I'm not saying it shouldn't be considered or discussed. Rather, I'm saying that this PR may look simple on its face but it may actually be a huge deal, with commensurate time investment.

### Hook inconsistencies?

K cool, but what about this...

>We have `deploy-hooks`, why not `provision-hooks`?

I'm sympathetic to the implied argument of this question. I don't totally disagree but I'll try to play devil's advocate and offer a rebuttal.

In the context of "what is it?" one could say, "`server.yml` and `deploy.yml` are both playbooks with roles, right? So, the presence or lack of hooks seems inconsistent." This is random, but what if we were talking about a private home's entryway door vs. a building's entryway revolving door. They are both doors and the presence or lack of a door knob may seem inconsistent. The different implementation of the door or playbook only makes sense when considering the broader context and purpose. Ok, now forget about doors (my hasty and inapplicable dilution to discussion).

The WordPress sites that users may deploy could vary quite a bit, driving a need for deploy hooks to facilitate customization. In contrast the "WordPress app server" itself (the stuff handled by `server.yml`) may differ much less between users. Most app server differences are configurable via existing play and role variables. Of course, app server differences would far more vast if Trellis were to position itself as a framework.

The belabored point is that although some users will desire to customize `server.yml`, the desire will probably be much less common than with `deploy.yml`. In addition, provision and deploy are conceptually different, which is part of the reason for the separation into distinct playbooks. I think these factors weaken the implied argument of "We have `deploy-hooks`, why not `provision-hooks`?"

### Should Trellis run your custom stuff, or should your custom stuff run Trellis?

Users will have custom stuff to run, as in [your example](https://github.com/roots/trellis/issues/830#issue-222051534):
>I always add a custom backup job that zips and pushes the DB and uploads directory to an Amazon s3 bucket nightly.

One angle on the issue of running custom stuff is to ask the question:
**🔴  Should Trellis run users' custom stuff or
🔵  should users' custom stuff run Trellis?**

As a concrete example,
🔴  should Trellis provide hooks in `server.yml` or
🔵  should users create a `custom.yml` playbook that uses `import_playbook: server.yml`. 
Alternatively, a more decoupled custom playbook could run only select Trellis roles.

I acknowledge that 
🔴  for a certain chunk of the task of deploying, Trellis runs your custom stuff (i.e., in Trellis deploy-hooks).
🔵  However, maybe your custom stuff (like CI tools) actually run Trellis deploys.

As examples for `server.yml` and provisioning,
🔴  Trellis could expand to initiating your infrastructure (e.g., via DigitalOcean API, terraform, etc.) and creating your S3 backups.
🔵  Or maybe you build out other tools to execute those tasks, and one step in your tool is to run Trellis `server.yml`. 

I'm rambling and am going to trail off without a strong opinion or recommendation on this particular issue. I think it's worth considering building more of this functionality into Trellis, or at least aggregating guides for extending Trellis and for setting up workflows that run Trellis at certain points.

Again, thank you!
Clarification, rebuttals and feedback welcome.

nbyloff:
I really appreciate your thorough response. I have been using separate custom.yml files like you suggest, and most likely the most reasonable solution currently. I have been using the initial method for a while now only because it forces me to not forget specific files or setups if I need them in the future.

But that most likely be solved with better documentation on individual projects. Again, thank you for the response.

swalkinshaw:
Closing this after @fullyint's response which does a great job of explaining our stance.

@nbyloff thanks for taking the time to do this and creating some discussion. Should always be noted that when we decline a PR like this, it's not because its "wrong". It's just not the solution we want at this time.

I think we all agree we can provide some better guidance on how to do this 👍

PR #887, opened 2017-09-13
Add `update_cache: true` to `apt` tasks

TangRufus:
Because `apt_repository` tasks skip `apt-get update` if repo is already added.

This patch changes nothing if `state: present` but necessary for `state: latest`

Question: Do we actually need `force: yes` in `apt` tasks?


TangRufus:
Duplicates #881

PR #890, opened 2017-09-16
Extract Trellis::Config

### LINKED ISSUE #886: Feature Request: Turn lib/trellis/vagrant.rb into a class
Turn [lib/trellis/vagrant.rb](https://github.com/roots/trellis/blob/master/lib/trellis/vagrant.rb) into a class so that other vagrant plugin can read site info from `load_wordpress_sites`.

Currently, we can't use `lib/trellis/vagrant.rb` in vagrant plugins because defining `ANSIBLE_PATH` dynamically gives errors/warnings.

Note: Constant `ANSIBLE_PATH` should be removed

Example usage:
```ruby
trellis = Trellis.new(ansible_path: '/path/to/trellis/root')

# Load wordpress_sites.yml and return it as a hash
trellis.wordpress_sites

```

TangRufus:
Helps vagrant plugins to read site info from `group_vars/development/wordpress_sites.yml`.

closes #886

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -19,7 +19,7 @@ ensure_plugins(vconfig.fetch('vagrant_plugins')) if vconfig.fetch('vagrant_insta
 
 trellis_config = Trellis::Config.new(root_path: ANSIBLE_PATH)
 
-Vagrant.require_version '>= 1.8.5'
+Vagrant.require_version '>= 1.9.6'
 
 Vagrant.configure('2') do |config|
   config.vm.box = vconfig.fetch('vagrant_box')
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -19,7 +19,7 @@ ensure_plugins(vconfig.fetch('vagrant_plugins')) if vconfig.fetch('vagrant_insta
 
 trellis_config = Trellis::Config.new(root_path: ANSIBLE_PATH)
 
-Vagrant.require_version '>= 1.9.6'
+Vagrant.require_version '>= 1.8.5'
 
 Vagrant.configure('2') do |config|
   config.vm.box = vconfig.fetch('vagrant_box')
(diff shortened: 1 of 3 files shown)


Vagrantfile
```diff
@@ -16,10 +17,9 @@ end
 
 ensure_plugins(vconfig.fetch('vagrant_plugins')) if vconfig.fetch('vagrant_install_plugins')
 
-wordpress_sites = load_wordpress_sites
-site_hosts = hosts(wordpress_sites)
+trellis_config = Trellis::Config.new(root_path: ANSIBLE_PATH)
 
-Vagrant.require_version '>= 1.8.5'
+Vagrant.require_version '>= 1.9.6'
```
swalkinshaw: Any reason for this?

lib/trellis/config.rb
```diff
@@ -0,0 +1,75 @@
...
+
+    def multisite_subdomains?
+      @using_multisite_subdomains ||= begin
+        wordpress_sites.any? do |(_name, site)|
+          site.dig('multisite', 'enabled') && site.dig('multisite', 'subdomains')
+        end
+      end
+    end
+
+    def canonicals
```
swalkinshaw: I'd prefer explicit names for these: `site_hosts_canonical`

lib/trellis/config.rb
```diff
@@ -0,0 +1,75 @@
...
+      @canonicals ||= site_hosts.map { |host| host['canonical'] }
+    end
+
+    def redirects
```
swalkinshaw: `site_hosts_redirects`

Vagrantfile
TangRufus: Because of `dig`. It needs Ruby 2.3

TangRufus:
Renamed

Vagrantfile
swalkinshaw: `dig` is awesome but only used in one method. Can we do without it for now?

I'd rather bump this separately. It does involve a little discussion since triggers and landrush plugins won't work :( (annoying since they are dead plugins).

TangRufus: reverted

swalkinshaw:
Thanks! This is a good improvement.

PR #891, opened 2017-09-16
Ansible: Retry ssh connection one more time

TangRufus:
Fix issues when using with non-standard ssh setups, for example: using with [Kryptonite](https://krypt.co/)


swalkinshaw:
why not 👍

PR #894, opened 2017-09-20
Default h5bp expires and cache busting to false

swalkinshaw:
Expires needs to be disabled by default and cache busting should be optional as well.


partounian:
Why was this done?

Asking for genuine curiosity, not_dev seems like a nice setup as everyone should have browser cache unless you have an edge case situation.

swalkinshaw:
See https://github.com/roots/trellis/pull/37 for older discussion on this.

PR #895, opened 2017-09-20
Ansible 2.4 compatibility

fullyint:
- update `include` to `import_tasks`, `include_tasks` or `import_playbook` ([docs](https://docs.ansible.com/ansible/devel/playbooks_reuse.html) and [docs](https://docs.ansible.com/ansible/devel/playbooks_reuse_includes.html)).
  This avoids deprecation warning. However, `server.yml` will still show deprecation warning till the composer galaxy role changes these [two `includes`](https://github.com/geerlingguy/ansible-role-composer/blob/1e9c94423584b6eb0e5b4f4cba5d730beb24d822/tasks/main.yml#L54-L58).
- bump Ansible requirement to 2.4 to accommodate `import_*` and `include_*`
- add warning for Ansible versions greater than tested with Trellis
  It happens that people update Ansible but not Trellis, then run into problems, not realizing they need to update their Trellis for compatibility, posting on discourse, etc.
  ```
  [WARNING]: You Ansible version is 2.5.0.0 but this version of Trellis has only been tested for
  compatability with Ansible 2.4.0.0 -> 2.4.2.0. It is advisable to check for Trellis updates or
  downgrade your Ansible version.
  ```
- adjust `when` conditions: `failed` and `success` [no longer account for `rc`](https://github.com/ansible/ansible/commit/2a041d10d26d5746132061da4b874f33cb73ffb0#diff-bc6bf5732a9c00dcb3cb182f4ce3b7c8R29)
- move dynamic HostKeyAlgorithms from `--ssh-extra-args` to `environment` keyword
  `--ssh-extra-args` seemed to no longer have any effect ([maybe related](https://github.com/ansible/ansible/commit/f921369445900dcc756821e40c9257d5359087af#diff-efa3a13dd7cbec87c75b359049e1c7d3L295)?)
- accommodate potential for multiple inventories (`inventory` list in `cli.options`)
  > It is now possible to specify mulitple inventory sources in the command line (-i /etc/hosts1 -i /opt/hosts2) _[--Ansible CHANGELOG](https://github.com/ansible/ansible/blob/devel/CHANGELOG.md)_
- remove `loader` kwarg from VariableManager.get_vars()

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,13 @@
 ### HEAD
+* Default h5bp expires and cache busting to false ([#894](https://github.com/roots/trellis/pull/894))
+* Deploys: Update WP theme paths for multisite subsites ([#854](https://github.com/roots/trellis/pull/854))
+* Vagrant: Support DHCP ([#892](https://github.com/roots/trellis/pull/892))
+* Extract Trellis::Config ([#890](https://github.com/roots/trellis/pull/890))
+* Redirect directly to https canonical domain ([#889](https://github.com/roots/trellis/pull/889))
+* WordPress Setup: Add Nginx `ssl_client_certificate` ([#869](https://github.com/roots/trellis/pull/869))
+* Update h5bp/server-configs-nginx ([#876](https://github.com/roots/trellis/pull/876))
+* Update ansible galaxy roles ([#872](https://github.com/roots/trellis/pull/872))
+* Update wp-cli to 1.3.0 ([#871](https://github.com/roots/trellis/pull/871))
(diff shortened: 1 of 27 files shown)


franzliedke:
Great work. :+1:

A silent upgrade to Ansible 2.4 means we can not deploy our site anymore. Took us a while to find out that the Ansible upgrade was the problem.

franzliedke:
Great, thanks! Can we get a new release as well? :innocent:

ztackett11:
How can I go about adding these changes to my trellis repo? i think i was able to update trellis to rc 2 (I think)

Issue #893, opened 2017-09-19
Feature Request: Normalize SSL provider tasks


TangRufus:
**What is the current behavior?**

 - `letsencrypt` has its own [role](https://github.com/roots/trellis/tree/master/roles/letsencrypt) whereas `self-signed-certificate` is embedded [inside](https://github.com/roots/trellis/blob/master/roles/wordpress-setup/tasks/self-signed-certificate.yml) `wordpress-setup`.

 - Although `letsencrypt` task is separated, its variables (for example: `letsencrypt_cert_ids`) are *leaded* to other tasks like `wordpress-setup` because they are used in Nginx templates.

**What is the expected or desired behavior?**

Normalize SSL provider tasks:

 - Separate `self-signed-certificate` task

 - SSL provider tasks (`letsencrypt` & `self-signed-certificate`) template their own Nginx config to `{{ nginx_path }}/includes.d/{{ item.key }}/ssl/xxx.conf` (Note: `ssl` subdirectory!)

 - Remove hardcoded Nginx ssl config, for example: 
 https://github.com/roots/trellis/blob/12ac783d24bc62b9426e88cc2c34d2294e1d078a/roles/wordpress-setup/templates/wordpress-site.conf.j2#L88-L103

 - Move SSL provider tasks after `wordpress-setup` to prevent [wordpress-setup/nginx-includes](https://github.com/roots/trellis/blob/ec9324ccbc1910115d382d0559784824084d683f/roles/wordpress-setup/tasks/nginx-includes.yml#L44) from deleting ssl nginx config.

**Please provide use cases for changing the current behavior:**

 - Serve as a reference for [third party SSL providers](https://github.com/typisttech/trellis-cloudflare-origin-ca)

 - Less trouble when working with `tags`, see: https://github.com/roots/trellis/pull/888#discussion_r139036684

 - Other tasks and custom Nginx template need less knowledge about SSL config. Whenever needed, include `{{ nginx_path }}/includes.d/{{ item.key }}/ssl/*.conf` is enough, see: https://github.com/roots/trellis/pull/888#discussion_r139283502

swalkinshaw:
I like the sound of all of this 👍  seems like a nice improvement with the separate roles

But I reserve the right to see it first 😛

roots-ladybug:
Hi @TangRufus,

[...]

PR #896, opened 2017-09-21
Normalize SSL providers

TangRufus:
# Normalize SSL providers `manual` and `self-signed`

 - Putting Nginx SSL configs in `{{ nginx_path }}/includes.d/{{ item.key }}/ssl`
 
 - Putting certificates in `{{ nginx_ssl_path }}/manual-certificates` or `{{ nginx_ssl_path }}/self-signed-certificates`

# Add `Certbot`

This is actually a failed attempt to replace `letsencrypt` with `Certbot`. I can't get [multiple-servers](https://roots.io/trellis/docs/ssl/#multiple-servers) implemented using `Certbot`. Help wanted!

See: #893


TangRufus:
Besides, Certbot doesn't:

- detect new hostnames like #630
- not support of `--staging` yet

swalkinshaw:
There's also two built-in Ansible modules which we could look into:

* http://docs.ansible.com/ansible/latest/openssl_certificate_module.html
* http://docs.ansible.com/ansible/latest/letsencrypt_module.html

TangRufus:
The openssl_certificate_module requires ansiable 2.4. Is #895 going to be merged soon?

The letsencrypt_module requires us to change Nginx back and forth for acme challenge. Certbot handles it for us. I think Certbot simplifies everything.

swalkinshaw:
If `openssl_certificate` actually helps/is simpler then we have no problem requiring 2.4 asap.

Our current LE implementation is a little convoluted so I'm open to any solution which simplifies it. Although I do think we have some more unique requirements with it.

TangRufus:
`openssl_certificate` looks nice. Updated `self-signed-certificate` to use it.

Removed `Certbot` because:
 - it doesn't play well with pyOpenSSL which required for`openssl_certificate` , see: https://github.com/certbot/certbot/issues/5111
 - it can't be used with `letsencrypt` role on the same server

Rebased upon d96a58f. Actual changes here: https://github.com/roots/trellis/compare/ansible-2.4...TangRufus:ssl-providers

swalkinshaw:
@TangRufus can this get a rebase?

TangRufus:
Rebased. Note that I modified `server.yml` and `dev.yml` because ssl providers need to be tagged. Otherwise, `wordpress-setup` deletes their nginx config files.

swalkinshaw:
Tested and working great 👍 

Only thing I ran into due to our current tags is if you want to enable SSL after, and *only* use a `wordpress` or `wordpress-setup` tag, Nginx will fail because `dhparams` weren't initially generated.

So if we want a common tag which would run all Nginx related tasks/roles, we might want to introduce a new one, or just apply `wordpress-setup` to `nginx` as well?

TangRufus:
I vote for applying `wordpress-setup` to `nginx` because this also helps when adding or removing sites, i.e.: `Create Nginx available sites` & `Enable or disable Nginx sites` tasks

PR #892, opened 2017-09-16
Vagrant: Support DHCP

TangRufus:
Allow running multiple Trellis machines at the same time.

Tested with 2 DHCP machines and 1 static IP machine running at the same time. Results:
```bash
$ cat /etc/exports
# VAGRANT-BEGIN: 501 78f3bfb6-4ae0-4305-b4a3-8cd68040aa6b
"/Users/me/Code/siteA/bedrock" 172.28.128.3 -alldirs -mapall=501:20
"/Users/me/Code/siteA/trellis" 172.28.128.3 -alldirs -mapall=501:20
# VAGRANT-END: 501 78f3bfb6-4ae0-4305-b4a3-8cd68040aa6b
# VAGRANT-BEGIN: 501 2bf7c671-523e-4ca5-9a81-e85797428609
"/Users/me/Code/siteB/bedrock" 172.28.128.4 -alldirs -mapall=501:20
"/Users/me/Code/siteB/trellis" 172.28.128.4 -alldirs -mapall=501:20
# VAGRANT-END: 501 2bf7c671-523e-4ca5-9a81-e85797428609
# VAGRANT-BEGIN: 501 8dafbb55-1263-4829-90e2-b76fb68323c0
"/Users/me/Code/siteC/site" 192.168.50.5 -alldirs -mapall=501:20
"/Users/me/Code/siteC/trellis" 192.168.50.5 -alldirs -mapall=501:20
# VAGRANT-END: 501 8dafbb55-1263-4829-90e2-b76fb68323c0

$ cat /etc/hosts
...
## vagrant-hostmanager-start id: 95b74937-d48d-468e-850a-ce2664ad1ea7
172.28.128.3	www.typist.dev
172.28.128.3	typist.dev
## vagrant-hostmanager-end

## vagrant-hostmanager-start id: 3fa4d5c6-050d-4545-bc5c-e4db40ec516c
172.28.128.4	ww2.typist.dev
172.28.128.4	ww3.typist.dev
## vagrant-hostmanager-end

## vagrant-hostmanager-start id: 5695f53b-c08e-46be-9d1f-dd08d9cc2842
192.168.50.5	example.dev
192.168.50.5	www.example.dev
## vagrant-hostmanager-end
```

The `ip_resolver` proc is taken from https://github.com/devopsgroup-io/vagrant-hostmanager/issues/86#issuecomment-176572400
Related https://github.com/roots/trellis/issues/295#issuecomment-129458413
Closes #873

PR #898, opened 2017-09-24
Vagrant: Use DHCP by default

TangRufus:
Follow up on #892

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Vagrant: Use DHCP by default ([#898](https://github.com/roots/trellis/pull/898))
 * [BREAKING] Normalize `apt` tasks ([#881](https://github.com/roots/trellis/pull/881))
 * Ansible 2.4 compatibility ([#895](https://github.com/roots/trellis/pull/895))
 * Default h5bp expires and cache busting to false ([#894](https://github.com/roots/trellis/pull/894))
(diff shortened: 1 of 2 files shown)


swalkinshaw:
If someone has a VM running and merges this in, do we know what the behaviour is?

TangRufus:
It should continue working until they stop the VM.

Best case scenario:
Everything works. Normal `vagrant halt && vagrant up` / `vagrant reload` will pick up the new dynamic ip.

Not so good scenario:
They need to run `vagrant destory` if any conflict happens. 

Worst case scenario:
If are unlucky as me, having some buggy vagrant plugins and buggy custom scripts:
1. `vagrant halt` **all** other VMs
2. power off and destroy the problematic VM via VirtualBox GUI
3. `sudo rm /etc/exports`
4. remove related hosts from `/etc/hosts`
5. `vagrant up`

swalkinshaw:
Okay thats about what I expected and I think it's fine 👍 

Of course they can always keep the static IP instead.

swalkinshaw:
@TangRufus did you mean to close this?

TangRufus:
No, mis-clicked. Sorry.

swalkinshaw:
Want to add a changelog entry? Will merge after that.

TangRufus:
done.

TangRufus:
Sorry guys, my instruction about `vagrant destroy` is incorrect. I have a buggy script "cached" the IP and changed DHCP's behaviour on initial `vagrant up`, making it works like static IP.

If `vagrant halt && vagrant up` or `vagrant reload` don't work for you, do not destroy the VM:
1. power off the problematic VM via VirtualBox GUI
2. (maybe) sudo rm /etc/exports
3. (maybe) remove related hosts from /etc/hosts
4. vagrant up

PR #901, opened 2017-10-04
Enable ioapic

### LINKED ISSUE #900: Performance can be improved dramatically with ioapic

- [x] This is a feature request

---

No enabling `ioapic` by default.

Enabling it by default.

jdevalk:
Fixes #900

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -140,6 +140,7 @@ Vagrant.configure('2') do |config|
     vb.name = config.vm.hostname
     vb.customize ['modifyvm', :id, '--cpus', vconfig.fetch('vagrant_cpus')]
     vb.customize ['modifyvm', :id, '--memory', vconfig.fetch('vagrant_memory')]
+    vb.customize ['modifyvm', :id, '--ioapic', 'on']
 
     # Fix for slow external network connections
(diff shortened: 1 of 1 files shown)


jdevalk:
Of course this should be tested for breakage, we've tested this on a couple machines but they're all 64 bit macs :)

swalkinshaw:
Never heard of this setting before 🤔 

VB docs specially say:

> On any host, you should enable the I/O APIC for virtual machines that you intend to use in 64-bit mode. This is especially true for 64-bit Windows VMs. See Section 3.3.2, “"Advanced" tab”. In addition, for 64-bit Windows guests, you should make sure that the VM uses the Intel networking device, since there is no 64-bit driver support for the AMD PCNet card; see Section 6.1, “Virtual networking hardware”.

Considering we only support Ubuntu 64bit guests... seems like a good default. Would be nice for a few other people to test this as well.

swalkinshaw:
Gave this a quick test and everything seemed fine. Would still like some other people to test. Especially on Windows. @kalenjohnson @QWp6t hint hint

swalkinshaw:
Thanks for this PR. Replacing it with https://github.com/roots/trellis/pull/901 which adds a few more config options too.

PR #904, opened 2017-10-05
Add option vagrant_prefer_nfs

TangRufus:
On Macs that using APFS, `vagrant up` gives:
```
==> default: Mounting NFS shared folders...
The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!
mount -o vers=3,udp 192.168.50.1:/Users/XXX/Code/site /vagrant-nfs-example.com
Stdout from the command:
Stderr from the command:
mount.nfs: access denied by server while mounting 192.168.50.1:/Users/XXX/Code/site
```
See:
 - https://github.com/hashicorp/vagrant/issues/8788
 - https://github.com/geerlingguy/drupal-vm/issues/1547

This pull request adds `vagrant_prefer_nfs` option so that non-windows users could have a choice to use default `synced_folder` type.

Only tested with vagrant 1.9.8 + macOS 10.13 + APFS.

For APFS user:
1. Set `vagrant_prefer_nfs: false`
2. `FORCE_ANSIBLE_LOCAL=true vagrant up`

Note: This won't work on vagrant 2.0, see: https://discourse.roots.io/t/the-requested-ansible-version-2-4-0-was-not-found-on-the-guest/10513

Discuss:
 - Should we implicitly set `ansible_local` when `vagrant_prefer_nfs: false`(like what we are doing for windows users)?
 - Bring back #439?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Add option vagrant_prefer_nfs ([#904](https://github.com/roots/trellis/pull/904))
 * [BREAKING] Normalize `apt` tasks ([#881](https://github.com/roots/trellis/pull/881))
 * Ansible 2.4 compatibility ([#895](https://github.com/roots/trellis/pull/895))
 * Default h5bp expires and cache busting to false ([#894](https://github.com/roots/trellis/pull/894))
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -69,7 +69,7 @@ Vagrant.configure('2') do |config|
 
   bin_path = File.join(ANSIBLE_PATH_ON_VM, 'bin')
 
-  if Vagrant::Util::Platform.windows? and !Vagrant.has_plugin? 'vagrant-winnfsd'
+  if (Vagrant::Util::Platform.windows? && !Vagrant.has_plugin?('vagrant-winnfsd')) || !vconfig.fetch('vagrant_prefer_nfs')
     trellis_config.wordpress_sites.each_pair do |name, site|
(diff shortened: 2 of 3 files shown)


swalkinshaw:
re: #439, we did enable ansible-local by default 😄 

I'm fine with adding this option regardless. Is there no better solution for APFS yet?

TangRufus:
> we did enable ansible-local by default

Is that true for non-wondows users?
https://github.com/roots/trellis/blob/master/lib/trellis/vagrant.rb#L40-L42

> Is there no better solution for APFS yet?

I don't find one.

Please test this pull request on non-APFS Macs and other platforms.

swalkinshaw:
> Is that true for non-windows users?

Yep 👍

TangRufus:
```ruby
# https://github.com/roots/trellis/blob/master/lib/trellis/vagrant.rb#L40-L42

def local_provisioning?
  @local_provisioning ||= Vagrant::Util::Platform.windows? || !which('ansible-playbook') || ENV['FORCE_ANSIBLE_LOCAL']
end
```

> To use Trellis for remote servers, we recommend installing Ansible locally on your host machine
> https://roots.io/trellis/docs/remote-server-setup/

If Ansible is installed as the doc suggests, `ansiable_local` is not used.

swalkinshaw:
Yeah, sorry I just meant it's the *default* for everyone. If someone follows our docs as of now, they'd get ansible local.

swalkinshaw:
Is this PR useful as is?

Or does it depend on:

> Should we implicitly set ansible_local when vagrant_prefer_nfs: false(like what we are doing for windows users)?

TangRufus:
After moving to a SSD harddisk, trellis works out of the box. I think this pull request should not be merged and APFS on traditional harddisk users should manually disable NFS and use ansible_local.

PR #907, opened 2017-10-17
Clarify DB creation task name

greatislander:
I've always found the description of the database creation task confusing (it's not creating a database _of_ sites). This PR changes it.


swalkinshaw:
Makes sense :)

PR #915, opened 2017-11-06
Fix raw_vars functionality for Ansible 2.4.1

fullyint:
This PR fixes broken `raw_vars` functionality for Ansible 2.4.1.

**Background.** In contrast to prior versions, Ansible 2.4.1 [populates `basedirs` only if a `task` is passed](https://github.com/ansible/ansible/commit/7f2c611dfceaef877de7dded084362d28bf5f8c7#diff-473ef6db3f086739e3c053f001d731b5R236) to the `VariableManager`'s `get_vars()`. Without a known `basedir`, no `group_vars` files are loaded, such that `get_vars()` no longer returns any group vars, only host vars and play vars. The `basedirs` definitions and group vars retrieval should go back to normal in Ansible 2.4.2 (fix in ansible/ansible#32269).

**Implications for Trellis.** The Trellis `raw_vars` feature retrieves and wraps var values in `{% raw %}` using the `v2_playbook_on_play_start` callback, not during any particular task. Thus Trellis did not pass any `task` to `get_vars()`. With Ansible 2.4.1, this means the `raw_vars` feature couldn't find nor process any group vars, risking jinja templating errors (e.g., for password vars with random characters).

**Implementation.** This PR passes a dummy `Task()` to `get_vars()` just so `basedirs` will be populated and group vars retrieved. The is no other effect of passing in the `task`: `task._role` is `None`, `task.environment` is `None`, and `task.get_vars()` returns `{}`.

**Future.** It should be ok to remove the dummy `task` once the Trellis minimum requirement for Ansible is 2.4.2+. I decided not to bother making the dummy task conditional on Ansible version 2.4.1 given that it had no problematic effects in 2.4.0 nor in 2.4.2.0-0.2.beta2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Fix `raw_vars` functionality for Ansible 2.4.1 ([#915](https://github.com/roots/trellis/pull/915))
 * Enable Virtualbox ioapic option ([#913](https://github.com/roots/trellis/pull/913))
 * Dynamically increase `ansible_group_priority` for selected env ([#909](https://github.com/roots/trellis/pull/909))
 * Bump Ansible `version_tested_max` to 2.4.1.0 ([#911](https://github.com/roots/trellis/pull/911))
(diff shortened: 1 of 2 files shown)


swalkinshaw:
👍

PR #918, opened 2017-11-13
Update to WP-CLI 1.4.1

greatislander:
Fixes SQL injection bug in `wp search-replace`, among other things.

See:

+ https://make.wordpress.org/cli/2017/11/13/version-1-4-1-released/
+ https://blog.ircmaxell.com/2017/10/disclosure-wordpress-wpdb-sql-injection-technical.html


retlehs:
thanks, and especially thanks for the changelog update :D

Issue #920, opened 2017-11-15
Let's encrypt error due to new agreement url 


vdrnn:
## Bug report

When I try to reprovision the server, I get an error on the **[letsencrypt : Generate the certificates]** task. Looks like the agreement url changed today, on the 15th of November.

I already had a certificate with this subdomain. I rebuilt the droplet on DO. 

What also feels wrong is that the old URL https://letsencrypt.org/documents/LE-SA-v1.1.1-August-1-2016.pdf works whereas the new one doesn't https://letsencrypt.org/documents/LE-SA-v1.2-November-15-2017.pdf ?

**Please provide steps to reproduce, including full log output:**

`non-zero return code
[...]

PR #921, opened 2017-11-15
Update acme-tiny commit hash

MWDelaney:
In response to #920. According to [this commit](https://github.com/diafygi/acme-tiny/commit/c4940d229a296b7643e86d2e9ab31a6e0099ba71) by @diafygi, this should also prevent this problem going forward by automatically grabbing the correct agreement URL, and according to [this post](https://community.letsencrypt.org/t/subscriber-agreement-update-november-15-2017/45607) over at LetsEncrypt, it is not necessary to re-sign the agreement every time it changes, so this problem should be resolved for good now?

### DIFF
--- a/roles/letsencrypt/defaults/main.yml
+++ b/roles/letsencrypt/defaults/main.yml
@@ -4,7 +4,7 @@ missing_hosts: "{{ site_hosts | difference((current_hosts.results | selectattr('
 letsencrypt_cert_ids: "{ {% for item in (generate_cert_ids | default({'results':[{'skipped':True}]})).results if not item | skipped %}'{{ item.item.key }}':'{{ item.stdout }}', {% endfor %} }"
 
 acme_tiny_repo: 'https://github.com/diafygi/acme-tiny.git'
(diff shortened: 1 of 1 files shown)


swalkinshaw:
Diff looks minimal which is good: https://github.com/diafygi/acme-tiny/compare/5a7b4e79bc9bd5b51739c0d8aaf644f62cc440e6...4ed13950c0a9cf61f1ca81ff1874cde1cf48ab32

Thanks 👍

tim-mortimer:
Thank you for the fix! This definitely had me scratching my head for a while.

Should we reprovision our servers as a result of this fix, or will the renewals still work correctly?

swalkinshaw:
You need to reprovision. You can just run the `letsencrypt` tags to make it fast.

BlueBamboo1221:
hello,sir!
I am tried as this method
but same error
can you help ?

MWDelaney:
No method is described in this thread. This is a pull request which was merged into Trellis a year ago.

Likewise no error is described in this thread so I don’t know what you mean by “same error”. What error are you receiving?

If you need technical support, please post on [our forum](https://discourse.roots.io/).

PR #929, opened 2017-12-05
Update to PHP 7.2

retlehs:
note: xdebug currently not avail, see https://github.com/oerdnj/deb.sury.org/issues/751

the `mcrypt` extension is also removed (see http://php.net/manual/en/migration71.deprecated.php)

Note: this PR will install all necessary PHP 7.2 packages and *stop* the existing 7.1 service if it exists. PHP 7.1 packages will still remain installed but unused. They can be manually uninstalled if need be. We aren't automatically cleaning up the 7.1 packages because people might have custom packages installed. This also allows for easier rollback if need be.


retlehs:
> xdebug 2.6.0 for PHP 7.x and xdebug 2.5.x for PHP 5.x has been published in the PPA.

https://github.com/oerdnj/deb.sury.org/issues/751#issuecomment-363094638

PR #935, opened 2018-01-11
Allow Trellis VMs to be created in Windows Subsystem for Linux

alwaysblank:
WSL does not support NFS, but Trellis's checks to determine whether or not to
use NFS for Vagrant shares assume that NFS is supported if the operating system
is not reported to be Windows. Unfortunately this means that it attempts to
use NFS in WSL and consequently fails..

This change simply adds an additional check: If Vagrant believes that it is
being run from WSL, it syncs folders without NFS (just as though it was
running in Windows without `vagrant-winnfsd`). Other than that, it should
have no effect on any other functionality.

Note: This applies to Vagrant when it is run as a Linux executable through
WSL, *not* when the Windows `vagrant.exe` is run from WSL.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -69,7 +69,7 @@ Vagrant.configure('2') do |config|
 
   bin_path = File.join(ANSIBLE_PATH_ON_VM, 'bin')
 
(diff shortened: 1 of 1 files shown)


swalkinshaw:
I'll have to take your word for it :)

Thanks

PR #936, opened 2018-01-12
Support for IPv6 in Letsencrypt challenges

gmaiztegi:
When trying to validate the hostnames prior to the certificate generation process, it will fail if the IPv6 address is set for those hostnames and the server uses the protocol. This should fix this behavior.

### DIFF
--- a/roles/letsencrypt/templates/nginx-challenge-site.conf.j2
+++ b/roles/letsencrypt/templates/nginx-challenge-site.conf.j2
@@ -1,4 +1,5 @@
 server {
+  listen [::]:80;
   listen 80;
   server_name {{ missing_hosts | join(' ') }};
   include acme-challenge-location.conf;


fullyint:
Looks good to me.

------

IPv6-related notes from  #885:

>### Testing IPv6
>Of course, to test IPv6, one's local machine and server must both enable IPv6. For Trellis to make IPv6 SSH connections (e.g., to run a playbook), adjust the [ListenAdresses](https://github.com/roots/trellis/blob/ec9324ccbc1910115d382d0559784824084d683f/roles/sshd/defaults/main.yml#L9-L10) and [AddressFamily](https://github.com/roots/trellis/blob/ec9324ccbc1910115d382d0559784824084d683f/roles/sshd/defaults/main.yml#L16):
```diff
 sshd_listen_addresses:
   - 0.0.0.0
+  - '::'

-sshd_address_family: inet
+sshd_address_family: any
```
>⚠️ **Trellis is not ready for IPv6 in production**. There are no `ip6tables` settings.

swalkinshaw:
Thanks!

PR #937, opened 2018-01-19
Update Vagrant requirement to v2.0.1.

alwaysblank:
This is necessary to support the changes introduced in commit 0015b95, specifically the use of the `Vagrant::Util::Platform.wsl?` test in the Vagrantfile.

Addressed on Discourse [here](https://discourse.roots.io/t/undefined-method-wsl/11351/4).


swalkinshaw:
Thanks!

PR #938, opened 2018-01-19
Update `vagrant_box_version` to `<= 201801.02.0`

TangRufus:
See: https://app.vagrantup.com/bento/boxes/ubuntu-16.04

---

Question:
The box is `ubuntu-16.04`. Should we use lower-bound version constraint instead of upper-bound?
I.e.: Will this be better? 

```
vagrant_box_version: '>= 201801.02.0'
```


swalkinshaw:
`>=` makes it easier for people to keep their box up to date and *should* be fine, but I guess there's a small chance some update could break Trellis. I would assume the likelyhood of that happening would be very low though?

swalkinshaw:
Going to merge this as is for now and we can see about flipping it after.

PR #940, opened 2018-01-19
Use requests lib

strarsis:
Use requests lib which simplifies/improves code and correctly handles IDNs.

Related to discussion:
https://discourse.roots.io/t/domains-with-special-characters-punycode-umlaut/11376/4


strarsis:
Maybe one should just use the already punycode-escaped IDNs in wordpress_sites config?

swalkinshaw:
I'd avoid pip and other dependencies. Did the punycode-escaping work for you? I know you had another error in your discourse thread but not sure if that was unrelated or not.

strarsis:
I escaped the domain myself and then use it that way (escaped) in config file. This works fine.

swalkinshaw:
I'd prefer to fix by encoding it properly if we can (something like I suggested in the thread), otherwise punycode-escaping seems correct since technically it's not a valid URI as is anyway.

PR #944, opened 2018-01-31
Update wp-cli to 1.5.0

greatislander:
See: https://make.wordpress.org/cli/2018/01/31/version-1-5-0-released/


fullyint:
Works for me. Thanks!

PR #955, opened 2018-03-03
Disable memcached UDP support by default

swalkinshaw:
memcached can be exploited via UDP. This disables it by default.

See https://blog.cloudflare.com/memcrashed-major-amplification-attacks-from-port-11211/
form more details.


fullyint:
lgtm

Optionally could make it a default var `memcached_port_udp: 0` but maybe it'd never be used.

swalkinshaw:
Oh I meant to do that. Added.

Also added changelog entry. This is technically a breaking change although I doubt anyone is actually using the UDP feature.

PR #956, opened 2018-03-06
Add support for sSMTP revaliases configuration

valentinocossar:
Hi,

I'm experiencing some issues with Mailgun and Gmail, some emails sent by a server configured with Mailgun as SMTP to a Gmail account have a non-compliant domain name as `From` header.

This is the Mailgun log.
```json
{
    "severity": "permanent",
    "tags": [],
    "storage": {
        "url": "https://sw.api.mailgun.net/v3/domains/example.com/messages/key_value",
        "key": "key_value"
    },
    "delivery-status": {
        "tls": true,
        "mx-host": "aspmx.l.google.com",
        "code": 550,
        "description": "",
        "session-seconds": 0.23837614059448242,
        "utf8": true,
        "attempt-no": 1,
        "message": "5.7.1 Our system has detected that this message is\n5.7.1 not RFC 5322 compliant:\n5.7.1 'From' header has non compliant domain name.\n5.7.1 To reduce the amount of spam sent to Gmail, this message has been\n5.7.1 blocked. Please visit\n5.7.1  https://support.google.com/mail/?p=RfcMessageNonCompliant\n5.7.1 and review RFC 5322 specifications for more information. q63si546488qki.5 - gsmtp",
        "certificate-verified": true
    },
    "recipient-domain": "example.com",
    "id": "id_value",
    "campaigns": [],
    "reason": "generic",
    "user-variables": {},
    "flags": {
        "is-routed": false,
        "is-authenticated": true,
        "is-system-test": false,
        "is-test-mode": false
    },
    "log-level": "error",
    "timestamp": 1519779343.260742,
    "envelope": {
        "transport": "smtp",
        "sender": "fail2ban@example.com",
        "sending-ip": "1.2.3.4",
        "targets": "info@example.com"
    },
    "message": {
        "headers": {
            "to": "info@example.com",
            "message-id": "id_value@example.com",
            "from": "Fail2Ban <fail2ban>",
            "subject": "[Fail2Ban] sshd: started on VPS"
        },
        "attachments": [],
        "size": 526
    },
    "recipient": "info@example.com",
    "event": "failed"
}
```

Seems this is happening due to the `FromLineOverride='Yes'` set by default in the Trellis sSMTP configuration.

To fix this issue I have two options:
1. Find every service on the server that could send an email and edit manually the configuration of the sender.
2. Change the value `FromLineOverride` to `No`. But changing this parameter causes the emails sent by the server to have a `From` header dependent to the user who sends the email (e.g. `web@examaple.com`). However, there's a sSMTP option to map every user to a custom `From` address using the aliases in the `/etc/ssmtp/revaliases`.

This pull request introduces the feature to customize the `/etc/ssmtp/revaliases` during provision in the sSMTP configuration task.

The feature is disabled by default because `FromLineOverride` it's enabled by default.

This is my personal use case to solve an issue, but I think that this feature could be useful to all people that want to have a complete control on the sSMTP configuration using Trellis.

Thank you.


swalkinshaw:
🤔  interesting.

This makes sense I think. `FromLineOverride=Yes` allows WordPress to easily set the `From` header which we'd want. But as you said, anything else on the server that sends email without explicitly setting the `From` will get a default like `"Fail2Ban <fail2ban>"` which isn't valid.

I'm wondering if there's a better default though. What if we left  `FromLineOverride=Yes` *and* set these default aliases?

valentinocossar:
I did a lot of test with sSMTP and found that if I left `FromLineOverride=Yes` the result would be dependent on which command the program uses to send the email.

I'm using OVH as server provider.

Here are some tests:
1. Sent via `mail` with specified `From`:
Command: `echo "This is the message body" | mail -s "This is the subject" info@example.com -aFrom:noreply@example.com`
Result: `From:noreply@example.com`
2. Sent via `mail` with non-specified `From`:
Command: `echo "This is the message body" | mail -s "This is the subject" info@example.com`
Result: `From:user_used_to_send_mail@hostname`, where the hostname is the same I have in the `/etc/mailname`
3. Sent via `ssmtp` with specified `From`:
Command: `echo "From:noreply@example.com" | ssmtp info@example.com`
Result: `noreply@example.com`
4. Sent via `ssmtp` with non-specified `From`:
Command: `echo "Subject:Test" | ssmtp info@example.com`
Result: `From:user_used_to_send_mail@hostname`, where the hostname is the same set in the `/etc/ssmtp/ssmtp.conf`
5. Via `sendmail` the results are the same as the `ssmtp` because sendmail is a symlink of `ssmtp`.

Maybe setting the mta to `mail` for Fail2Ban would solve the problem but it's not a solution because it's not the only service that sends email from the server. That's why I specified that one of the solutions for the problem is to change the `From` for every service that sends emails, but I don't think it's easy. In fact, I've got other emails with the same problem reported in the log above but from other services that are not Fail2Ban.

In my opinion, this solution is more generic, it's not only to solve this specific problem but it's a more complete control over the sSMTP configuration if anyone needs it.

pySilver:
I've had this issue some time ago and beside using revalises I've added

```
- name: ssmtp configure default sending domain
  template:
    src: mailname.j2
    dest: /etc/mailname
```
with 
```
# mailname.jd file:
{{ mail_hostname }}
```

valentinocossar:
Hi @pySilver, thank you for the feedback, I already have `/etc/mailname` on my server with the default domain set by OVH and it's a valid domain name (`vps112233.ovh.net`) but this is not resolving the issue for me. The `From` header is overwritten by Fail2Ban (or other services) every time that sends an email and when it's not overwritten by a service the server sends the email with the default domain seen above.

valentinocossar:
@swalkinshaw at the end, what you think about the implementation of this feature? 🙂

swalkinshaw:
Thanks @valentinocossar

valentinocossar:
Thank you @swalkinshaw!

PR #958, opened 2018-03-07
Update Ansible default plugin paths in config files

fullyint:
[ansible/ansible#12378](https://github.com/ansible/ansible/pull/12378/commits/4aea1f6568eae668728390f06d57ec6afb314ca2#diff-b77962b6b54a830ec373de0602918318R187) changed Ansible's default plugin paths. At the time, Ansible set these defaults in `lib/ansible/constants.py` but now in [`lib/ansible/config/base.yml`](https://github.com/ansible/ansible/blob/v2.4.3.0-1/lib/ansible/config/base.yml#L522) (i.e., [ansible config](http://docs.ansible.com/ansible/latest/intro_configuration.html#callback-plugins); 
location change in ansible/ansible#27448).

This PR updates these default paths in Trellis `ansible.cfg` and `lib/trellis/vagrant.rb`.

The CHANGELOG entry marks this as `[BREAKING]` for the uncommon case that users have plugins in any of the out-of-date locations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* [BREAKING] Update Ansible default plugin paths in config files ([#958](https://github.com/roots/trellis/pull/958))
 * Add Nginx `ssl.no-default.conf` to drop requests for unknown hosts ([#888](https://github.com/roots/trellis/pull/888))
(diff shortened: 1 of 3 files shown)


partounian:
To implement the past few commits, do we just need to run deploy or do we need reconfigure the server with the ansible-playbook command?

swalkinshaw:
Just the commits in this PR? Don't need to do anything other than merge.

Issue #902, opened 2017-10-04
DHCP + NFS fails on initial `vagrant up`


TangRufus:
## Submit a feature request or bug report

- [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/guidelines/blob/master/CONTRIBUTING.md)
- [ ] This is a feature request
- [x] This is a bug report
- [x] This request isn't a duplicate of an [existing issue](https://github.com/roots/trellis/issues)
- [x] I've read the [docs](https://roots.io/trellis/docs) and followed them (if applicable)
- [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/c/trellis) forums

Replace any `X` with your information.

---

**What is the current behavior?**

`vagrant up` the first time would give:
```
Unmounting Virtualbox Guest Additions ISO from: /mnt
==> default: Checking for guest additions in VM...
==> default: Setting hostname...
==> default: Configuring and enabling network interfaces...
No guest IP was given to the Vagrant core NFS helper. This is an
internal error that should be reported as a bug.
```

**What is the expected or desired behavior?**

It should start up the VM without error.

---

## Bug report

**Please provide steps to reproduce, including full log output:**

`vagrant up` In a freshly cloned trellis repo.

**Please describe your local environment:**

Ansible version: 2.4.0.0

OS: 10.13

Vagrant version: 2.0.0

**Where did the bug happen? Development or remote servers?**

Local machine.

**Other relevant information:**

Related pull requests: #898, #892

TangRufus:
More info: If you have `vagrant up` previously, you won't encounter this issue.

Related vagrant issue: https://github.com/hashicorp/vagrant/issues/7070

Current workaround: Use static IP on initial `vagrant up`, then change it back to `dhcp`

swalkinshaw:
I just tried this again. The one additional step I did which I didn't before when testing was to delete the local `.vagrant`.

Tried twice and both times got stuck at `Installing NFS client...`.

I'm going to revert that PR for now.

swalkinshaw:
Should note I'm using Vagrant 1.9.7

TangRufus:
Culprit: Me not tested the PRs thoroughly. Third bug within a month, i feel ashamed of myself. Sorry guys.

Complicity:
```
 INFO retryable: Retryable exception raised: #<Vagrant::Errors::VirtualBoxGuestPropertyNotFound: Could not find a required VirtualBox guest property:
[...]

PR #960, opened 2018-03-08
Vagrant: Use DHCP by default

TangRufus:
Upstream issues resolved.

Related: https://github.com/roots/trellis/issues/902

Tested on:
* macOS 10.13.3
* ansible 2.4.3.0
* vagrant-bindfs 1.1.0
* vagrant-hostmanager 1.8.7

# Help Wanted

This patch needs tests on other OS.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Vagrant: Use DHCP by default ([#960](https://github.com/roots/trellis/pull/960))
 * Fix `raw_vars` feature to properly handle int values ([#959](https://github.com/roots/trellis/pull/959))
 * [BREAKING] Update Ansible default plugin paths in config files ([#958](https://github.com/roots/trellis/pull/958))
 * Add Nginx `ssl.no-default.conf` to drop requests for unknown hosts ([#888](https://github.com/roots/trellis/pull/888))
--- a/vagrant.default.yml
+++ b/vagrant.default.yml
@@ -1,5 +1,5 @@
 ---
-vagrant_ip: '192.168.50.5'
+vagrant_ip: 'dhcp'
 vagrant_cpus: 1
 vagrant_memory: 1024 # in MB
 vagrant_box: 'bento/ubuntu-16.04'


swalkinshaw:
Do you have a reference to the upstream issues?

TangRufus:
I tried `DHCP` out of curiosity and it suddenly worked. Not sure what fixed it.

swalkinshaw:
Working for me too! Even deleted `.vagrant` to test like last time.

```
macOS 10.10.5
ansible 2.4.0.0
Vagrant 2.0.2
vagrant-bindfs (1.1.0)
vagrant-hostmanager (1.8.7)
```

oxyc:
I tried this again over at Drupal VM. One earlier issue I had is now fixed but I still encounter a few annoyances:

1. If I switch to DHCP on an existing VM using `vagrant reload` I get and error. Running `vagrant halt; vagrant up` works.
    > NFS requires a host-only network to be created.
    > Please add a host-only network to the machine (with either DHCP or a static IP) for NFS to work.
2. If I try and switch back to an IP after having used DHCP it seems pretty impossible. It just gets stuck (this might be drupal vm related though as we also support vagrant-hostsupdater)
    > drupalvm: Warning: Remote connection disconnect. Retrying...
    > drupalvm: Warning: Connection reset. Retrying...

swalkinshaw:
Thanks @oxyc. 2nd point doesn't seem like a big deal since I assume it would be rare. But the 1st point is more important since anyone with an existing VM who updates this code would get the DHCP default breaking their current VMs.

Wonder if there's a way to skip it for existing VMs.

TangRufus:
> If I switch to DHCP on an existing VM using vagrant reload I get and error. Running vagrant halt; vagrant up works.

Run `$ vagrant hostmanager` when this error show up:
```bash
NFS requires a host-only network to be created.
Please add a host-only network to the machine (with either DHCP or a static IP) for NFS to work.
```

swalkinshaw:
Still on the fence about this.

Here's a nice summary of potential issues: https://github.com/geerlingguy/drupal-vm/pull/1559#issuecomment-387200433

PR #962, opened 2018-03-09
Add option to enable FastCGI background updates

inthedeepend:
- Options in development,staging,production group_vars
- Conditionally add parameter in NGINX site host file `wordpress-site.conf.j2`

### DIFF
--- a/roles/wordpress-setup/templates/wordpress-site.conf.j2
+++ b/roles/wordpress-setup/templates/wordpress-site.conf.j2
@@ -34,6 +34,7 @@ server {
 
   {% block cache_conditions -%}
   {% if item.value.cache is defined and item.value.cache.enabled | default(false) -%}
+
   # Fastcgi cache conditions
   set $skip_cache 0;
   if ($query_string != "") {
@@ -45,6 +46,11 @@ server {
   if ($http_cookie ~* "{{ item.value.cache.skip_cache_cookie | default(nginx_skip_cache_cookie) }}") {
     set $skip_cache 1;
   }
+
+  # Update an expired cache item in the background, serve stale cache to client.
+  {% if item.value.cache.enabled and item.value.cache.background_update %}
+    fastcgi_cache_background_update on
+  {% endif %}
 
   {% endif -%}
   {% endblock -%}
--- a/roles/wordpress-setup/templates/wordpress-site.conf.j2
+++ b/roles/wordpress-setup/templates/wordpress-site.conf.j2
@@ -47,11 +47,6 @@ server {
     set $skip_cache 1;
   }
 
-  # Update an expired cache item in the background, serve stale cache to client.
-  {% if item.value.cache.enabled and item.value.cache.background_update %}
-    fastcgi_cache_background_update on
-  {% endif %}
-
   {% endif -%}
   {% endblock -%}
 
@@ -195,6 +190,11 @@ server {
     fastcgi_cache_valid {{ item.value.cache.duration | default(nginx_cache_duration) }};
     fastcgi_cache_bypass $skip_cache;
     fastcgi_no_cache $skip_cache;
+
+    # Update an expired cache item in the background, serve stale cache to client.
+    {% if item.value.cache.background_update | default(true) %}
+    fastcgi_cache_background_update on;
+    {% endif %}
 
     {% endif -%}
     {% endblock -%}
--- a/roles/wordpress-setup/templates/wordpress-site.conf.j2
+++ b/roles/wordpress-setup/templates/wordpress-site.conf.j2
@@ -34,7 +34,6 @@ server {
 
   {% block cache_conditions -%}
   {% if item.value.cache is defined and item.value.cache.enabled | default(false) -%}
-
   # Fastcgi cache conditions
   set $skip_cache 0;
   if ($query_string != "") {
@@ -190,11 +189,7 @@ server {
     fastcgi_cache_valid {{ item.value.cache.duration | default(nginx_cache_duration) }};
     fastcgi_cache_bypass $skip_cache;
     fastcgi_no_cache $skip_cache;
-
-    # Update an expired cache item in the background, serve stale cache to client.
-    {% if item.value.cache.background_update | default(true) %}
-    fastcgi_cache_background_update on;
-    {% endif %}
+    fastcgi_cache_background_update {{ item.value.cache.background_update | default(nginx_cache_background_update) }};
 
     {% endif -%}
     {% endblock -%}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Enable `fastcgi_cache_background_update` by default ([#962](https://github.com/roots/trellis/pull/962))
 * Bump Ansible `version_tested_max` to 2.5.3 ([#981](https://github.com/roots/trellis/pull/981))
 * deploy.sh: Return non-zero exit code when misuse  ([#990](https://github.com/roots/trellis/pull/990))
 * Add CSP `frame-ancestors`, make `X-Frame-Options` conditional ([#977](https://github.com/roots/trellis/pull/977))
--- a/group_vars/development/wordpress_sites.yml
+++ b/group_vars/development/wordpress_sites.yml
@@ -17,4 +17,3 @@ wordpress_sites:
       provider: self-signed
     cache:
       enabled: false
-      background_update: false
--- a/group_vars/development/wordpress_sites.yml
+++ b/group_vars/development/wordpress_sites.yml
@@ -17,3 +17,4 @@ wordpress_sites:
       provider: self-signed
     cache:
       enabled: false
+      background_update: false
--- a/group_vars/production/wordpress_sites.yml
+++ b/group_vars/production/wordpress_sites.yml
@@ -19,4 +19,3 @@ wordpress_sites:
       provider: letsencrypt
     cache:
       enabled: false
-      background_update: false
--- a/group_vars/production/wordpress_sites.yml
+++ b/group_vars/production/wordpress_sites.yml
@@ -19,3 +19,4 @@ wordpress_sites:
       provider: letsencrypt
     cache:
       enabled: false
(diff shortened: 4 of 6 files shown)


roles/wordpress-setup/templates/wordpress-site.conf.j2
```diff
@@ -46,6 +47,11 @@ server {
     set $skip_cache 1;
   }
 
+  # Update an expired cache item in the background, serve stale cache to client.
+  {% if item.value.cache.enabled and item.value.cache.background_updates %}
```
swalkinshaw: `background_update` not `background_updates`

swalkinshaw:
I wonder if there's any reason *not* to make this the default?

roles/wordpress-setup/templates/wordpress-site.conf.j2
```diff
@@ -46,6 +47,11 @@ server {
     set $skip_cache 1;
   }
 
+  # Update an expired cache item in the background, serve stale cache to client.
+  {% if item.value.cache.enabled and item.value.cache.background_update %}
+    fastcgi_cache_background_update on
```
partounian: forgot the semi-colon at the end

partounian:
In what cases does it make sense to enable background_update?

swalkinshaw:
The better question is the opposite to me.

Right now the cache is time based. Say it's 60s, and a client hits the site 60.1s after it was last cached. This would result in a cache miss and that client would incur the cost of a full request hitting WP.

With background updates, and the same scenario above, that client would still result in a cache hit, but the cache would be updated in the background. This means no actual client would *ever* incur a cache miss. They'd either get stale cached data, or the next cache hit. This is obviously preferable as long as it works as we'd expect. Meaning if we trust Nginx to properly update the cache in the background within a reasonable time frame, it's a better user experience with barely any more cache time added.

Here's a good article on this: https://siipo.la/blog/never-miss-the-cache-with-nginx-microcaching

roles/wordpress-setup/templates/wordpress-site.conf.j2
partounian: Also should this be in the cache_config block?

partounian:
Then according to that article should we also consider the following?
```
fastcgi_cache_lock on;
fastcgi_cache_use_stale updating;
```
Also, currently fastcgi_cache_valid is set to 30s, should we specify 200 301 302 404 and possibly increase caching to 1m?

swalkinshaw:
We already set those.

Time is just a default, anyone can change it. Maybe we should only specific `200` though or make it a variable. I'm not really sure.

roles/wordpress-setup/templates/wordpress-site.conf.j2
swalkinshaw: yeah I'd prefer this to be grouped in `cache_config` 👍

partounian:
My mistake I only checked site config file and not the general nginx.conf

inthedeepend:
Updated PR with cache_config grouping and missing `;`

I don't know where I stand on making this default behaviour or not, I think it could lead to some confusion - like I said in my initial suggestion - this isn't a perfect solution for all applications. Which makes me think it should be opt in behaviour.

roles/wordpress-setup/templates/wordpress-site.conf.j2
```diff
@@ -175,6 +176,11 @@ server {
     fastcgi_cache_bypass $skip_cache;
     fastcgi_no_cache $skip_cache;
 
+    # Update an expired cache item in the background, serve stale cache to client.
+    {% if item.value.cache.background_update %}
+    fastcgi_cache_background_update on
```
partounian: ; is still missing

partounian:
I vote for it being default behavior and replacing `if item.value.cache.background_update` with setting the value as `{{ item.value.cache.background_update | default(true) }}`. Even if we set it false we want an if check but a `{{ value | default(false) }}`

swalkinshaw:
>  like I said in my initial suggestion - this isn't a perfect solution for all applications. 

@inthedeepend I'm still curious why you think this is the case. Is there a downside/con I'm missing?

inthedeepend:
@swalkinshaw - maybe I'm missing something actually, I guess this works in conjunction to whatever the cache expiry is in the group_vars, and that's set at the discretion/requirements of the project. +1 for default. My bad.

inthedeepend:
Set to default to true. :)

roles/wordpress-setup/templates/wordpress-site.conf.j2
```diff
@@ -175,6 +176,11 @@ server {
     fastcgi_cache_bypass $skip_cache;
     fastcgi_no_cache $skip_cache;
 
+    # Update an expired cache item in the background, serve stale cache to client.
+    {% if item.value.cache.background_update | default(true) %}
+    fastcgi_cache_background_update on;
```
partounian: Sorry for the confusion I believe you should remove the if and simply put the one-liner
`fastcgi_cache_background_update {{ item.value.cache.background_update | default(on) }};`
Also, please make sure to test if you have not done so :)

Thank you for being patient with us and thank you for your contribution :)

fullyint:
@inthedeepend This is great. Thank you!

It works as is, but more style opinions incoming! :scream:

For the sake of consistency with existing Trellis caching options, I think there are two patterns this new `fastcgi_cache_background_update` option should follow.

**Where defaults are defined.** Notice that the cache options in the [docs](https://roots.io/trellis/docs/fastcgi-caching/) are absent from [`wordpress_sites`](https://github.com/roots/trellis/blob/75518c1fc6bed541765de51b7b8a3191a7f064b6/group_vars/production/wordpress_sites.yml#L20-L21) (except `enabled`) which works because they are in the role [`defaults`](https://github.com/roots/trellis/blob/75518c1fc6bed541765de51b7b8a3191a7f064b6/roles/wordpress-setup/defaults/main.yml#L30-L32). This keeps `wordpress_sites` streamlined.

**How defaults are defined.** The cache directives in `wordpress-site.conf.j2` follow [this pattern](https://github.com/roots/trellis/blob/master/roles/wordpress-setup/templates/wordpress-site.conf.j2#L174):

```
directive {{ item.value.cache.directive | default(nginx_cache_directive) }};
```

A user could define `nginx_cache_directive` as a global override and/or set a per-site override by defining the `cache.directive` in `wordpress_sites`.

:star: To be consistent with these patterns, could you...
* remove `background_update` setting from `wordpress_sites`
* add `nginx_cache_background_update: on` to [`Fastcgi cache params` defaults](https://github.com/roots/trellis/blob/75518c1fc6bed541765de51b7b8a3191a7f064b6/roles/wordpress-setup/defaults/main.yml#L30-L32)
* drop all changes from `wordpress-site.conf.j2` (including the new blank lines, comments, and `{% if %}` blocks), leaving only the following:
  ```diff
    fastcgi_no_cache $skip_cache;
  + fastcgi_cache_background_update {{ item.value.cache.background_update | default(nginx_cache_background_update) }};
  ```

I'm inclined to drop the explanatory comment because no other `Fastcgi cache settings` have comments, but I don't feel too strongly about it.

I'm inclined to not add the extra blank lines because then the templated conf on the server would have double blank lines in some spots. Of course, it could be argued that the goal should be readability in the Trellis `wordpress-site.conf.j2` template (hence your new blank lines) rather than perfect formatting in the conf on the server (the current focus).

**http codes.** 
>should we specify 200 301 302 404 and possibly increase caching to 1m? ([comment above](https://github.com/roots/trellis/pull/962#issuecomment-372137826))

Looks like 200, 301, and 302 are covered by default ([nginx docs](http://nginx.org/en/docs/http/ngx_http_fastcgi_module.html#fastcgi_cache_valid)). I don't know much about why and when to also include 404, but I'm inclined to continue with the default because desire for an alternative hasn't ever come up that I remember. Users could change it using child templates.

partounian:
Hey @inthedeepend any chance you can update this with the fixes? I would love for this to be merged. 
If you are not able to, I wonder what is etiquette for me to apply these changes.

swalkinshaw:
Thanks everyone ❤️

TangRufus:
>Note that it is necessary to [allow](https://nginx.org/en/docs/http/ngx_http_fastcgi_module.html#fastcgi_cache_use_stale_updating) the usage of a stale cached response when it is being updated. 
>
> --- [https://nginx.org/en/doc/http/ngx_http_fastcgi_module.html#fastcgi_cache_background_update](https://nginx.org/en/doc/http/ngx_http_fastcgi_module.html#fastcgi_cache_background_update)

The [blog post](https://siipo.la/blog/never-miss-the-cache-with-nginx-microcaching) @swalkinshaw linked also updated about the issues with 404 pages on 17 May 2018.

Do we need to add `fastcgi_cache_use_stale` as well?

swalkinshaw:
We already have ` fastcgi_cache_use_stale updating error timeout invalid_header http_500;` in the base Nginx conf.

`fastcgi_cache_use_stale updating` should take care of background updates I assume?

fullyint:
@TangRufus thanks for pointing out the article update regarding 404 pages. I haven't tested the need nor the solution, but the description makes sense to me. Do you envision a change like this?

```diff
- fastcgi_cache_valid {{ item.value.cache.duration | default(nginx_cache_duration) }};
+ fastcgi_cache_valid 200 301 302 404 {{ item.value.cache.duration | default(nginx_cache_duration) }};
```

Or we could do shorter expiration for 404s, but I doubt it makes a big difference:

```diff
  fastcgi_cache_valid {{ item.value.cache.duration | default(nginx_cache_duration) }};
+ fastcgi_cache_valid 404 1s;
```

craigpearson:
On one of our installs nginx was running on version 1.11.5, as this feature was added in 1.11.10 this causes an emergency log of:

```
nginx: [emerg] unknown directive "fastcgi_cache_background_update" in /etc/nginx/sites-enabled/example.com.conf:78
```

To update nginx manually you can run:

```
sudo add-apt-repository ppa:nginx/stable
sudo apt-get update
sudo apt-get install nginx
```

PR #968, opened 2018-03-16
Refactor --subdomains flag in the Install WP task

primozcigler:
According to the [documentation](https://developer.wordpress.org/cli/commands/core/multisite-install/) the existance of the --subdomains is boolean value. If the `item.value.multisite.subdomains` was set to false (or not set) this task was treating it as it's set to true.

This commit is fixing this.


swalkinshaw:
😱 good find. Did this actually cause a bug or break something? I actually wonder if `--subdomains=false` works despite their docs.

Either way, it's more accurate 👍

primozcigler:
@swalkinshaw it does not work as it was before. That's how I found the bug 😉 

And I've also tested it and with my fix, it works as it's supposed to.

Issue #969, opened 2018-03-22
Feature request: Add AWS CodeCommit Repo support


newloong:
## Submit a feature request or bug report

- [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/guidelines/blob/master/CONTRIBUTING.md)
- [x] This is a feature request
- [x] This is a bug report
- [x] This request isn't a duplicate of an [existing issue](https://github.com/roots/trellis/issues)
- [x] I've read the [docs](https://roots.io/trellis/docs) and followed them (if applicable)
- [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/c/trellis) forums

---

**What is the current behavior?**

No AWS CodeCommit Repo support

**What is the expected or desired behavior?**

Add AWS CodeCommit Repo support

---

## Bug report

Please provide steps to reproduce, including full log output:

Got the following error using AWS CodeCommit as Repo(repo: ssh://git-codecommit.us-west-2.amazonaws.com/v1/repos/bedrock.example.com)

TASK [Ensure repo is valid] 
System info:
  Ansible 2.4.2.0; Darwin
  Trellis at "Git: Ignore `vagrant.local.yml`"

Invalid Git repository.
Ensure that your site's `repo` variable is defined in
`group_vars/staging/wordpress_sites.yml` and uses the SSH format (example:
git@github.com:roots/bedrock.git)
More info:
> https://roots.io/trellis/docs/deploys/

fatal: [staging_host]: FAILED! => {"changed": false}
	to retry, use: --limit @/Users/name/sites/example.com/trellis/deploy.retry

PLAY RECAP 
localhost                  : ok=0    changed=0    unreachable=0    failed=0
staging_host               : ok=2    changed=0    unreachable=0    failed=1

**Please describe your local environment:**

Ansible version: 2.4.2.0

OS: macOS 10.13.3 (17D102)

Vagrant version: 2.0.2

**Where did the bug happen? Development or remote servers?**

Staging

## Feature Request

Add AWS CodeCommit Repo support

swalkinshaw:
I'm assuming CodeCommit doesn't work if you append the `.git` extension?

newloong:
@swalkinshaw Yes, it doesn't work, both append .git extension and without .git extension. The URI of CodeCommit repo like:
`ssh://Your-SSH-Key-ID@git-codecommit.us-east-2.amazonaws.com/v1/repos/MyDemoRepo`

[...]

PR #975, opened 2018-04-01
Add AWS CodeCommit Repo support.

newloong:
Hey,

This pull request add AWS CodeCommit repo support.

Related: #969 

Tested on:

macOS 10.13.3
ansible 2.4.3.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Support git url format `ssh://user@host/path/to/repo` ([#975](https://github.com/roots/trellis/pull/975))
 * Fix path to h5bp/mime.types ([#974](https://github.com/roots/trellis/pull/974))
 * Vendor h5bp Nginx configs ([#973](https://github.com/roots/trellis/pull/973))
 * Add support for sSMTP revaliases configuration ([#956](https://github.com/roots/trellis/pull/956))
--- a/deploy.yml
+++ b/deploy.yml
@@ -28,6 +28,6 @@
           Ensure that your site's `repo` variable is defined in `group_vars/{{ env }}/wordpress_sites.yml` and uses the SSH format (example: git@github.com:roots/bedrock.git)
           More info:
           > https://roots.io/trellis/docs/deploys/
-      when: project.repo is not defined or not project.repo | match("(^ssh:|.+@.+:.+)")
+      when: project.repo is not defined or not project.repo | match("^ssh:\/\/.*@.*|.*@.*:.*\.git")
   roles:
     - deploy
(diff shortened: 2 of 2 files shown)


swalkinshaw:
I'm wondering if this check is even useful at this point. There's later connection test to make sure we can connect to the Git repo.

fullyint:
@newloong Thanks for this. We discussed internally and agree we should enable the `ssh://` form.

**Background.** I think #459 added the regexp validation to quickly warn users against the `http(s)://` format that often doesn't allow connection to private repos during later deploy tasks. I think #516 adjusted the regexp to ensure users had the `:` in their repo's git format (:x: `git@github.com/roots/bedrock` but ✅ `git@github.com:roots/bedrock`).

**Your regexp.** Could you revise your regexp to disallow the `http(s)` repo format? See proposed revision in example below. Does that regexp work for your CodeCommit repo?

**Example revision.** Maybe the fail `msg` could give the `ssh://` format as an example too, e.g.,

```
    - name: Ensure repo is valid
      connection: local
      fail:
        msg: |
          Invalid Git repository format.
          Ensure that your site's `repo` variable is defined in
          `group_vars/{{ env }}/wordpress_sites.yml` and uses an SSH format.
          Examples:
            - git@github.com:roots/bedrock.git
            - ssh://git@github.com/roots/bedrock
          More info:
          > https://roots.io/trellis/docs/deploys/
      when: project.repo is not defined or not project.repo | match("(^ssh:|.+@.+:.+)")
```

fullyint:
@newloong Thanks for the revised regexp. I like how you got rid of the unnecessary `( )` from my suggested example.

**.git** I should have mentioned it explicitly, but given that `git clone git@github.com:roots/bedrock` succeeds without the `.git` at the end, could you remove the regexp's requirement that `repo` values in `\.git`? I realize the original regexp in trellis had the `.git` so it made sense that you kept it.

**slashes.** Your pattern requires `ssh:\/\/`. I don't think it is necessary to escape the slashes. I don't think slashes are special/control characters in this context (a la https://docs.python.org/2/library/re.html#regular-expression-syntax) and aren't being used as pattern delimiters. Could you use the visually simpler `ssh://`? 

**`.*` vs `.+`** This last comment is trivial and you can ignore it if you want. I think each instance of `.*` truly needs "one or more" characters (`.+`); it would be an incorrect format if a user were to specify zero characters in any of those locations of the pattern, a la the "zero or more" of `.*`. Once again, I realize the original regexp used `.*`, but I think it'd be nice to get the extra accuracy given that it comes at no additional characters/cost.

**summary.** So, altogether I'm proposing:
```diff
- match("^ssh:\/\/.*@.*|.*@.*:.*\.git")
+ match("^ssh://.+@.+|.+@.+:.+")
```

Thanks again! Your help with this PR brings Trellis a good improvement.

PR #977, opened 2018-04-10
Add CSP frame-ancestors, make X-Frame-Options conditional

fullyint:
This PR adds a workaround to https://discourse.roots.io/t/safari-conflicting-multiple-x-frame-options/10077 until https://core.trac.wordpress.org/ticket/40020 is resolved.  

### `frame-ancestors` CSP

This PR adds the [frame-ancestors](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors) CSP directive which has obsoleted the `X-Frame-Options` header.

Note that it is acceptable to have [multiple CSPs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy#Multiple_content_security_policies) (i.e., in case users add other CSP headers).

> CSP is designed to be fully backward compatible... Browsers that don't support it still work with servers that implement it, and vice-versa: browsers that don't support CSP simply ignore it, functioning as usual, defaulting to the standard same-origin policy for web content.  _[--reference](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP)_

### Dynamic `X-Frame-Options`

This PR retains the `X-Frame-Options` header for the sake of older browsers, but returns an empty `X-Frame-Options` value for WordPress Customizer content. This prevents the Safari browser's conflict between `SAMEORIGIN` and the `ALLOW-FROM` option that WordPress adds on its own.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Add CSP `frame-ancestors`, make `X-Frame-Options` conditional ([#977](https://github.com/roots/trellis/pull/977))
 * Common: Install `git` instead of `git-core`  ([#989](https://github.com/roots/trellis/pull/989))
 * Add `xdebug.remote_autostart` to simplify xdebug sessions  ([#985](https://github.com/roots/trellis/pull/985))
 * Enable nginx to start on boot ([#980](https://github.com/roots/trellis/pull/980))
--- a/roles/nginx/templates/h5bp/directive-only/extra-security.conf
+++ b/roles/nginx/templates/h5bp/directive-only/extra-security.conf
@@ -6,11 +6,7 @@ set $x_frame_options SAMEORIGIN;
 if ($arg_customize_changeset_uuid) {
   set $x_frame_options "";
 }
-add_header X-Frame-Options $x_frame_options;
-
-# The HTTP Content-Security-Policy (CSP) frame-ancestors directive specifies valid parents
-# that may embed a page using <frame>, <iframe>, <object>, <embed>, or <applet>.
-add_header Content-Security-Policy "frame-ancestors 'self'";
+add_header X-Frame-Options $x_frame_options always;
 
 # MIME type sniffing security protection
 #	There are very few edge cases where you wouldn't want this enabled.
@@ -25,3 +21,7 @@ add_header X-XSS-Protection "1; mode=block" always;
 # CSP can be quite difficult to configure, and cause real issues if you get it wrong
 # There is website that helps you generate a policy here http://cspisawesome.com/
 # add_header Content-Security-Policy "default-src 'self'; style-src 'self' 'unsafe-inline'; script-src 'self' https://www.google-analytics.com;" always;
+
+# The HTTP Content-Security-Policy (CSP) frame-ancestors directive specifies valid parents
+# that may embed a page using <frame>, <iframe>, <object>, <embed>, or <applet>.
+add_header Content-Security-Policy "frame-ancestors 'self'" always;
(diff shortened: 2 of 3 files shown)


swalkinshaw:
Where is `arg_customize_changeset_uuid` from? Does that come from the customizer?

This looks good, but I think we should just move this out of the h5bp directory now and make it ours. If we customize these templates, they belong to us :)

fullyint:
### arg_customize_changeset_uuid

> Where is `arg_customize_changeset_uuid` from? Does that come from the customizer?

Yep, from the Customizer's embedded content:
```
<iframe src="https://example.com/?customize_changeset_uuid=8c0f9f64-2daf..."></iframe>
```
Nginx supports variables such as [$arg\__name_](http://nginx.org/en/docs/http/ngx_http_core_module.html#variables) for the "argument _name_ in the request line." Nginx allows checking the presence of query parameters with `if ($arg_name) {}`.

I figured the presence of that query parameter was the most convenient indicator of whether a request was for the Customizer's embedded content. Although a malicious embed could spoof this strategy by adding a fake `customize_changeset_uuid` query parameter, it should fail. 
* An attempt to embed a real `customize_changeset_uuid` on some other domain should fail because WP code adds `X-Frame-Options` and `frame-ancestors` headers for embedded content. Also, embed content will fail with `non existing changeset UUID` unless the embed and parent `ORIGIN` are the same, and the user is logged in to wp-admin. Discussion at [trac-39128](https://core.trac.wordpress.org/ticket/39128).
* An attempt to use a fake `customize_changeset_uuid` query parameter outside the Customizer context just to fool this PR's proposed conditional `X-Frame-Options` will fail for [most modern browsers](https://caniuse.com/#feat=contentsecuritypolicy2) because this PR's added `Content-Security-Policy: frame-ancestors 'self'` usually overrides `X-Frame-Options` anyway.

### Move from h5bp to Trellis core

>I think we should just move this out of the h5bp directory now and make it ours.

I added a commit commenting out the `X-Frame-Options` in the h5bp file to prevent duplication, then adding this PR's revised headers to a new `{% block embed_security -%}` in `wordpress-site.conf.j2`. The headers restricting embeds are enabled by default but users could define `nginx_embed_security: false` globally in a `group_vars` file or per site in `wordpress_sites`, the latter location taking precedence.

My choice of `wordpress-site.conf.j2` as the home for these headers, vs. a new include file, was in spirit of "Fewer templates/includes, but still DRY," a section in #740's comments. I chose against trying to add the headers as a new Trellis core file in `includes.d` because I suspect users see `nginx-includes` and `includes.d` as exact parallels, and the user's province only. I also didn't want to figure out coordination with the [`includes.d` cleanup](https://roots.io/trellis/docs/nginx-includes/#file-cleanup) feature.

If Trellis begins to have or desire multiple Trellis core Nginx include files, perhaps the [`Copy h5bp configs`](https://github.com/roots/trellis/blob/170594515710bf8867464bc42e288392e5dd3833/roles/nginx/tasks/main.yml#L28) task could become `Copy Nginx configs` and use `with_items: [h5bp, trellis]` where `h5bp` is the same directory as currently and `trellis` is a new directory with Trellis core confs such as `embed-security-headers.conf`. In such a case of non-templated files (copied only), these directories could be in a `roles/nginx/files` directory instead of `roles/nginx/templates` where `h5bp` is currently.

strarsis:
@fullyint: Can this config block/option key `nginx_embed_security` also 
be used for allowing to embed a specific WordPress site? 
For example, the site example.com should be embeddable by others,
specifically everything that ends with `?embed=1` in the URL.

Issue #943, opened 2018-01-29
[DEPRECATION WARNING]: The use of 'include' for tasks has been deprecated.


runofthemill:
## Submit a feature request or bug report

- [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/guidelines/blob/master/CONTRIBUTING.md)
- [ ] This is a feature request
- [x] This is a bug report
- [x] This request isn't a duplicate of an [existing issue](https://github.com/roots/trellis/issues)
- [x] I've read the [docs](https://roots.io/trellis/docs) and followed them (if applicable)
- [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/c/trellis) forums

Replace any `X` with your information.

---

**What is the current behavior?**

```
- extracting geerlingguy.daemonize to /home/vagrant/trellis/vendor/roles/geerlingguy.daemonize
- geerlingguy.daemonize (1.2.0) was installed successfully
    default: Running ansible-playbook...
[DEPRECATION WARNING]: The use of 'include' for tasks has been deprecated. Use
'import_tasks' for static inclusions or 'include_tasks' for dynamic inclusions.
 This feature will be removed in a future release. Deprecation warnings can be
disabled by setting deprecation_warnings=False in ansible.cfg.
[DEPRECATION WARNING]: include is kept for backwards compatibility but usage is
 discouraged. The module documentation details page may explain more about this
 rationale.. This feature will be removed in a future release. Deprecation
warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.

PLAY [WordPress Server: Install LEMP Stack with PHP 7.1 and MariaDB MySQL] *****
```

**What is the expected or desired behavior?**

No deprecation warning for "`The use of 'include' for tasks`"

---

## Bug report

(delete this section if not applicable)

**Please provide steps to reproduce, including full log output:**

X

**Please describe your local environment:**

Ansible version: 2.4.2.0

OS: macOS 10.13.3

Vagrant version: 2.0.1

**Where did the bug happen? Development or remote servers?**

Development

**Please provide a repository or your `wordpress_sites` config (if possible):**

N/A

**Is there a related [Discourse](https://discourse.roots.io/) thread or were any utilized (please link them)?**

No

runofthemill:
I'll try and submit a PR against this to fix in the next few days :)

fullyint:
[...]

Issue #634, opened 2016-08-09
Feature request: Semantic Versioning/More Frequent Releases


runofthemill:
## Submit a feature request or bug report
- [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/guidelines/blob/master/CONTRIBUTING.md)
- [x] This is a feature request
- [ ] This is a bug report
- [x] This request isn't a duplicate of an [existing issue](https://github.com/roots/trellis/issues)
- [ ] I've read the [docs](https://roots.io/trellis/docs) and followed them (if applicable)

Replace any `X` with your information.

---

**What is the current behavior?**

Since 0.9.7 was released, there have been 100+ commits to the repo, and recently some breaking changes have been introduced. It's challenging to keep Trellis up to date while at the same time avoiding large breaking changes (e.g. Upgrade Ubuntu from 14.04 Trusty to 16.04 Xenial) when smaller bug fixes are interspersed with larger ones. It's impossible to get SSL with Let's Encrypt to work on 0.9.7, but merging `HEAD` will introduce breaking changes. Additionally, the documentation for Trellis has changed a bit, and so no longer corresponds to the latest release. And since all the docs for Trellis/Bedrock/Sage are in one repo, it's not as easy as selecting the 0.9.7 tag to find the right docs.

**What is the expected or desired behavior?**

It would be great if Trellis adopted Semantic Versioning, or something with similar principles. Alternatively, some sort of rules around feature branching to prevent the issues mentioned.

SemVer isn't necessarily a panacea, but I wanted to start a discussion and see what people think. Perhaps the solution is more frequent releases, or more comprehensive documentations to changes. It's asking for more work upfront, but should reduce issues that users run into and don't know how to address (along with hopefully fewer help requests on Discourse). 

Again, this is really to start a discussion about how to make Trellis easier to use/better documented.

---
## Feature Request

**Please provide use cases for changing the current behavior:**

See above :)

swalkinshaw:
We mostly agree with you and have discussed this internally before. Couple points:

1. Yes we should release more often.
2. Breaking changes since 0.9.7 is fine even semantically speaking. All these updates have been to master without a release so there's nothing wrong with that (except that we should have done more releases in between).

Our plan (which wasn't much of one), was to start enforcing sem ver as of 1.0. 1.0 has taken way too long as we've been trying to cram too much in before it happened. Something else always came up to move it back like Ubuntu 16.04.

[...]

PR #981, opened 2018-04-24
Bump Ansible version_tested_max to 2.5.3

fullyint:
### Jinja2 test format

Converts Jinja2 tests from filter format to `var is testname` format ([porting guide](https://docs.ansible.com/ansible/2.5/porting_guides/porting_guide_2.5.html#jinja-tests-used-as-filters)).

### Erroneous warnings in Ansible 2.5.0

Encourages users on Ansible 2.5.0 to upgrade to avoid erroneous warnings fixed in ansible/ansible#37538. After extensive investigation I found only one hack to prevent the warning in 2.5.0 but it seemed slightly risky (modified `remote_user` and `become_pass` entries in [`MAGIC_VARIABLE_MAPPING`](https://github.com/ansible/ansible/blob/v2.5.0/lib/ansible/constants.py#L110) processed in [`clean_facts`](https://github.com/ansible/ansible/blob/v2.5.0/lib/ansible/vars/clean.py#L66-L67)). I chose instead the conservative approach of printing encouragement to upgrade. An example of what users would see:
```
PLAY [Ensure necessary variables are defined] ***********************************
 [WARNING]: You Ansible version is 2.5.0. Consider upgrading your Ansible version to
avoid erroneous warnings such as `Removed restricted key from module data...`

PLAY [Test Connection and Determine Remote User] ********************************
...

TASK [connection : Check whether Ansible can connect as root] *******************
ok: [138.68.203.29 -> localhost]
...

TASK [connection : Set remote user for each host] *******************************
 [WARNING]: Removed restricted key from module data: ansible_user = admin
...
```

### Release version number format

Uses the release version number format 2.5.1 instead of 2.5.1.0 in `vagrant.default.yml` and  `lib/trellis/plugins/vars/version.py` because 2.5.1 is the format of `__version__`. Vagrant's ansible_local fails to find a version 2.5.1.0 to install, but succeeds with 2.5.1.

### Deprecation warnings for `include`

As a side note, Ansible 2.5.0 and 2.5.1 no longer print `[DEPRECATION WARNING]: The use of 'include' for tasks has been deprecated` (see #943). I haven't looked into it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Bump Ansible `version_tested_max` to 2.5.3 ([#981](https://github.com/roots/trellis/pull/981))
 * deploy.sh: Return non-zero exit code when misuse  ([#990](https://github.com/roots/trellis/pull/990))
 * Add CSP `frame-ancestors`, make `X-Frame-Options` conditional ([#977](https://github.com/roots/trellis/pull/977))
 * Common: Install `git` instead of `git-core`  ([#989](https://github.com/roots/trellis/pull/989))
--- a/deploy.yml
+++ b/deploy.yml
@@ -28,6 +28,6 @@
           Ensure that your site's `repo` variable is defined in `group_vars/{{ env }}/wordpress_sites.yml` and uses the SSH format (example: git@github.com:roots/bedrock.git)
           More info:
           > https://roots.io/trellis/docs/deploys/
-      when: project.repo is not defined or not project.repo | match("^ssh://.+@.+|.+@.+:.+")
+      when: project.repo is not defined or project.repo is not match("^ssh://.+@.+|.+@.+:.+")
   roles:
     - deploy
--- a/lib/trellis/plugins/vars/version.py
+++ b/lib/trellis/plugins/vars/version.py
@@ -19,7 +19,7 @@ if version_info[0] > 2:
         'Please use Python 2.7.').format(version_info[0], version_info[1], version_info[2]))
 
 version_requirement = '2.4.0.0'
-version_tested_max = '2.5.1'
+version_tested_max = '2.5.2'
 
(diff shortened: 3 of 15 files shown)


runofthemill:
What do you think about making 2.5.1 the minimum requirement for the 1.0.0 milestone?

fullyint:
I added a commit to cover Ansible 2.5.2. Its release was [fast-tracked](https://groups.google.com/forum/#!topic/ansible-project/VE9m8JkeOKg) ([changelog](https://github.com/ansible/ansible/blob/stable-2.5/changelogs/CHANGELOG-v2.5.rst)). It works fine.

@runofthemill I suppose making 2.5.x the minimum requirement would avoid this PR's detection and warning about 2.5.0. I haven't thought through other benefits. I imagine you have some in mind.

I've become wary of bumping the minimum Ansible requirement unless absolutely necessary because of complicating things for users with multiple Trellis projects on multiple Ansible versions (examples: [one](https://discourse.roots.io/t/updated-to-ansible-2-4-deploys-broken-now-what/10588), [two](https://discourse.roots.io/t/trellis-no-longer-supports-ansible-2-3-1-0/10670)).

runofthemill:
@fullyint I've had that issue too, working on an older project that needed Ansible 2.3.x, and it was definitely a complication, so I see your reservations there. 

Honestly I think it's just a good opportunity to get ahead of some issues that would require a minor version release that anyone with multiple Trellis projects will have to deal with eventually. If Trellis is going to follow sem-ver once it hits 1.0.0 (see #634) in my mind it makes sense to deal with sooner rather than later, to be able to take advantage of all the work that's gone into improving functionality & bug fixes in Ansible.

On the other hand, it'd probably be easy to do down the line too as a minor release, so take the above as me just thinking aloud :)

fullyint:
I added a commit to cover Ansible 2.5.3. It works fine.

PR #989, opened 2018-05-18
Common: Install `git` instead of `git-core`

TangRufus:
Because `git-core` is now a dummy package of `git`.

See: http://git.661346.n2.nabble.com/git-core-vs-git-package-on-ubuntu-tp7576083p7576085.html


swalkinshaw:
TIL, thanks 🎉

PR #990, opened 2018-05-18
deploy.sh: Return non-zero exit code when misuse

TangRufus:
- Exit with `127` when not enough arguments
- Exit with `1` when hosts file not exist

See: http://www.tldp.org/LDP/abs/html/exitcodes.html

Use case: Avoid CI servers report false positive when deployment fail.


swalkinshaw:
Agreed! Thanks

PR #992, opened 2018-05-20
Ubuntu 18.04

swalkinshaw:
Adds support for the new Ubuntu 18.04 Bionic LTS release. This also *defaults* to Bionic in development (via the default Vagrant box).

This change is marked as `[BREAKING]`. Read below for more details.

Trellis will continue to work on both 16.04 and 18.04. If you want to update Trellis to the latest version but *continue* using 16.04 Xenial, you'll encounter two minor issues to deal with:

1. Revert the `vagrant_box` and `vagrant_box_version` options in `vagrant.default.yml` to the 16.04 values you previously had. Or better yet, create a `vagrant.local.yml` config to override those default values.
2. You'll see a warning that Trellis is now built for 18.04. You can safely ignore this error message if you want to stay on 16.04.

### DIFF
--- a/roles/mariadb/defaults/main.yml
+++ b/roles/mariadb/defaults/main.yml
@@ -1,6 +1,6 @@
-mariadb_keyserver: keyserver.ubuntu.com
+mariadb_keyserver: "hkp://keyserver.ubuntu.com:80"
 mariadb_keyserver_id: "0xF1656F24C74CD1D8"
-mariadb_ppa: "deb [arch=amd64,i386,ppc64el] http://ftp.osuosl.org/pub/mariadb/repo/10.2/ubuntu xenial main"
+mariadb_ppa: "deb [arch=amd64] http://nyc2.mirrors.digitalocean.com/mariadb/repo/10.2/ubuntu {{ ansible_distribution_release }} main"
 
 mariadb_client_package: mariadb-client
 mariadb_server_package: mariadb-server
--- a/vagrant.default.yml
+++ b/vagrant.default.yml
@@ -2,8 +2,8 @@
 vagrant_ip: '192.168.50.5'
 vagrant_cpus: 1
 vagrant_memory: 1024 # in MB
-vagrant_box: 'bento/ubuntu-16.04'
-vagrant_box_version: '>= 201801.02.0'
+vagrant_box: 'bento/ubuntu-18.04'
+vagrant_box_version: '>= 201807.12.0'
 vagrant_ansible_version: '2.5.3'
 vagrant_skip_galaxy: false
 vagrant_mount_type: 'nfs'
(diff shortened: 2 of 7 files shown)


roles/mariadb/defaults/main.yml
```diff
@@ -1,6 +1,6 @@
-mariadb_keyserver: keyserver.ubuntu.com
+mariadb_keyserver: "hkp://keyserver.ubuntu.com:80"
```
TangRufus: Why not `hkps://`?

swalkinshaw: Just copied from Mariadb site but updated to `hkps` and it works 👍

swalkinshaw:
I'm not sure there's a good seamless upgrade path here. In the past I think we just merged changes like this and people worked around it? The validation definitely makes it harder to deal with though.

If someone is running 16.04 on a server and they pull the latest Trellis updates, they can't really use Trellis at all without modifications or upgrading to 18.04.

@fullyint any thoughts?

swalkinshaw:
Turns out that Python 3 is the default on 18.04 which causes problems with Trellis/Ansible right now. Default Trellis behaviour is to let Vagrant install Ansible on the VM (this also applies to Windows usage).

1. we have validation that prevents Trellis from running on python3.
2. things may actually be broken until Ansible 2.5+ is required.

vagrant.default.yml
```diff
@@ -2,7 +2,7 @@
 vagrant_ip: '192.168.50.5'
 vagrant_cpus: 1
 vagrant_memory: 1024 # in MB
-vagrant_box: 'bento/ubuntu-16.04'
+vagrant_box: 'bento/ubuntu-18.04'
 vagrant_box_version: '>= 201801.02.0'
```
TangRufus: Looks like

swalkinshaw:
Done some updates to this. It should now work with either 16.04 (Xenial) or 18.04 (Bionic) and either Python 2 or 3 so it's quite flexible.

There's only two issues for people wanting to continue with 16.04:

1. they'll need to update the Vagrant config to switch back to the 16.04 box
2. they'll see a warning (but not error) that Trellis is built for 18.04 now

partounian:
Maybe we should add a little command to run dist upgrades?

swalkinshaw:
@partounian wouldn't this mostly just be running `apt-get upgrade`? This seems like something that should be run manually because it may need user intervention, or at least for the user to read the output before continuing. Usually for apt installs Trellis just silently agrees and continues on but this could be a dangerous operation.

partounian:
@swalkinshaw

```bash
sudo apt update 
sudo apt upgrade
sudo apt dist-upgrade
sudo apt autoremove
sudo apt install update-manager-core
sudo do-release-upgrade
```

swalkinshaw:
Ah thanks. But still I'd worry about trying to run all those tasks unattended.

vagrant.default.yml
swalkinshaw: I've updated it to the latest 👍

roles/mariadb/defaults/main.yml
swalkinshaw: I switched this back to non-hkps because 16.04 doesn't support it. I'm fine with keeping this "non-secure" version since the official MariaDB instructions use this as well. It makes it way easier for people to stay on 16.04 for now.

PR #1002, opened 2018-06-14
Skip create web root when using mounted_volume

biinari:
When `mounted-volume` playbook is used, let that playbook handle the creation of web root


swalkinshaw:
Can you give some background on this? I'm not exactly sure what it is or why it's needed.

biinari:
Oops sorry, that was just meant for our fork. Doesn't need to be pushed upstream.

PR #1003, opened 2018-06-14
Configurable Vagrant mount type

biinari:
In some circumstances it is not possible to use an NFS (e.g. in combination
with ecryptfs).

Allow user to configure vagrant synced folder mount type in vagrant.local.yml like any of:

```yml
vagrant_mount_type: 'nfs'
vagrant_mount_type: 'rsync'
vagrant_mount_type: 'smb'
vagrant_mount_type:
```

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -69,13 +69,16 @@ Vagrant.configure('2') do |config|
 
   bin_path = File.join(ANSIBLE_PATH_ON_VM, 'bin')
 
-  if Vagrant::Util::Platform.wsl? || (Vagrant::Util::Platform.windows? and !Vagrant.has_plugin? 'vagrant-winnfsd')
+  vagrant_mount_type = vconfig.fetch('vagrant_mount_type')
+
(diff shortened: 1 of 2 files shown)


Vagrantfile
```diff
@@ -69,26 +69,27 @@ Vagrant.configure('2') do |config|
 
   bin_path = File.join(ANSIBLE_PATH_ON_VM, 'bin')
 
-  if Vagrant::Util::Platform.wsl? || (Vagrant::Util::Platform.windows? and !Vagrant.has_plugin? 'vagrant-winnfsd')
+  vagrant_mount_type = vconfig.fetch('vagrant_mount_type')
+
+  if vagrant_mount_type != 'nfs' || Vagrant::Util::Platform.wsl? || (Vagrant::Util::Platform.windows? && !Vagrant.has_plugin?('vagrant-winnfsd'))
+    vagrant_mount_type = nil if vagrant_mount_type == 'nfs'
```
swalkinshaw: `nfs` is still a valid `type`. Why not just set it to that instead of this?

biinari: @swalkinshaw this needs to be set to `nil` when on a Windows platform without support for nfs to allow the vagrant driver to choose type automatically.

swalkinshaw:
Looks good, thanks!

swalkinshaw:
Looks good, thanks!

PR #1007, opened 2018-06-25
Update xdebug tunnel configuration

nathanielks:
This updates ssh connections to allow tcp forwarding when a tunnel is active, otherwise the tunnel won't work. I also updated the ip address used to form the tunnel.


swalkinshaw:
👍

Issue #1011, opened 2018-07-09
 sorry, tersmitten.swapfile was not found on https://galaxy.ansible.com.


matgargano:
n.b. a PR fixing this is incoming

When setting up Trellis, specifically running `ansible-galaxy install -r requirements.yml` I am receiving:

```
- downloading role 'swapfile', owned by tersmitten
 [WARNING]: - swapfile was NOT installed successfully: - sorry,
tersmitten.swapfile was not found on https://galaxy.ansible.com.
```

a quick google led me to https://github.com/Oefenweb/ansible-swapfile/issues/11 where it seems that tersmitten changed the owner of the repo (or his username in github) to Oefenweb.

Changing the requirements from `src: tersmitten.swapfile` to `  src: oefenweb.swapfile` appears to work

roots-ladybug:
Hi @matgargano,

It looks like the issue template is missing from this issue. Please take a look at the [Contribution Guidelines](https://github.com/roots/guidelines/blob/master/CONTRIBUTING.md), which will tell you **exactly** what your ticket has to contain in order to be processable.

Please **do not** use the issue tracker for personal support requests (use the [Roots Discourse](https://discourse.roots.io/) to ask the Roots Community for help, or if you want the Roots Team to dedicate some time to your issue, we [offer our services](https://roots.io/services/) as well).

PR #1012, opened 2018-07-09
#1011 change name of swapfile ansible requirement to match updated na…

matgargano:
…me on github

see https://github.com/roots/trellis/issues/1011


swalkinshaw:
Thanks!

PR #1015, opened 2018-08-17
Ability to deploy using git submodules

jawngee:
I know you guys aren't fans of submodules, but some of us have the need and it was a tiny change to enable.

### DIFF
--- a/group_vars/all/main.yml
+++ b/group_vars/all/main.yml
@@ -7,6 +7,7 @@ apt_security_package_state: latest
 apt_dev_package_state: latest
 ntp_timezone: Etc/UTC
 ntp_manage_config: true
+deploy_using_git_archive: false
 www_root: /srv/www
 ip_whitelist:
   - "{{ (env == 'development') | ternary(ansible_default_ipv4.gateway, ipify_public_ip | default('')) }}"
--- a/group_vars/all/main.yml
+++ b/group_vars/all/main.yml
@@ -7,10 +7,13 @@ apt_security_package_state: latest
 apt_dev_package_state: latest
 ntp_timezone: Etc/UTC
 ntp_manage_config: true
-deploy_using_git_archive: false
 www_root: /srv/www
 ip_whitelist:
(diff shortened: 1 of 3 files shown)


group_vars/all/main.yml
```diff
@@ -11,6 +11,10 @@ www_root: /srv/www
 ip_whitelist:
   - "{{ (env == 'development') | ternary(ansible_default_ipv4.gateway, ipify_public_ip | default('')) }}"
 
+# Deploy using git archive to bundle up the repo, if using git submodules, set this to false
+deploy_using_git_archive: true
```
swalkinshaw: I'd prefer a `deploy_method` variable with two options: `archive` and `copy` (or a better name).

roles/deploy/tasks/prepare.yml
```diff
@@ -26,17 +26,28 @@
...
+- name: Copy files to new build dir
+  shell: cp -pr {{ project_source_path }}/* {{ deploy_helper.new_release_path }}
+  when: deploy_using_git_archive==false and project.repo_subtree_path is not defined
+
+- name: Remove unwanted files/folders from new release
+  shell: find {{ deploy_helper.new_release_path }} -name '{{item}}' | xargs rm -rf
+  with_items: "{{ project.project_unwanted_items | default(project_unwanted_items) }}"
+  become: true
+  become_user: "{{ web_user }}"
+  when: deploy_using_git_archive==false
```
swalkinshaw: It would be nicer to use `block` to separate the two methods, or even put them in their own files and use `include` based on the method variable.

roles/deploy/tasks/prepare.yml
```diff
@@ -26,17 +26,28 @@
     path: "{{ deploy_helper.new_release_path }}"
     state: directory
 
+- name: Copy files to new build dir
+  shell: cp -pr {{ project_source_path }}/* {{ deploy_helper.new_release_path }}
```
swalkinshaw: This doesn't properly support `repo_subtree_path` does it?

PR #1021, opened 2018-09-30
Allow customizing Nginx `worker_connections`

TangRufus:
> If you need more connections than this[8000], you start optimizing your OS. That's probably the point at which you hire people who are smarter than you as this is *a lot* of requests.
> Should be < worker_rlimit_nofile.
>
> https://github.com/h5bp/server-configs-nginx/blob/c5c6602232e0976d9e69d69874aa84d2a2698265/nginx.conf#L19-L25

Nginx Default `worker_connections`: 512
Trellis / h5bp Default `worker_connections`: 8000
Trellis / h5bp Default `worker_rlimit_nofile`: 8192


swalkinshaw:
Thanks 👍

PR #1028, opened 2018-11-03
#1027 make output_buffering optional

matgargano:
If you want to use WordFence's 2FA with Google Authenticator and it to prompt you for a token and not use the silly scheme "{password} wf{token}" as your password, WordFence requires output buffering. see https://www.wordfence.com/help/tools/two-factor-authentication/ 

I believe this will allow you to automatically turn output buffering on if you add a `php.yml` to your target environment's `group_vars`

### DIFF
--- a/group_vars/staging/php.yml
+++ b/group_vars/staging/php.yml
@@ -1,0 +1,9 @@
+php_error_reporting: 'E_ALL'
+php_display_errors: 'On'
+php_display_startup_errors: 'On'
+php_track_errors: 'On'
+php_mysqlnd_collect_memory_statistics: 'On'
+php_opcache_enable: 0
+
+xdebug_remote_enable: 1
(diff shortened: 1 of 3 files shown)


group_vars/staging/php.yml
```diff
@@ -0,0 +1,9 @@
+php_error_reporting: 'E_ALL'
```
swalkinshaw: There's no need for this file right? I'd prefer it removed from this PR.

swalkinshaw:
I don't know about your specific use case, but I'm fine to add this variable as an option 👍

matgargano:
@swalkinshaw the php.yml was copypasta, i've removed it

swalkinshaw:
Thank you!

matgargano:
@swalkinshaw thank you for maintaining this project!

PR #1029, opened 2018-11-14
Update README.md

barryels:
For first time users it might be useful to be explicit about which directory they should be in before running `vagrant up`, else Vagrant will display an error (`A Vagrant environment or target machine is required to run this command.`) when the `Vagrantfile` is not found.


swalkinshaw:
Thanks, yeah I didn't see this explicitly mentioned so it's useful.

PR #997, opened 2018-06-06
build-before: Checkout project source code to local temporary directory

TangRufus:
This pull request clones project files according to `wordpress_sites.yml` configuration(`repo`, `branch`, etc) into a local temporary directory during `build-before`.

Build directory is kept(not deleted) after deploy to speed up subsequent deploys(as `git` and `yarn` caches). By default, build directory is located inside `$TMPDIR/trellis` or `/tmp/trellis`. This is to allow the OS to clean them up automatically (after [reboot](https://askubuntu.com/questions/20783/how-is-the-tmp-directory-cleaned-up/857154#857154) or [once a few days](https://superuser.com/questions/187071/in-macos-how-often-is-tmp-deleted#comment1920236_187105)). After build directory cleanup, the first [`git clone`](https://github.com/TangRufus/trellis/blob/ca2c70646697d0379a7da2d483e00038b40e3849/deploy-hooks/build-before.yml#L16-L23) and [`yarn install`](https://github.com/TangRufus/trellis/blob/ca2c70646697d0379a7da2d483e00038b40e3849/deploy-hooks/build-before.yml#L25-L29) become slow. If you want to avoid build directory cleanup, set `project_build_path` to something else (but not under any `vagrant_synced_folders` to avoid unnecessary file sync between your machine and the vagrant VM).

Tips: 
- [Cache `$TMPDIR/trellis` or `/tmp/trellis` on CI servers](https://github.com/TypistTech/tiller-circleci/blob/f135607a38c81c9ef4e61817f54568fd1f20dc1a/config.yml#L37-L40)
- Change `project_build_path` if you have multiple bedrock projects using the same site key

---

**What is the current behavior?**

`deploy-hooks/build-before.yml` uses whatever project files present locally, ignore the configuration(`repo`, `branch`, etc) in `group_vars/<env>/wordpress_sites.yml`

To ensure deploy works as expected, the developer needs to:
1. `$ cd /path/to/local/bedrock`
1. `$ git checkout THE_BRANCH`
1. Confirm the branch is up to date with 'origin/THE_BRANCH', nothing to commit, working tree clean: `$ git fetch && git status`
1. Actually deploy

These steps are undocumented. This pull request removes the needs of steps 1~3.

**Is there a related [Discourse](https://discourse.roots.io/) thread or were any utilized (please link them)?**

https://discourse.roots.io/t/locally-compiled-sage-assets-on-trellis-deploy/5911

PR #1030, opened 2018-11-18
Extract Sage 9 compiler

TangRufus:
Second attempt of #997 plus compile multiple sage themes

---

#### What's wrong with #997?

Technically nothing's wrong. However, changing from `project_local_path` to `project_build_path` confuses developers (https://github.com/roots/trellis/pull/997#issuecomment-397906560, https://github.com/roots/trellis/pull/997#issuecomment-398448690). Using a different variable name is not enough to show the differences between those variables:
- `project_local_path`: the local bedrock path, e.g: `/path/to/trellis/../site/example1`, the one that you developing on
- `project_build_path`: the temporary build path, under `$TMPDIR/trellis` or `/tmp/trellis`, not respecting `project.repo_subtree_path`

Thus, this attempt removes the needs of changing `deploy-hooks/build-before.yml` and requires develoeprs to add their sage theme names in `group_vars/<env>/wordpress_sites.yml`:

```diff
  # group_vars/<env>/wordpress_sites.yml
  wordpress_sites:
    example.com:
+     sage_themes:
+       - my-first-sage-theme-name
+       - my-second-sage-theme-name
```

#### Why multiple sage theme? Or, In what cases it will be helpful?

- Solution to https://discourse.roots.io/t/ansible-get-wrong-theme-directory-for-second-site-in-trellis/6408
- Allow multsite network to use multiple themes

#### Testing needed!!

We have been using it on production for a while with file structure described here in https://github.com/roots/trellis/issues/883#issuecomment-329052189

Never tested on [*the roots-offical-docs way*](https://roots.io/trellis/docs/installing-trellis/#create-a-project).

#### Discussion: Ansible Galaxy Role?

Since `deploy-hooks/build-before.yml` is useless when you don't use sage and support for Laravel projects is planned https://github.com/roots/trellis/pull/951, would it make sense if we extract `build-before.yml` into a separate ansible galaxy role?

---

cc: @codepuncher, @BufferOverNo, @JackAlexander1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+### HEAD
+* Extract Sage 9 compiler ([#1030](https://github.com/roots/trellis/pull/1030))
+
 ### 1.0.1: January 16th, 2019
 * Add Python 2 explicitly ([#1061](https://github.com/roots/trellis/pull/1061))
 
--- a/deploy-hooks/build-before.yml
+++ b/deploy-hooks/build-before.yml
@@ -1,10 +1,14 @@
-# Placeholder `deploy_build_before` hook for building theme assets on the
+# Placeholder `deploy_build_before` hook for building frontend assets on the
 # host machine and then copying the files to the remote server
 #
-# ⚠️ This example assumes your theme is using Sage 9
-# An example for themes built with Sage 8 can be found at: https://git.io/vdgUt
+# ⚠️ The following is an simplified example of:
+#  1. installing npm dependencies
+#  2. compiling assets for production
+#  3. uploading assests to remote servers
 #
-# Uncomment the lines below and replace `sage` with your theme folder
+# An role for themes built with Sage 9 can be found at: https://github.com/ItinerisLtd/trellis-sage-9-compiler
+# An example for themes built with Sage 8 can be found at: https://git.io/vdgUt
+# More user contributed extensions can be found at: https://roots.io/trellis/docs/user-contributed-extensions/
 #
 # ---
 # - name: Install npm dependencies
@@ -12,11 +16,6 @@
 #   delegate_to: localhost
 #   args:
 #     chdir: "{{ project_local_path }}/web/app/themes/sage"
-#
-# - name: Install Composer dependencies
-#   command: composer install --no-ansi --no-dev --no-interaction --no-progress --optimize-autoloader --no-scripts
-#   args:
-#     chdir: "{{ deploy_helper.new_release_path }}/web/app/themes/sage"
 #
 # - name: Compile assets for production
 #   command: yarn build:production


TangRufus:
> I'm not sure how generic we can make this. Composer + Yarn install are common/generic. yarn build:production is more Sage specific, and then especially copying the dist folder.

Instead of making *compile steps (composer, yarn, copy, etc)* generic, I suggest normalizing the way we complie frontend assets in deploy hooks (similar idea to #896, allowing thirdy party roles to extend trellis).

For example, this is what developers would do if they using both sage 8 and sage 9 themes on the same server.

```diff
  # requirements.yml

+ - src: XXX.trellis-sage-8-compiler
+   version: 1.2.3

+ - src: YYY.trellis-sage-9-compiler
+   version: 1.2.3
```

```diff
  # group_vars/all/deploy-hooks.yml

+ deploy_build_before:
+   - "{{ playbook_dir }}/vendor/roles/XXX.trellis-sage-8-compiler/tasks/main.yml"
+   - "{{ playbook_dir }}/vendor/roles/YYY.trellis-sage-9-compiler/tasks/main.yml"
```

```diff
  # group_vars/<env>/wordpress_sites.yml

  wordpress_sites:
    example.com:
+     sage_8_compiler:
+       - my-sage-eight-theme
    another-example.com:
+     sage_9_compiler:
+       - my-sage-nine-theme
    multisite-example.com:
+     sage_9_compiler:
+       - my-sage-nine-theme
+       - my-second-sage-nine-theme
+     sage_8_compiler:
+       - my-sage-eight-theme
+       - my-second-sage-eight-theme
```

*same idea goes to Laravel mix*

---

> As is right now, I don't like having sage_themes leaked into wordpress_sites.yml by default (even if it's just a comment).

How about this:
- extract `build-before.yml` into a separate ansible galaxy role
- remove `deploy-hooks/build-before.yml` from trellis
- remove the comments from `wordpress_sites.yml`
- galaxy role documents instruct developers to define `sage_themes` (or `sage_8_compiler` / `sage_9_compiler`) in `wordpress_sites.yml`

swalkinshaw:
Sounds like a good plan 👍 

Only question is if it makes sense for the wp sites property `sage_themes` or whatever to be more generic? Maybe just `theme_builder` or something like that?

swalkinshaw:
Yeah that makes sense. Go for it 👍

TangRufus:
**Help wanted**

Although I extracted to [trellis-sage-9-compiler](https://github.com/ItinerisLtd/trellis-sage-9-compiler), there are some more tests and todos needed.

See: https://github.com/ItinerisLtd/trellis-sage-9-compiler#todos-help-wanted

---

Question:
- Should I completely remove sage from trellis, i.e: remove those comments from `wordpress_sites.yml` and remove the role from `requirements.yml`?

swalkinshaw:
Yeah I think we should remove all mentions of it, except for maybe an updated default `deploy-hooks/build-before.yml`?

Maybe `deploy-hooks/build-before.yml` could include a more generic simplified example with a `npm install && `npm build` type thing. It could also mention the sage compiler and link to docs?

Either way, we'd need to update our docs to include a page on Trellis + Sage I think. This would deserve more than an entry in the user contributed extensions page.

@retlehs what do you think?

TangRufus:
> we should remove all mentions of it,
> `deploy-hooks/build-before.yml` could include a more generic simplified example with a `npm install && `npm build` type thing.

Done.

swalkinshaw:
@retlehs had some thoughts before on this. Can you explain them?

retlehs:
i would prefer to close this PR entirely and just add the sage 9 compiler to the user contributed roles docs for if they want to use it...

we can't make it harder/more confusing to use sage out of the box with trellis

swalkinshaw:
Process now:
* uncomment the example deploy hook

Process after:
* add new galaxy role (+ install)
* define `deploy_build_before`
* update `wordpress_sites`

Is that correct? Obviously it's more work, but it's done in a more extensible way.

retlehs:
if we're going to do this, there needs to be a theme agnostic ansible role that we can include in trellis by default

there shouldn't be anything specific to sage 8, sage 9, or even a theme - since plugins should be allowed to take advantage of this too

ideally there's some simple way to configure this in trellis (maybe more inline with how a CI config looks?), but i don't have a better idea of what that could be right now

PR #1031, opened 2018-11-25
Python3 support

swalkinshaw:
This makes Trellis compatible with both Python 2 and 3. To use Python 3 though, Ansible >= 2.5 is required. However, since we already tell people to avoid `2.5.0`, let's just say use `2.5.3` at minimum.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Python 3 support ([#1031](https://github.com/roots/trellis/pull/1031))
 * Allow customizing Nginx `worker_connections` ([#1021](https://github.com/roots/trellis/pull/1021))
 * Update wp-cli to 2.0.1 ([#1019](https://github.com/roots/trellis/pull/1019))
 * [BREAKING] Update wp-cli to 2.0.0 and verify its PGP signature ([#1014](https://github.com/roots/trellis/pull/1014))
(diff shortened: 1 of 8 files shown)


fullyint:
I haven't reviewed yet. There may be some ideas in the old [python-3-compat](https://github.com/roots/trellis/tree/python-3-compat) branch, e.g., in 9e83883

Here are some hastily pasted related notes I had from months ago...
- Users can override the dynamic determination of interpreter by simply defining somewhere `ansible_python_interpreter` var (e.g., per host, or more globally in `group_vars`). See https://github.com/ansible/ansible/issues/21311
- httplib (letsencrypt test acme challenges) taken from https://github.com/ansible/ansible/blob/e4efc191c5be078ead687553538f92d035946f1f/lib/ansible/module_utils/urls.py#L45-L49

I'll try to make time to review my old PY3 to-do notes and add them here.

swalkinshaw:
Oh thanks, either I never knew about or forgot 😳 

Looks like those changes are basically split 50/50 between Python on local and remote. This PR is definitely only for Python 3 on local machine.

Most of the changes are similar in concept, but my branch didn't seem to need a lot of them? Keep in mind I didn't test deploys yet.

swalkinshaw:
Going to merge this as is. We'll finish Ubuntu 18.04 separately this is useful on its own.

@fullyint feel free to post any other notes you may have.

PR #1038, opened 2018-12-12
Revert "Ubuntu 18.04"

retlehs:
Reverts roots/trellis#992

issues with those running python 2.x locally right now


designst:
I just replaced the apt-package `python-software-properties` with `software-properties-common` and it worked. So i don't think that you have to revert the whole changes.

swalkinshaw:
Closing since https://github.com/roots/trellis/pull/1039 is the fix.

PR #1043, opened 2018-12-13
Create wp-cli.local.yml file with @vagrant alias for every site

retlehs:
props @QWp6t 

ref #1042


swalkinshaw:
Should this be done via Ansible? 🤔

TangRufus:
Note: https://github.com/roots/trellis-cli/pull/52 is not the same as this one. 

https://github.com/roots/trellis-cli/pull/52 doesn’t support vagrant/valet environments

swalkinshaw:
Oh oops, thanks. This PR was WIP anyway so if its picked back up we can re-open.

PR #1044, opened 2018-12-14
Don't force install Ansible Galaxy in dev

swalkinshaw:
By default Vagrant runs `ansible-galaxy` with the `--force` option causing every Galaxy role to be re-downloaded *every* time. This is annoying, slow, and wasteful.

This customizes the `galaxy_command` to run without the `--force` flag.

Borrowed from this great SO answer: https://stackoverflow.com/questions/46045192/how-to-automatically-install-ansible-galaxy-roles-using-vagrant

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -119,6 +119,7 @@ Vagrant.configure('2') do |config|
     ansible.playbook = File.join(provisioning_path, 'dev.yml')
     ansible.galaxy_role_file = File.join(provisioning_path, 'requirements.yml') unless vconfig.fetch('vagrant_skip_galaxy') || ENV['SKIP_GALAXY']
     ansible.galaxy_roles_path = File.join(provisioning_path, 'vendor/roles')
+    ansible.galaxy_command = 'ansible-galaxy install --role-file=%{role_file} --roles-path=%{roles_path}'
 
     ansible.groups = {
       'web' => ['default'],


TangRufus:
`ansible-galaxy` won't install/upgrade existing roles without `--force`.

Example:
1. fork and use Trellis as ususal
1. add `my-role v1` to `requirements.yml`
1. `$ vagrant up`
1. `my-role v1` is installed
1. bump `my-role` to v2 in `requirements.yml`
1. `$ vagrant up`
1. `my-role` won't be upgraded and stays v1 because of missing `--force` flag

Should we warn users about this?

swalkinshaw:
Oh I did mean to mention that. And it's the reason why I didn't merge this because it's a trade off.

Maybe my experience with the default is so much more annoying because I constantly create/destroy VMs for testing whereas most users wouldn't do it as often?

TangRufus:
> Maybe my experience with the default is so much more annoying because I constantly create/destroy VMs for testing whereas most users wouldn't do it as often?

Agree it is annoying. Somethimes I just copy and paste the ``vendor` folder from another project. And, always set `vagrant_skip_galaxy` to `true`.

Perphas a warning in the docs is enough?
Reasons:
- it doesn't make any differences to first time users
- whenever `requirements.yml` is changed, it's normal/expected to run `$ anisble-galaxy` manually

swalkinshaw:
>  it's normal/expected to run $ anisble-galaxy manually

That would match up with every package manager as well. If you're using npm, composer, etc, everyone is used to having to run them manually to pull in latest versions.

connerbw:
Just got bitten by this today....

https://roots.io/trellis/docs/windows/

> If your host machine is running Windows, the workaround is to run Ansible on the VM (since it's running Ubuntu) and not locally. You do not need to install Ansible manually. When you run vagrant up, the Vagrantfile will detect Windows and run the ansible_local provisioner. This provisioner installs Ansible and the external Ansible roles/packages on the VM **(so you can also skip manually running the ansible-galaxy install command).**

Per the docs. I don't have ansible installed Running anisble-galaxy, for me, in is non trivial.

>If you're using npm, composer, etc, everyone is used to having to run them manually to pull in latest versions.

It's not the same? I think I ran ansible once in the last 3 months? I didn't even run it, it's just a side effect of `vagrant --provision`. I'm trying to understand the use case where this being slow matters. How many times is a dev supposed to run `vagrant up --provision`? And if this kind of slowness matters to them, then  why couldn't they switch `vagrant_skip_galaxy: true` in [vagrant.default.yml](https://github.com/pressbooks/trellis/blob/master/vagrant.default.yml)  ?

swalkinshaw:
@connerbw fair points. To be honest, I forgot about the `vagrant_skip_galaxy` option 😳.

I think overriding that in `.vagrant.local.yml` is probably the better option here. Think I'll revert this.

PR #1045, opened 2018-12-14
Ansible 2.7 compatibility

### LINKED ISSUE #1041: Upgrade vagrant_ansible_version?
## Submit a feature request

- [x] This is a feature request
- [ ] This is a bug report

---

Running Ansible 2.7.4 I always get the following warning:
```
 [WARNING]: You Ansible version is 2.7.4 but this version of Trellis has only
been tested for compatability with Ansible 2.4.0.0 -> 2.5.3. It is advisable to
check for Trellis updates or downgrade your Ansible version.
```
Besides, there's also a typo: "You" should be "Your".
Otherwise I see no problems, I only get one additional warning:
```
TASK [wordpress-setup : Warn about deprecated Nginx includes directory] ********
[DEPRECATION WARNING]: Using tests as filters is deprecated. Instead of using 
`result|search` use `result is search`. This feature will be removed in version
 2.9. Deprecation warnings can be disabled by setting 
deprecation_warnings=False in ansible.cfg.
```

Maybe this warning could be removed by upgrading `vagrant_ansible_version`. I don't know if anyone else has tested, but I definitely see no problems here with the current Ansible version.

retlehs:
updated the connection role to account for this change in ansible versions:

2.5: `[0;32m            "\u001b[0;32mroot", [0m`
2.7: `[0;32m            "\u001b[0;33mroot", [0m`

props @swalkinshaw 

close #1041

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Bump Ansible `version_tested_max` to 2.7.5 ([#1045](https://github.com/roots/trellis/pull/1045))
 * Add Vagrant `ssh-config` to `~/.ssh/config` on `vagrant up` ([#1042](https://github.com/roots/trellis/pull/1042))
 * [BREAKING] Add Ubuntu 18.04 support and default to it ([#992](https://github.com/roots/trellis/pull/992))
 * Python 3 support ([#1031](https://github.com/roots/trellis/pull/1031))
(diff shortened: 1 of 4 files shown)


swalkinshaw:
The difference in output is due to a different colour used by success or changed status. We have `changed_when: false` on the task but maybe that isn't making it return "success" anymore.

So in 2.6 or 2.7, the task now reports as changed which changes the colour output.

We should ideally try and find a less brittle way of determining if `root` can connect if possible although I know @fullyint did a lot of work on this so maybe there isn't really.

PR #1047, opened 2018-12-14
Allow overriding rollback variables

TangRufus:
This pull request makes `rollback.yml` respects `project_root` and `project_current_path` in `group_vars/<env>/xxx.yml` like `deploy.yml`.

Use case: Rollback on Kinsta.


TangRufus:
cc: @codepuncher

swalkinshaw:
Much better, thanks 👏

PR #1048, opened 2018-12-16
Add `vault_wordpress_env_defaults`

TangRufus:
`vault_wordpress_env_defaults` works the same way as `wordpress_env_defaults` but expecting to be encrypted by anisble-vault.

Use case:
- non-environment-specific plugin license keys


swalkinshaw:
Seems useful, thanks 👍

PR #1053, opened 2018-12-20
Use Ruby for ssh-config trigger

swalkinshaw:
This replaces the bash implementation in https://github.com/roots/trellis/pull/1042 with a Ruby one instead.

It has some improvements and safety guards. Most important of all, it creates a backup of the existing config file before doing anything.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -137,10 +137,10 @@ Vagrant.configure('2') do |config|
     if !Vagrant::Util::Platform.windows?
       config.trigger.after :up do |trigger|
         # Add Vagrant ssh-config to ~/.ssh/config
-        trigger.run = {
-          path: File.join(ANSIBLE_PATH, 'bin/ssh-vagrant-config.sh'),
-          args: [main_hostname]
-        }
+        trigger.info = "Adding vagrant ssh-config for #{main_hostname } to ~/.ssh/config"
+        trigger.ruby do
+          update_ssh_config(main_hostname)
+        end
       end
     end
   end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Use Ruby script for ssh-config trigger ([#1053](https://github.com/roots/trellis/pull/1053))
 * Update to PHP 7.3 ([#1052](https://github.com/roots/trellis/pull/1052))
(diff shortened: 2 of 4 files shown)


Vagrantfile
```diff
@@ -136,10 +136,10 @@ Vagrant.configure('2') do |config|
     if !Vagrant::Util::Platform.windows?
```
QWp6t: Shouldn't this be enabled on Windows now that it no longer relies on bash? 🤔

Vagrantfile
```diff
@@ -136,10 +136,10 @@ Vagrant.configure('2') do |config|
...
       config.trigger.after :up do |trigger|
         # Add Vagrant ssh-config to ~/.ssh/config
-        trigger.run = {
-          path: File.join(ANSIBLE_PATH, 'bin/ssh-vagrant-config.sh'),
-          args: [main_hostname]
-        }
+        trigger.info = "Adding vagrant ssh-config for #{main_hostname } to ~/.ssh/config"
+        trigger.ruby do |_, _|
```
TangRufus: Can we omit the block parameters?

```diff
- trigger.ruby do |_, _|
+ trigger.ruby do
```

Vagrantfile
swalkinshaw: `~/.ssh/config` is a safe assumption on Windows as well?

QWp6t: `%userprofile%/.ssh/config` in windows parlance, where `userprofile` is an environment variable to the user's home directory. while not all versions of windows come with openssh (windows 10 comes with it since earlier this year), git requires ssh, and so most windows users who are using trellis will probably have openssh and this is a safe assumption in those cases.

swalkinshaw: And if that file exists, we can safely assume it's the same format I'm guessing?

QWp6t: Yeah, it follows the same format, but one thing that would concern is me is whether the path to the private key is in the windows (e.g., `d:\Sites\example.com\trellis\etc...`) or unix (e.g., `/mnt/d/Sites/example.com/trellis/etc...`) format. There's probably not an easy way to handle this, but I'm thinking we could rely on Ruby/Vagrant to figure it out.

My Ruby is terrible, but I can test whatever you wanna send me.

PR #1054, opened 2018-12-21
Move remote DB check to playbook level

swalkinshaw:
We should skip the entire `mariadb` role if no local DB is needed.

This is simpler and also more consistent with the `letsencrypt` role.


swalkinshaw:
🤔 trying out moving the client portion out of here so the role can be conditional. It makes it simpler to install `mysql-client` vs `mariadb-client` since that can be done without using the MariaDB PPA.

`mariadb-client` ends up installing `mysql-common` as well anyway. And their official site says:

> The mysql-client package (up to MySQL 5.7) also works with MariaDB server.
> All client APIs, protocols and structs are identical.

So I'm not sure there's any downside to this?

PR #1057, opened 2019-01-04
Let's Encrypt compatibility with python3

ouun:
The python script is not executable with Trellis >1.0.0 and Ptyhon >3. As I'm not familiar with python, this urgently needs some more tests but fixed it for me.
Please see: https://discourse.roots.io/t/task-letsencrypt-generate-the-certificates-usr-bin-env-python-no-such-file-or-directory/14493

PR #1060, opened 2019-01-08
Revert #1059 and #1057

retlehs:


swalkinshaw:
More background on this here: https://www.google.com/url?q=https://discourse.roots.io/t/task-letsencrypt-generate-the-certificates-usr-bin-env-python-no-such-file-or-directory/14493/8?u%3Dswalkinshaw&sa=D&source=hangouts&ust=1547066974113000&usg=AFQjCNFxsKJnqT7RCtnq7wFfdIC3fwIy3g

tl;dr we're going to ensure that python2 is an explicit dependency for now

PR #1061, opened 2019-01-08
Add Python 2 explicitly

retlehs:
ref #1060 and https://discourse.roots.io/t/task-letsencrypt-generate-the-certificates-usr-bin-env-python-no-such-file-or-directory/14493/8

also snuck the unzip package on there :)

### DIFF
--- a/roles/common/defaults/main.yml
+++ b/roles/common/defaults/main.yml
@@ -26,6 +26,8 @@ _apt_packages_default:
   dbus: "{{ apt_package_state }}"
   git: "{{ apt_package_state }}"
   libnss-myhostname: "{{ apt_package_state }}"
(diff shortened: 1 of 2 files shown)


roles/common/defaults/main.yml
```diff
@@ -26,6 +26,8 @@ _apt_packages_default:
   dbus: "{{ apt_package_state }}"
   git: "{{ apt_package_state }}"
   libnss-myhostname: "{{ apt_package_state }}"
+  python2.7-minimal: "{{ apt_package_state }}"
```
swalkinshaw: Looks like we need to do the actual plain `python` package to get `/usr/bin/python` which is what Ansible needs.

PR #1062, opened 2019-01-13
Don't add HSTS header for self-signed

mockey:
HSTS doesn't make much sense when using self-signed certs. Also an annoying warning is shown for each request in Firefox console:
`Strict-Transport-Security: The connection to the site is untrustworthy, so the specified header was ignored.`
So I simply added a condition to the template.


swalkinshaw:
Thanks for this. I agree and I *think* this makes the most sense at first thought.

PR #1075, opened 2019-03-14
Prevent direct access for .blade.php files

Log1x:
In Trellis' current state, `.blade.php` files are viewable in plain-text by accessing them directly. Laravel avoids this problem by having views outside of the `public` directory, but unfortunately, that isn't a sane solution with WordPress.

Outside of Trellis, this also calls for an addition to the Sage docs to provide examples on preventing access with apache and/or nginx.

Credit to pkarjala from [this thread](https://discourse.roots.io/t/blade-templates-directly-accessible-via-web-browser-on-server/15059) for finding this.

[Real World Example](https://roots-example-project.com/app/themes/sage/resources/views/index.blade.php)

### DIFF
--- a/roles/wordpress-setup/templates/wordpress-site.conf.j2
+++ b/roles/wordpress-setup/templates/wordpress-site.conf.j2
@@ -122,8 +122,8 @@ server {
   }
   {% endblock %}
   
-  {% block blade_twig_php -%}
-  # Prevent Blade and Twig files from being accessed directly.
+  {% block blade_twig_templates -%}
+  # Prevent Blade and Twig templates from being accessed directly.
   location ~* \.(blade.php|twig)$ {
     deny all;
   }
--- a/roles/wordpress-setup/templates/wordpress-site.conf.j2
+++ b/roles/wordpress-setup/templates/wordpress-site.conf.j2
@@ -124,7 +124,7 @@ server {
   
   {% block blade_twig_templates -%}
   # Prevent Blade and Twig templates from being accessed directly.
-  location ~* \.(blade\.php)(/.*)?$ {
+  location ~* \.(blade\.php|twig)(/.*)?$ {
     return 401;
   }
   {% endblock %}
--- a/roles/wordpress-setup/templates/wordpress-site.conf.j2
+++ b/roles/wordpress-setup/templates/wordpress-site.conf.j2
@@ -124,8 +124,8 @@ server {
   
   {% block blade_twig_templates -%}
   # Prevent Blade and Twig templates from being accessed directly.
-  location ~* \.(blade\.php|twig)$ {
-    deny all;
+  location ~* \.(blade\.php)(/.*)?$ {
+    return 401;
   }
   {% endblock %}
 
--- a/roles/wordpress-setup/templates/wordpress-site.conf.j2
+++ b/roles/wordpress-setup/templates/wordpress-site.conf.j2
@@ -124,7 +124,7 @@ server {
   
   {% block blade_twig_templates -%}
   # Prevent Blade and Twig templates from being accessed directly.
-  location ~* \.(blade.php|twig)$ {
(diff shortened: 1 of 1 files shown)


Log1x:
> might as well block twig files, too

Done.

roles/wordpress-setup/templates/wordpress-site.conf.j2
```diff
@@ -121,6 +121,13 @@ server {
     deny all;
   }
   {% endblock %}
+  
+  {% block blade_twig_templates -%}
+  # Prevent Blade and Twig templates from being accessed directly.
+  location ~* \.(blade.php|twig)$ {
```
QWp6t: Shouldn't the `.` be escaped? `blade\.php`? Or does it not matter?

Log1x: You're right, it should've been. Accidentally overlooked it and didn't catch it during testing due to the period matching any character.

Log1x:
This should be ready to get merged assuming we don't want to add another default extension. I believe Blade and Twig as defaults should be sane.

roles/wordpress-setup/templates/wordpress-site.conf.j2
```diff
@@ -121,6 +121,13 @@ server {
     deny all;
   }
   {% endblock %}
+  
+  {% block blade_twig_templates -%}
+  # Prevent Blade and Twig templates from being accessed directly.
+  location ~* \.(blade\.php|twig)(/.*)?$ {
```
swalkinshaw: Why the `(/.*)`? This is going to match a lot more than just files ending in `.blade.php` or `.twig`

Log1x: see https://github.com/roots/docs/pull/180#issuecomment-474637910

Log1x:
Curious which error code we should settle on? Possibilities are `401`, `403,` or maybe even a `301` back to homepage since returning a 401/403 will return the default web server template instead of being handled by WordPress.

roles/wordpress-setup/templates/wordpress-site.conf.j2
swalkinshaw: 🤔 okay but I still don't understand that situation. Maybe @QWp6t can clear it up

Log1x: Meh, you're probably right in terms of it not being an issue in Trellis unless the end-user messed with nginx configuration. Do you just want to leave it as a safe-guard or revert it?

swalkinshaw: The twig case is the one that worries me slightly more since there's no `.php` in it as well. So `example.com/foo.twig/asdf` would match? It likely doesn't matter, but I don't want to break any URLs which might be valid.

Log1x: It would indeed match– but as far as WordPress permalinks go it'd be a non-issue as `.` is replaced with `-` when editing a slug.

QWp6t:
I would suggest returning whatever the Not Allowed code is. I think that's 403 iirc (I'm on my phone, too lazy to look it up).

roles/wordpress-setup/templates/wordpress-site.conf.j2
QWp6t: @swalkinshaw I don't think `(/.*)` is needed here.

https://roots-example-project.com/app/themes/sage/resources/views/index.blade.php/more-stuff-here

Seems to work fine.

I was only suggesting it for Apache because after testing on a few Apache servers, they all displayed the blade file.

Log1x:
Apologies for the delay, I had sworn I had pushed an update already. The regex is now reverted as well as returning a 403 (`deny all`) instead of a 401.

This should now be ready.

PR #1077, opened 2019-04-10
Use all canonical site hosts for Landrush TLD

knowler:
This is useful if you want to use WordPress’ [multisite domain mapping](https://wordpress.org/support/article/wordpress-multisite-domain-mapping/) locally and if you’ve specified additional canonical hosts which are not sub-domains of the main canonical host.

In other words it lets you rename your domains from `client.example.test` to `client.test` and have them work (since dnsmasq via Landrush will create a `/etc/resolver/` entry for the domain).

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -58,7 +58,7 @@ Vagrant.configure('2') do |config|
     config.hostmanager.aliases = hostnames + redirects
   elsif Vagrant.has_plugin?('landrush') && trellis_config.multisite_subdomains?
     config.landrush.enabled = true
-    config.landrush.tld = config.vm.hostname
+    config.landrush.tld = trellis_config.site_hosts_canonical
     hostnames.each { |host| config.landrush.host host, vconfig.fetch('vagrant_ip') }
   else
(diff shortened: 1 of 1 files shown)


Vagrantfile
```diff
@@ -58,7 +58,7 @@ Vagrant.configure('2') do |config|
     config.hostmanager.aliases = hostnames + redirects
   elsif Vagrant.has_plugin?('landrush') && trellis_config.multisite_subdomains?
     config.landrush.enabled = true
-    config.landrush.tld = config.vm.hostname
+    config.landrush.tld = trellis_config.site_hosts_canonical.select { |host| !host[/\.#{main_hostname}$/] }
```
swalkinshaw: Minor but you can use `reject` without the negative operator. Also `end_with`:

```ruby
trellis_config.site_hosts_canonical.reject { |host| host.end_with?(".#{main_hostname}") }
```

knowler: Awesome!

swalkinshaw:
Thanks 🎉

Issue #949, opened 2018-02-13
[SOLUTION] Failed to connect to api.ipify.org at port 443: [Errno 61] Connection refused


partounian:
## Submit a feature request or bug report

- [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/guidelines/blob/master/CONTRIBUTING.md)
- [ ] This is a feature request
- [x] This is a bug report
- [x] This request isn't a duplicate of an [existing issue](https://github.com/roots/trellis/issues)
- [x] I've read the [docs](https://roots.io/trellis/docs) and followed them (if applicable)
- [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/c/trellis) forums

We should make a note somewhere that if you receive the error
`Failed to connect to api.ipify.org at port 443: [Errno 61] Connection refused`
you need to disable **DNSCrypt** if you have that running.

jedisct1:
> you need to disable DNSCrypt if you have that running

Why?

partounian:
Because it doesn't work with DNSCrypt, ipify is supposed to check the client's IP, but they block connections through DNSCrypt, or at least AdGuard's DNSCrypt servers.

artifex404:
Maybe adding a nicer error message mentioning DNSCrypt?

I got an error message:

> Failed to connect to api.ipify.org at port 443: [Errno 8] nodename nor
> servname provided, or not known
> fatal: [101.37.64.228 -> localhost]: FAILED! => {"changed": false}

partounian:
@swalkinshaw Thoughts? We got 2 diff errors, but both are for using DNSCrypt, should we mention this in a troubleshooting section of some kind?

swalkinshaw:
I'm open to any docs improvements around this. Not sure where specifically it should live though.

artifex404:
Maybe a custom error message is enough if Ansible is unable to get external IP?

`- fail:
    msg: "External IP resolution failed. Check that your DNS servers are working. Try to disable DNSCrypt if you are using it."
    when: ip_resolved_status == "undefined"`

swalkinshaw:
@artifex404 that sounds like a good idea 👍 We could add that early on in `common` somewhere. Do you want to try adding it?

PR #1078, opened 2019-04-11
Show custom error message if external IP resolution fails

artifex404:
This is a PR regarding this issue #949


swalkinshaw:
Simple enough, thanks 🎉

PR #1086, opened 2019-06-13
Update apt packages before checking essentials

amelotrgr:
This step will prevent errors such as "No package matching 'build-essential' is available."
That happened a lot for me when I run the provisioning on Digital Ocean droplets Ubuntu 18.04.

### DIFF
--- a/roles/common/tasks/main.yml
+++ b/roles/common/tasks/main.yml
@@ -110,6 +110,10 @@
   when: openssh_6_8_plus and validate_ssh | default(true)
   tags: [sshd]
 
+- name: Update apt packages
+  apt:
+    update_cache: yes
+
 - name: Checking essentials
   apt:
     name: "{{ item.key }}"


swalkinshaw:
Does this happen after the first provision?

Should note that the task below sets `cache_valid_time` (the default is 60 minutes). The docs say

> Update the apt cache if its older than the cache_valid_time. This option is set in seconds.
As of Ansible 2.4, if explicitly set, this sets update_cache=yes.

amelotrgr:
Yes

> Does this happen after the first provision?
> 
> Should note that the task below sets `cache_valid_time` (the default is 60 minutes). The docs say
> 
> > Update the apt cache if its older than the cache_valid_time. This option is set in seconds.
> > As of Ansible 2.4, if explicitly set, this sets update_cache=yes.

Yes. That happens when I create a new droplet in Digitalocean and, right after, I run the provisioning. The problem is solved when I get into the VM and run a basic apt-get update.

According to the the Ansible docs, it says that this playbook is basically an apt-get update command. I tried that and worked.

slackday:
Can confirmed it just happened to me as well on a new Ubuntu 18.04 droplet on Digital Ocean. Solved by manually running `apt-get update` on the server as @amelotrgr suggested.

mAAdhaTTah:
Just ran into this myself, provisioning a new droplet. The solution in the PR worked for me. I wonder if this happens because the cache on a new droplet isn't "stale" enough for the Checking Essentials task to be required to update it.

swalkinshaw:
>  I wonder if this happens because the cache on a new droplet isn't "stale" enough for the Checking Essentials task to be required to update it.

That makes sense. Going to merge this since it's straightforward enough.

Issue #1082, opened 2019-05-15
Fix for Vagrantfile when using VMware


pepijn-vanvlaanderen:
* [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)
* [x] This request isn't a duplicate of an existing issue
* [x] I've read the [docs](https://roots.io/documentation/) and followed them (if applicable)
* [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/) community

## Description

Vagrant up does not work when using VMware as provider. 
This is due to a config issue in the Vagrantfile on line 154:
```vmw.name = config.vm.hostname```

This should be
```vmw.vmx['displayName'] = config.vm.hostname```

## Steps to reproduce

1. Run vagrant up --provider vmware_fusion

**Expected behavior:** 
The box will start

**Actual behavior:**
```
There are errors in the configuration of this machine. Please fix
the following errors and try again:

VMware Desktop Provider:
* The following settings shouldn't exist: name
```

**Reproduces how often:** 
100% of the time

## Versions
Latest

ricsmo:
I was just going to report the same issue and submit the same fix.

swalkinshaw:
Fix would be welcome 👍

swalkinshaw:
Closed by #1082

PR #1089, opened 2019-06-19
Fixed Vagrant VMware provider display name

ricsmo:
See: https://github.com/roots/trellis/issues/1082

There isn't a "name" parameter in VMware Vagrant config: https://www.vagrantup.com/docs/vmware/configuration.html


swalkinshaw:
Thank you ❤️

PR #1090, opened 2019-06-21
Ensure ssh config dir exists before writing file

swalkinshaw:
Ref: https://discourse.roots.io/t/playbookcli-object-has-no-attribute-options/15809

This assumed `~/.ssh/` would exist and just the file didn't.

Now it will create the directory as well if it doesn't exist.

cc @retlehs


geekodour:
why is this not merged yet, anything needs to be added? @retlehs ping

PR #1091, opened 2019-06-22
Setup composer HTTP basic authentication for multiple repositories

TangRufus:
## Usage

```yaml
# group_vars/<env>/vault.yml

vault_wordpress_sites:
  typist.tech:
    composer_authentications:
      - { hostname: my.com, username: my-username, password: my-password }
      - { hostname: your.com, username: your-username, password: your-password }
  example.com:
    composer_authentications:
      - { hostname: his.com, username: his-username, password: his-password }
      - { hostname: her.com, username: her-username, password: her-password }
```

### Loops

Bumping ansible `version_requirement` to `2.5.0.0` to start using [`loop`](https://docs.ansible.com/ansible/latest/user_guide/playbooks_loops.html).

> With the release of Ansible 2.5, the recommended way to perform loops is the use the new loop keyword instead of with_X style loops.
>
> https://docs.ansible.com/ansible/latest/porting_guides/porting_guide_2.5.html?highlight=loop#migrating-from-with-x-to-loop

## Setup packagist.com authentication

Keeping the original `Setup packagist.com authentication` task for backward compatibility. 

Maybe remove it in Trellis v2?

### Vagrant

When you remove an item from `composer_authentications`, re-provisioning vagrant VM won't delete the removed item. 

Send pull request if you have a better way to do it.

### Why the ugly `include_tasks: tasks/composer-authentications.yml`?

Seems ansible has no way to make a `foreach` inside a `foreach` loop. 

Send pull request if you have a better way to do it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -94,7 +94,3 @@ workflows:
           name: syntax-check-python-2-ansible-2.5
           python-version: '2'
           ansible-version: ~=2.5.0
-      - trellis/syntax-check:
-          name: syntax-check-python-2-ansible-2.4
-          python-version: '2'
-          ansible-version: ~=2.4.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+### Head
+* Setup composer HTTP basic authentication for multiple repositories ([#1091](https://github.com/roots/trellis/pull/1091))
+
 ### 1.0.3: April 30th, 2019
 * Prevent direct access for `.blade.php` files ([#1075](https://github.com/roots/trellis/pull/1075))
 * Show custom error message if external IP resolution fails ([#1078](https://github.com/roots/trellis/pull/1078))
--- a/group_vars/all/helpers.yml
+++ b/group_vars/all/helpers.yml
@@ -18,3 +18,9 @@ ssl_enabled: "{{ item.value.ssl is defined and item.value.ssl.enabled | default(
 ssl_stapling_enabled: "{{ item.value.ssl is defined and item.value.ssl.stapling_enabled | default(true) }}"
 cron_enabled: "{{ site_env.disable_wp_cron and (not item.value.multisite.enabled | default(false) or (item.value.multisite.enabled | default(false) and item.value.multisite.cron | default(true))) }}"
 sites_use_ssl: "{{ wordpress_sites.values() | map(attribute='ssl') | selectattr('enabled') | list | count > 0 }}"
+
+# For backward compatibility, to be removed in Trellis v2.
+site_packagist_org_authentications:
+  - { hostname: repo.packagist.com, username: token, password: "{{ vault_wordpress_sites[site].packagist_token | default('') }}" }
+site_composer_authentications: "{{ vault_wordpress_sites[site].composer_authentications | default([]) }}"
+composer_authentications: "{{ site_packagist_org_authentications + site_composer_authentications }}"
--- a/lib/trellis/plugins/vars/version.py
+++ b/lib/trellis/plugins/vars/version.py
@@ -14,7 +14,7 @@ except ImportError:
     from ansible.utils.display import Display
     display = Display()
 
-version_requirement = '2.4.0.0'
+version_requirement = '2.5.0.0'
 version_tested_max = '2.7.5'
 python3_required_version = '2.5.3'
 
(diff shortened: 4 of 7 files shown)


swalkinshaw:
Well this is much better! Shame we didn't do it this way originally.

> Keeping the original Setup packagist.com authentication task for backward compatibility.
> Maybe remove it in Trellis v2?

It might be possible to keep the existing `packagist_token` variable but append it to the new var if its defined with the default hostname/username we use now? It doesn't get rid of the variable setting, but it would eliminate the existing task at least?

TangRufus:
> It might be possible to keep the existing packagist_token variable but append it to the new var if its defined with the default hostname/username we use now? 

Done.

The current implementation is ugly. Leave a comment if you think of more elegant alternatives. 

Note: Need help testing this pull requests on vagrant!

TangRufus:
rebased onto https://github.com/roots/trellis/pull/1092 and removed ansible 2.4 from trhe test matrix

swalkinshaw:
Working fine for me locally. Config settings got set properly. Only thing I noticed was this:


```plain
TASK [wordpress-install : Setup composer authentications - example.com] ********
skipping: [default] => (item=None)
changed: [default] => (item=None)
changed: [default] => (item=None)
```

`item=None` is a bit confusing but it did work.

TangRufus:
Added \`loop_control.label\` and `changed_when: false`.

Example (vagrant): 

```
TASK [wordpress-install : Setup composer authentications - google.com] *********
skipping: [default] => (item=repo.packagist.com)
ok: [default] => (item=his.com)
ok: [default] => (item=her.com)

TASK [wordpress-install : Setup composer authentications - typist.tech] ********
ok: [default] => (item=repo.packagist.com)
ok: [default] => (item=my.com)
ok: [default] => (item=your.com)
```

Example (deploy): 

```
TASK [deploy : Setup composer authentications] *********************************
skipping: [123.123.123.123] => (item=repo.packagist.com)
ok: [123.123.123.123] => (item=my.com)
ok: [123.123.123.123] => (item=your.com)
```

swalkinshaw:
✨ fancy, didn't know about that

swalkinshaw:
Thank you!

tomektomczuk:
how would you authenticate plugins from private repository as well as paid plugins like WPMU?

TangRufus:
This feature has not been documented in https://roots.io/docs/trellis. One example can be found here https://discourse.roots.io/t/gitlab-private-repo-as-composer-package/20592/19


PRs to https://github.com/roots/docs/ is welcomed.

tomektomczuk:
What about WPMU? Will this work ?

`website_adres:
    composer_authentications:
      - { hostname: wpmudev.com, username: api_key , password: null }`

TangRufus:
Try it and report back please.

tomektomczuk:
> Try it and report back please.

it doesn't work
`failed: [3.67.6.215] (item=None) => {"censored": "the output has been hidden due to the fact that 'no_log: true' was specified for this result", "changed": false}`

any suggestions?

swalkinshaw:
I suggest running the composer commands manually locally (or on your server) to see if it works. It's likely to be an actual issue with Trellis itself.

Here's what Trellis does: https://github.com/TangRufus/trellis/blob/9c5ac561a8df49bfa3c5294dea543687f74ed18a/roles/deploy/hooks/build-after.yml#L13-L16

So if you run `composer config --auth http-basic.wpmudev.com myuser mypassword` then you'll be able to debug and see if it works or not.

tomektomczuk:
so composer config --auth http-basic.wpmudev.com on local came up with {"username":"APIvalue","password":null}. It also added auth.json file next to composer.json. On the server this command gave me 
[RuntimeException] http-basic.wpmudev.com is not defined.

When I removed composer_authentications from the vault and deploy with auth.json 
`{
    "http-basic": {
        "wpmudev.com": {
            "username": "APIvalue",
            "password": null
        }
    }
}`

Deploy went through

TangRufus:
Can I confirm `"password": null` is a JSON `null` instead of a string `"null"` (with quotes)?

tomektomczuk:
> Can I confirm `"password": null` is a JSON `null` instead of a string `"null"` (with quotes)?

yes, I have it as `composer_authentications: - { hostname: wpmudev.com, username: 123456789 , password: null }`

however when I run `composer config --auth http-basic.wpmudev.com`

```
auth.json looks like {
    "http-basic": {
        "wpmudev.com": {
            "username": "myuser",
            "password": "mypassword"
        }
    }
}
```

TangRufus:
Thanks for reporting. Created https://github.com/roots/trellis/issues/1306 to track the issue.
PRs are welcomed.

PR #1092, opened 2019-06-23
Move to CircleCI

Log1x:
Should be pretty optimized. Since we install Ansible without a `requirements.txt` – I wasn't sure what to use for the cache checksum so I went with `requirements.yml` for now.

Also removed the `--force` on the `ansible-galaxy install` so it can use the cached roles without forcefully redownloading them.


swalkinshaw:
I *think* without `--force` updated role versions won't be downloaded (ansible-galaxy kind of sucks).

TangRufus:
> I think without --force updated role versions won't be downloaded (ansible-galaxy kind of sucks).

```diff
  - restore_cache:
    keys:
      - ansible-v1-{{ checksum "requirements.yml" }}
-     - ansible-v1-
```

When `requirements.yml` checksum changes:
Instruct CircleCI not to `restore_cache` so that `--force` is not required.


When `requirements.yml` checksum not changed:
CircleCI resue the cache (with correct role versions installed), so that `--force` is not required.

See: https://github.com/Log1x/trellis/pull/1/

TangRufus:
Question: Do we need to lock anisble version? Or, run tests on different versions of ansible?

Example: Current Trellis doesn't work with ansible 2.8

.circleci/config.yml
```diff
@@ -0,0 +1,57 @@
...
+          - checkout
+          - restore_cache:
+              keys:
+                - ansible-v1-{{ checksum "requirements.yml" }}
+          - run:
+              name: Install Python dependencies in a venv
+              command: |
+                virtualenv venv
+                . venv/bin/activate
+                pip install ansible
```
swalkinshaw: Yeah as @TangRufus mentioned, this should be locked to a specific version. Let's pick the latest supported version? `2.7.5` right now

TangRufus:
Question: Any reason why we use `circleci/python:2.7-stretch` and `circleci/python:3.6-stretch` instead of `circleci/python:2-stretch` and `circleci/python:3-stretch`?

Log1x:
> Question: Any reason why we use `circleci/python:2.7-stretch` and `circleci/python:3.6-stretch` instead of `circleci/python:2-stretch` and `circleci/python:3-stretch`?

Nah. The 2/3-stretch should be fine.

TangRufus:
https://github.com/Log1x/trellis/pull/2 updated

swalkinshaw:
@Log1x do you need to port those changes over to here?

Log1x:
Do we want to test that many variations of Ansible? If I recall, the test will always pass even on versions that are incompatible since it is merely just doing a syntax check– so it seems somewhat redundant to run 11 jobs.

TangRufus:
> Do we want to test that many variations of Ansible?

Yes, beacuse anisble introduce new syntax. For example: anisble 2.5 addes [`loop`](https://docs.ansible.com/ansible/latest/user_guide/playbooks_loops.html).

Testing against different ansible versions helps finding out the [lowest supported version](https://github.com/roots/trellis/blob/e489fe793062d012569e7879541809fd2b220f7b/lib/trellis/plugins/vars/version.py#L17-L19).

swalkinshaw:
Looks good! Thank you both

PR #1093, opened 2019-06-23
CircleCI: Add `syntax-check`

TangRufus:
Another approach to #1092 

This pull request focus on checking syntax on playbooks. Thus, it tests against different ansible versions instead of different python verions.

#1092 will be useful when we have tests against `.py` files.

Not testing on ansible 2.4 beacuse `add-apt-repository ppa:ansible/ansible-2.4 && apt update && apt install anisble` seems to install anisble v2.5. Reason is unknown.
See: https://circleci.com/gh/TangRufus/trellis/37


swalkinshaw:
Can we do both? Python version shouldn't matter that much for Ansible, but it might?

So it's a matrix of python version + ansible version.

TangRufus:
> Python version shouldn't matter that much for Ansible, but it might?

Good catch.
Working on it - https://github.com/TangRufus/trellis/tree/add-circleci

PR #1099, opened 2019-08-07
Add an asible-vault proxy.

schuhwerk:
It encrypts/decrypts all files called vault.yml in the group_vars directory (and all sub-directories).
I had it in my repo and thought it might me helpful.

### DIFF
--- a/bin/vault.sh
+++ b/bin/vault.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+
+show_usage() {
+echo "Usage: vault <command> [<options>...]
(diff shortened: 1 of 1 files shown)


retlehs:
thanks for the PR! have you seen https://github.com/roots/trellis-cli yet?

swalkinshaw:
Yeah this is really useful, but it exists in trellis-cli already and we want that to be the place for these kinds of things. Eventually we'll even move features over there to make the Trellis playbooks simpler.

schuhwerk:
Nice! I haven't seen it. Thank you for the link (and the great work you do)!

PR #1100, opened 2019-08-17
Improve dependency requirements

swalkinshaw:
The ultimate purpose of this PR to improve integration with https://github.com/roots/trellis-cli but these changes are worthwhile enough on their own in my opinion.

Two important but small changes:

1. rename `requirements.yml` to `galaxy.yml`. The "requirements" naming is historically for `pip` dependencies, and since it can be named anything, `galaxy.yml` makes the most sense since it's only for ansible-galaxy and galaxy packages.
2. add a proper `requirements.txt`. It was pretty silly that we never had this anyway. Trellis has dependencies on Ansible (and `passlib` on macOS) which we only ever talked about in the README and docs. This just codifies them in a `requirements.txt` which people can use with `pip` (and `virtualenv`) if they use it. If they don't, nothing changes anyway.

These changes will also make it possible to trellis-cli to automatically install dependencies.


requirements.txt
```diff
@@ -0,0 +1,2 @@
+ansible==2.7.5
```
swalkinshaw: This should probably be a range with our min supported version

TangRufus:
Looks good to me

swalkinshaw:
We don't mention Galaxy at all on purpose right now since Vagrant takes care of it by default. But I do need to update the real docs eventually.

PR #1102, opened 2019-09-02
Bump ansible requirement to 2.7.12

swalkinshaw:
Ansible <2.7.12 has some vulnerabilities


TangRufus:
We can remove ansible <2.7.12 from https://github.com/roots/trellis/blob/e776cc0ad94cb0c4caf8550695fcb828cac1eb29/.circleci/config.yml as well.

swalkinshaw:
Removed!

PR #1096, opened 2019-07-10
Fixes failing ansible_local installation

ipsq:
Currently, there is a bug in the way ansible_local is being installed resulting in an error code. 

See also:
- https://github.com/hashicorp/vagrant/issues/10914#issuecomment-503055527
- https://bugs.launchpad.net/ubuntu/+source/openssl/+bug/1832919

PR #1104, opened 2019-09-03
Fix issues with Vagrant ansible_local provisioner

swalkinshaw:
Replacement for https://github.com/roots/trellis/pull/1096. Thank you @ipsq for the original but there was another issue to fix and a tweak to make.


swalkinshaw:
See https://github.com/roots/trellis/pull/1096

PR #1116, opened 2019-10-13
wordpress-setup: Block dependency manager files by Nginx

TangRufus:
Block these dependency managers:
- composer
- npm
- yarn
- bundler

`auth.json` should never be commited under `web_root`. Blocking it because better safe than worry.

bundler should never be commited under `web_root` neither. Blocking it prevent https://github.com/roots/bedrock-capistrano misuses.

Note: This needs testing.

### DIFF
--- a/roles/wordpress-setup/templates/wordpress-site.conf.j2
+++ b/roles/wordpress-setup/templates/wordpress-site.conf.j2
@@ -129,6 +129,36 @@ server {
   }
   {% endblock %}
 
+  {% block dependency_managers -%}
+  # composer
+  location ~* composer\.(json|lock)$ {
+    deny all;
+  }
+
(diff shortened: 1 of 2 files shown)


swalkinshaw:
I'm good with this in concept 👍

roles/wordpress-setup/templates/wordpress-site.conf.j2
```diff
@@ -129,6 +129,42 @@ server {
   }
   {% endblock %}
 
+  {% block composer -%}
```
swalkinshaw: I'm not sure these each need blocks defined. Maybe just one for all of them?

swalkinshaw:
Looks great, thanks 🎉

PR #1120, opened 2019-11-14
Change self-signed certificate expiry time

robrecord:
from 3650 days to 825, the new maximum time allowed under macOS Catalina


swalkinshaw:
Thanks 🎉

PR #1063, opened 2019-01-20
[WIP] App-agnostic / Laravel support

retlehs:
follow this branch/PR as we work on making trellis more app-agnostic and add laravel support out of the box

todo
* [ ] docs updates
* [ ] wordpress separation/tagging
* [ ] laravel support

ref #914, #951

Issue #914, opened 2017-11-05
Support any PHP + Composer based application


retlehs:
with some modifications to trellis we can support more than just a wordpress setup

i've personally modified trellis to support a laravel app and it didn't take too much effort. trellis itself is a very valuable project and it would be nice to be more CMS/app-agnostic out of the box

- [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/guidelines/blob/master/CONTRIBUTING.md)

retlehs:
see #187 for previous initial attempt by @nathanielks

inthedeepend:
I have a fork of Trellis working with Symfony - I'd be happy to contribute on this.

partounian:
I would love this as I am currently tearing up trellis to support my php app.

nlemoine:
Would love to see that happen. There's not many differences between a WordPress and any other PHP application.

hemangajmera:
This one is open from very long time. There are some pull request too which does the work.

@swalkinshaw are you aware what is blocking it?

swalkinshaw:
No I'm not. @retlehs would be the person to ask but it's been so long I'm not sure he'd remember 😓

baradhili:
would be happy to test this @swalkinshaw

PR #1123, opened 2019-11-20
[WIP] App-agnostic / Laravel support Updates

lukasbesch:
I used #1063 as a starting point to support both WordPress and Laravel.

Inspired by: [pepijn-vanvlaanderen/trellis-laravel](https://github.com/pepijn-vanvlaanderen/trellis-laravel), [ptibbetts/anstead](https://github.com/ptibbetts/anstead) / #951, #914

I tested the provisioning and deployment on a fresh remote machine running both a bare-bone Bedrock installation and a Laravel app.

**Don't use this in production!**

Summary of changes:
- [x] Merged current trellis version
- [x] Updated the apps `type` property behaviour.
Defaults to `wordpress`. Can be `laravel`.
- [x] Add a `web_root` property to apps.
Defaults to `web` or to `public` if `type` is `laravel`.
- [x] App-agnostic configuration
  - `app_env_defaults`
  - `nginx_skip_cache_uri`
  - `nginx_skip_cache_cookie`
  -  nginx `site.conf` (uploads, X-Frame-Options, multisite)
- [x] App-agnostic install and setup tasks.
- [ ] [Database creation looks for multiple environment variables](https://github.com/roots/trellis/blob/8b223322241940b257e274744842c64012fc2dbb/roles/app-setup/tasks/database.yml#L5-L18). _I think it can be done better._
- [x] Deploy sub tasks for Laravel:
  - run migrations
  - flush caches
  - link storage
  - reload php
  - restart queue workers
- [x] Add Redis role
- [x] Add Node support using NVM
- [x] Added [Example for building assets on the remote machine](https://github.com/roots/trellis/blob/8b223322241940b257e274744842c64012fc2dbb/deploy-hooks/build-after.yml)
- [ ] Conditional roles Memcached, Redis, Node (#951 solves this)
only if sites with the corresponding app type are available (like `sites_use_ssl`)
or if explicitly specified.
- [ ] docs update

Adding Node should be discussed. I use Laravel Mix and it loads env variables like `MIX_ASSET_URL` during build, so the config has to be the same as on production ideally. An alternative approach would be to apply the production configuration locally (making a backup of the original .env file or something).


hemangajmera:
It seems there is lots of good commits in this pull request. Any idea when are we planning to merge this.

vesper8:
please merge this!

swalkinshaw:
Closing this and https://github.com/roots/trellis/pull/1123 because they are both quite old and out of date now. I'm open to someone contributing a newer PR to accomplish this but please post in either https://github.com/roots/trellis/issues/914 or https://github.com/roots/trellis/pull/951 first to discuss the solution.

PR #1124, opened 2019-12-07
Add git_sha and release_version to .env on deploy

swalkinshaw:
This will provide `GIT_SHA` and `RELEASE_VERSION` env variables. They can be useful for many purposes including application/error monitoring software.


tjedrak:
Hi @swalkinshaw 
I tried to set Sentry and use `RELEASE_VERSION` & `GIT_SHA` from .env but there are no variables there and I have Trellis 1.4.0.
Screenshot 2020-06-29 at 15 27 27

Whenever I try to deploy local I never manage to get these two vars.

swalkinshaw:
> deploy local

What does this mean exactly? Usually deploying is for remote servers only. You wouldn't get these env vars in Vagrant if that's the question.

tjedrak:
Sorry, by deploying locally I meant `vagrant up --provision`. My goal was to test Sentry (send event) on local env. I used your [gist](https://gist.github.com/swalkinshaw/aab8ceff6001c65519e4334574a4f111) but couldn't manage to do it; now I understand why.

PR #1132, opened 2020-01-08
`composer install` with `--classmap-authoritative` during deploy

TangRufus:
See:
- https://getcomposer.org/doc/articles/autoloader-optimization.md#optimization-level-2-a-authoritative-class-maps
- https://docs.ansible.com/ansible/latest/modules/composer_module.html#parameters

### DIFF
--- a/roles/deploy/defaults/main.yml
+++ b/roles/deploy/defaults/main.yml
@@ -50,6 +50,8 @@ update_db_on_deploy: true
 
 # Most scripts are used in development instead of remote servers. Use with caution.
 composer_no_scripts: true
+# Whether to autoload classes from classmap only.
+composer_classmap_authoritative: true
 
 # Helpers
 project: "{{ wordpress_sites[site] }}"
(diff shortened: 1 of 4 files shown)


roles/deploy/defaults/main.yml
```diff
@@ -48,6 +48,9 @@ project_current_path: "{{ project.current_path | default('current') }}"
 # Whether to run `wp core update-db` at end of each deploy
 update_db_on_deploy: true
 
+# Whether to autoload classes from classmap only.
+composer_classmap_authoritative: false
```
swalkinshaw: Are you thinking that `true` isn't safe enough as the default?

TangRufus: Correct.
I am running `--classmap-authoritative` to see any of my sites break.

Maybe default to `true` on next major relrease

TangRufus: After a second thought, it is very difficult to make a wp plugin that:
- generates classes dynamically, and
- relies on bedrock's composer autoloader

Defaults `composer_classmap_authoritative` to `true`

Ready for review

PR #1139, opened 2020-01-15
Lint: Fix `[301] Commands should not change things if nothing needs doing`

TangRufus:


roles/common/tasks/reload_nginx.yml
```diff
@@ -2,6 +2,7 @@
 - name: reload nginx
   command: nginx -t
   notify: "{{ (role_path | basename == 'common') | ternary('perform nginx reload', omit) }}"
+  changed_when: false
```
swalkinshaw: 🤔 I think this is breaking things. Both in production and with Vagrant Nginx isn't reloading and you just get the default Nginx page. I'll test without it.

swalkinshaw: Yep this prevents nginx from actually reloading

TangRufus: Looks like we need `changed_when: true` to trigger `notify`. See: https://serverfault.com/a/799282

swalkinshaw: We still need this as `changed_when: true` true? Which I'm guessing fails the lint... any idea what the solution is?

TangRufus: `changed_when: true` passes `ansible-lint` and triggers nginx reload. Tested use cases:
1. `vargant up` the first time
2. change site domains and then `vagrant provision`

https://stackoverflow.com/a/35604895

TangRufus: It leads to the question that should we `changed_when: true` on other tasks:
- `Get list position of current symlinked release`: No because it doesn't change anything
- `Collect list of releases`: No because it doesn't change anything
- `Get name of current symlinked release`: No because it doesn't change anything
- `Create or close Xdebug SSH tunnel`: Yes because it changes the tunnel. See: 78ebc81

swalkinshaw: Agreed with those 👍

PR #1140, opened 2020-01-15
Lint: Fix `[303] rsync used in place of synchronize module`

TangRufus:


roles/wordpress-install/tasks/main.yml
```diff
@@ -11,10 +11,11 @@
   with_dict: "{{ wordpress_sites }}"
 
 - name: Copy .env file into web root
-  command: rsync -ac --info=NAME /tmp/{{ item.key }}.env {{ www_root }}/{{ item.key }}/{{ item.value.current_path | default('current') }}/.env
+  synchronize:
```
swalkinshaw: This fails for me locally with Vagrant.

```
rsync: link_stat "/tmp/example.com.env" failed: No such file or directory (2)
rsync error: some files could not be transferred (code 23) at /BuildRoot/Libr
ary/Caches/com.apple.xbs/Sources/rsync/rsync-52.200.1/rsync/main.c(996)
[sender=2.6.9]

failed: [default] (item=example.com) => {"changed": false, "cmd": "/usr/bin/rsync --delay-updates -F --compress --checksum --archive --rsh=/usr/bin/ssh -S none -i /Users/scott/dev/yes-trellis/.vagrant/machines/default/virtualbox/private_key -o Port=2222 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null --rsync-path=sudo rsync --out-format=<<CHANGED>>%i %n%L /tmp/example.com.env vagrant@127.0.0.1:/srv/www/example.com/current/.env", "item": {"key": "example.com", "value": {"admin_email": "admin@example.test", "cache": {"enabled": false}, "local_path": "../site", "multisite": {"enabled": false}, "site_hosts": [{"canonical": "example.test", "redirects": ["www.example.test"]}], "ssl": {"enabled": false, "provider": "self-signed"}}}, "rc": 23}
```

TangRufus: Fixed by adding `delegate_to: "{{ inventory_hostname }}" `

swalkinshaw:
Working 👍

PR #1141, opened 2020-01-15
Lint: Fix `[602] Don't compare to empty string`

TangRufus:
See: https://github.com/ansible/ansible-lint/issues/457#issuecomment-455038475

### DIFF
--- a/roles/wordpress-install/tasks/composer-authentications.yml
+++ b/roles/wordpress-install/tasks/composer-authentications.yml
@@ -7,9 +7,9 @@
   no_log: true
   changed_when: false
   when:
-    - composer_authentication.hostname is defined and composer_authentication.hostname != ""
-    - composer_authentication.username is defined and composer_authentication.username != ""
-    - composer_authentication.password is defined and composer_authentication.password != ""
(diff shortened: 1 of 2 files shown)


roles/wordpress-install/tasks/composer-authentications.yml
```diff
@@ -7,9 +7,9 @@
   no_log: true
   changed_when: false
   when:
-    - composer_authentication.hostname is defined and composer_authentication.hostname != ""
```
swalkinshaw: Do we even need to check these? Seems pretty defensive. By default they just aren't defined right? So this is only guarding against people defining them with no value?

TangRufus: We need it (at least the `password` one) because of `password` could be empty. See: https://github.com/roots/trellis/blob/faaf213ae2186c6ccf55159becd8910e20c82bf1/group_vars/all/helpers.yml#L24

I suggest we remove all 3 `when` conditions when we depercate [`packagist_token`](https://github.com/roots/trellis/pull/762)

PR #1145, opened 2020-01-21
Fix `DEPLOY_UNFINISHED` not being copied over to `release`

TangRufus:
The tar generated by `git archive` doesn't include `DEPLOY_UNFINISHED`
https://github.com/roots/trellis/blob/ab70d8ed84ef54054de2024e5c2639e047c0b905/roles/deploy/tasks/prepare.yml#L29-L39


swalkinshaw:
😱 nice catch

TangRufus:
Should we tag this a breaking change?

Before this PR, failed releases count towards `deploy_keep_releases`. Failed releases only being deleted when reaching `deploy_keep_releases`.

After this PR, a successful deploy will removes all releases with `DEPLOY_UNFINISHED` during finalize.

The behavior breaks how `rollback.yml` work currently.

swalkinshaw:
It's technically breaking but it seems like it's correcting behaviour more than anything (behaviour which people probably didn't notice?)

PR #1147, opened 2020-01-22
Bump ansible requirement to 2.8.0

TangRufus:
https://github.com/roots/trellis/pull/1141 bases on https://github.com/ansible/ansible-lint/issues/457#issuecomment-455038475
which depends on https://github.com/ansible/ansible/pull/51030 which merged into anisble v2.8.0.

See: https://discourse.roots.io/t/provisioning-server-not-working-because-ansible-ssh-extra-args-is-undefined/17484


swalkinshaw:
I think we have some 2.7 compat code which would be nice to get rid of too. It's probably under `lib` but I don't remember exactly.

TangRufus:
Never worked with python. No idea where to start.
Can you help? @swalkinshaw

swalkinshaw:
Main thing I remember is this: https://twitter.com/shoshmcgosh/status/1219781290565619712

I believe with 2.8 we can just rely on `context` existing. That might be the only compat code actually

swalkinshaw:
Updated in https://github.com/roots/trellis/pull/1166

PR #1148, opened 2020-01-22
Void rolled back releases

TangRufus:
### Problem A

1. deploy (`current` symlink to `releases/11111111`)
1. deploy some changes (`current` symlink to `releases/22222222`)
1. find out `releases/22222222` is buggy; rollback (`current` symlink to `releases/11111111`)
1. deploy bug fixes (`current` symlink to `releases/33333333`)
1. find out `releases/33333333` is also buggy; rollback (`current` symlink to `releases/22222222`)

We already know `releases/22222222` is buggy and rolled back in step 3. However, Trellis doesn't know. Trellis rolls back to the last release which older than `current`

https://github.com/roots/trellis/blob/ac87cd5af42ea8194499606119b610db265ec501/roles/rollback/tasks/prior-release.yml#L21

### Solution A

Write `DEPLOY_UNFINISHED` to release we move away from.

---

### Problem B

If the last release which older than `current` contains `DEPLOY_UNFINISH`, rollback fails.

https://github.com/roots/trellis/blob/ac87cd5af42ea8194499606119b610db265ec501/roles/rollback/tasks/main.yml#L8-L11

### Solution B

Clean up old and failed (`DEPLOY_UNFINISH`) releases at the beginning of `rollback.yml`

---

Do we count this as beraking changes?

### DIFF
--- a/roles/rollback/tasks/main.yml
+++ b/roles/rollback/tasks/main.yml
@@ -1,22 +1,31 @@
 ---
+- name: Get real path of current symlinked release
+  command: "readlink {{ project_current_path }}"
+  args:
+    chdir: "{{ project_root }}"
+  register: current_release_readlink_result
+
+- name: Clean up old and failed releases
+  deploy_helper:
+    state: clean
+    path: "{{ project_root }}"
+    current_path: "{{ project_current_path }}"
+    release: "{{ current_release_readlink_result.stdout }}"
+    keep_releases: "{{ project.deploy_keep_releases | default(deploy_keep_releases | default(omit)) }}"
+
 - import_tasks: user-release.yml
   when: release is defined
 
 - import_tasks: prior-release.yml
   when: release is not defined
 
-- name: Check whether target release was from a successful deploy
-  stat:
-    path: "{{ new_release_path }}/DEPLOY_UNFINISHED"
-  register: target
-
-- name: Fail if target release was from failed deploy
-  fail:
-    msg: "Cannot switch to release at {{ new_release_path }}. It is from an unfinished deploy. You may manually specify a different release using --extra-vars='release=12345678901234'."
-  when: target.stat.exists | default(False)
-
 - name: Link 'current' directory to target release
   file:
     path: "{{ project_root }}/{{ project_current_path }}"
     src: "{{ new_release_path }}"
     state: link
+
+- name: Write unfinished file to old symlinked release
+  file:
+    path: "{{ current_release_readlink_result.stdout }}/DEPLOY_UNFINISHED"
+    state: touch
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Nginx: Block `composer/installed.json` ([#1150](https://github.com/roots/trellis/pull/1150))
 * Run `git clean` after checking `git clone` is successful ([#1151](https://github.com/roots/trellis/pull/1151))
 * Lint: Fix: `[206] Variables should have spaces before and after: {{ var_name }}` ([#1152](https://github.com/roots/trellis/pull/1152))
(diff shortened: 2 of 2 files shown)


roles/rollback/tasks/main.yml
```diff
@@ -15,8 +15,19 @@
...
+
 - name: Link 'current' directory to target release
   file:
     path: "{{ project_root }}/{{ project_current_path }}"
     src: "{{ new_release_path }}"
     state: link
+
+- name: Write unfinished file to old symlinked release
+  file:
+    path: "{{ current_release_real_path.stdout }}/DEPLOY_UNFINISHED"
```
TangRufus: `deploy_helper.unfinished_filename` is not available during rollback.
Therefore, I hardcoded it.

swalkinshaw:
What if we just deleted roll backed releases? The only downside I can think of is that if you decide to re-deploy a rolled back release it will just take a little longer since it's a full new deploy.

I'm guessing that's rare though since you usually roll back for a reason and apply a fix like you detailed.

TangRufus:
> What if we just deleted roll backed releases?

Do you mean replace https://github.com/roots/trellis/pull/1148/files#diff-1ff8e19172c2564b1d6f0cdc568feebdR28-R31  with something like `$ rm -dfr {{ current_release_readlink_result.stdout }}` (file module of course)?


> The only downside I can think of is that if you decide to re-deploy a rolled back release it will just take a little longer since it's a full new deploy.

Not much differences in deploy speed I believe.
Because `shared/source` & composer caches are untouched by `rollback.yml`

swalkinshaw:
Yep exactly that. I don't know if this changes any of the assumptions in the prior/user release roles though.

TangRufus:
Note that 0c2dd89 might delete more releases than you expected.

During `Clean up old and failed releases`, ansible `keep_releases: "{{ project.deploy_keep_releases | default(deploy_keep_releases | default(omit)) }}"`
Then, at the end of rollback (`Remove the old symlinked release`), ansible delete 1 more.
Thus, you will have max `deploy_keep_releases - 1` releases after rollback.

roles/rollback/tasks/main.yml
```diff
@@ -1,22 +1,31 @@
 ---
+- name: Get real path of current symlinked release
+  command: "readlink {{ project_current_path }}"
+  args:
+    chdir: "{{ project_root }}"
+  register: current_release_readlink_result
+
+- name: Clean up old and failed releases
```
swalkinshaw: Do we need this task then? They get cleaned on deploys

TangRufus: Still needed because deploy cleanup happens at https://github.com/roots/trellis/blob/ac87cd5af42ea8194499606119b610db265ec501/roles/deploy/tasks/finalize.yml#L8-L14

If deploy fails before that task, there will be invalid release on the server.

swalkinshaw:
Seems ok to me. What do you think of the delete method vs the other one?

TangRufus:
I suggest we keep the _rolled back_ release instead of deleteing it, i.e: revert https://github.com/roots/trellis/pull/1148/commits/0c2dd896cbad602cf0a3e78e1f805052bdf15952

Reasons:
- users might want to look at the source code for debugging
- user might have files generated under the _rolled back_ release, example: https://github.com/ItinerisLtd/trellis-backup-during-deploy
- keeping the _rolled back_ release doesn't hurt and it will be deleted during next successful deploy

swalkinshaw:
Sounds good to me 👍

TangRufus:
Dropped `0c2dd89` & rebased onto master & added changelog

TangRufus:
rebased, ready for review

PR #1161, opened 2020-03-23
Add nvm support in build-before deploy hook

strarsis:
Add support for node versions using nvm to the build-before deploy hook example.
Also, switch back from yarn to npm because npm should have become fast enough now.

This solves issues with lots of red text when the wrong (like system) node version is used instead the one known to build correctly for the theme project.

### DIFF
--- a/deploy-hooks/build-before.yml
+++ b/deploy-hooks/build-before.yml
@@ -7,8 +7,28 @@
 # Uncomment the lines below and replace `sage` with your theme folder
 #
 # ---
+#
+# - name: Use project specific node version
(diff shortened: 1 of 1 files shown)


swalkinshaw:
I think it would be helpful to have this documented somewhere but I'm not sure if this is the best place for it.

1. this file is meant to hold the example for Sage and its docs use `yarn`, not `npm`
2. it makes it more complicated and assumes people use nvm
3. I'm not sure the `install` part should be done inline in this file?

Ideas:
1. publish more examples in our docs and link there from this file
2. create some new repo with the examples as code

strarsis:
@swalkinshaw: Yes, as an example this is better suited.

PR #1164, opened 2020-03-31
Update to PHP 7.4

### LINKED ISSUE #1163: Update to PHP 7.4
## Summary

Trellis should install PHP 7.4.

## Motivation

Support for the latest PHP features.

## Additional context

- [PHP 7.3 was merged into Trellis][merge-7.3] about two week after its release.
- [PHP 7.4 has been out][php-releases] since November 28, 2019.
- [WordPress 5.3 fully supports PHP 7.4][wp] and has been out since November 12, 2019.

[merge-7.3]: https://github.com/roots/trellis/pull/1052
[php-releases]: https://www.php.net/supported-versions.php
[wp]: https://wordpress.org/news/2019/11/kirk/

markjaquith:
fixes #1163

I left the PHP 7.2 service code in there, as well as adding new PHP 7.3 service code. I figure it doesn't hurt to keep it (it's just a few lines), and if anyone is upgrading from a 2018 version of Trellis to the PHP 7.4 version, it'll handle them leapfrogging PHP 7.3 completely.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### Head
+* Update PHP to 7.4 ([#1164](https://github.com/roots/trellis/pull/1164))
 * Update `wp_cli_version` to 2.4.0 ([#1131](https://github.com/roots/trellis/pull/1131))
 * Fix `subjectAltName` for self-signed certificates ([#1128](https://github.com/roots/trellis/pull/1128))
 * `composer install` without `--no-scripts` during deploy ([#1133](https://github.com/roots/trellis/pull/1133))
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Trellis will configure a server with the following and more:
 
 * Ubuntu 18.04 Bionic LTS
 * Nginx (with optional FastCGI micro-caching)
-* PHP 7.3
+* PHP 7.4
 * MariaDB (a drop-in MySQL replacement)
 * SSL support (scores an A+ on the [Qualys SSL Labs Test](https://www.ssllabs.com/ssltest/))
 * Let's Encrypt for free SSL certificates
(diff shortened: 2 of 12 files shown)


swalkinshaw:
I'll give this a try tonight but looks great 🎉

markjaquith:
> Didn't test on vagrant.

I did! Patch applied cleanly against my `/trellis/` directory, and I ran `vagrant provision`. No problems at all.

TangRufus:
I mean I didn't

swalkinshaw:
🎉 thank you

axi:
I also updated my /trellis  and ran `ansible-playbook server.yml -e env=staging`

Website (test server ;-) went down for few seconds when ansible ensured php7.3-fpm is down...

Shouldn't that test (ensuring php7.3-fpm is down) be called at the end of provisioning as https://roots.io/docs/trellis/master/remote-server-setup/#re-provisioning states that
> Re-provisioning is always assumed to be a safe operation

Note sure I should create an issue on that. Any idea to prevent that on production ? Just comment that portion should do I think

swalkinshaw:
@axi you mean everything worked properly but php7.3-fpm service was still running? Nginx will be pointed to php7.4-fpm at that point though so it couldn't actually cause any issues? But yes, it should be stopped manually ideally.

> Re-provisioning is always assumed to be a safe operation

This statement is slightly misleading I guess. It's assumed to be "safe" (non-breaking) if the changes themselves are safe.

axi:
@swalkinshaw 
On a running server, during provisionning, php7.3-fpm is shutdown before nginx config has been updated to point to php7.4-fpm. I thing shutting down php7.3-fpm at the end of provisionning should fix the short period where nginx still points to 7.3 but 7.3 is down. It's just a matter of order.
Update went well on every other points ;-)

PR #1166, opened 2020-04-04
Bump ansible requirement to 2.8.0

swalkinshaw:


swalkinshaw:
@TangRufus one thing that confused me... your original PR said 2.8 was required for another PR that was already merged. Does it >=2.7.12 <2.8 had actually been broken already?

TangRufus:
> Does it >=2.7.12 <2.8 had actually been broken already?

Yes.
Current Trellis master doesn't work on ansible 2.7.x
https://discourse.roots.io/t/provisioning-server-not-working-because-ansible-ssh-extra-args-is-undefined/17484

PR #1177, opened 2020-04-22
Make Fail2ban settings extensible

raph-topo:
add the ability to add custom services in `group_vars/`;
add the ability to include custom filters in a new `fail2ban-filters/` folder;
inspired from the `nginx-includes` task in `wordpress-setup` role.

- [x] custom services
- [x] custom filters
- [x] readme
- [x] changelog
- [ ] docs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Make Fail2ban settings extensible ([#1177](https://github.com/roots/trellis/pull/1177))
 * Support Ansible 2.9 ([#1169](https://github.com/roots/trellis/pull/1169))
 * [BREAKING] Remove `nginx_includes_deprecated` feature ([#1173](https://github.com/roots/trellis/pull/1173))
 * Bump Ansible version_tested_max to 2.8.10 ([#1167](https://github.com/roots/trellis/pull/1167))
--- a/roles/fail2ban/README.md
+++ b/roles/fail2ban/README.md
@@ -1,19 +1,17 @@
-## What is ansible-fail2ban?
+## What is this role?
 
-It is an [ansible](http://www.ansible.com/home) role to install and configure fail2ban.
+This role installs and configures [Fail2ban](https://github.com/fail2ban/fail2ban).
 
-### What problem does it solve and why is it useful?
-
-Security is important and fail2ban is an excellent tool to harden your server with minimal or even no configuration.
+Fail2ban is an excellent tool to harden your server with minimal configuration.
 
 ## Role variables
 
-Below is a list of default values along with a description of what they do.
+Below is a list of available variables, their description and their default value within Trellis.
 
-```
+```yaml
 # Which log level should it be output as?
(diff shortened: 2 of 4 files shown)


raph-topo:
A use case would be adding services and filters to integrate the [WP-Fail2ban plugin](https://wordpress.org/plugins/wp-fail2ban/).

swalkinshaw:
Thanks for the suggestion and use case. By default, every variable in Ansible can be overridden in `group_vars` as it takes precedence. So right now, without this PR, you could do the following:

```yaml
# group_vars/all/fail2ban.yml

fail2ban_services:
  - name: ssh
    port: ssh
    filter: sshd
    logpath: /var/log/auth.log
  - name: postfix
    port: smtp,ssmtp
    filter: postfix
    logpath: /var/log/mail.log
```

The `README` in the role points this out: https://github.com/roots/trellis/blob/5de70af75e24d0c00826f079532099728361179a/roles/fail2ban/README.md#example-playbook

The one very minor downside here is you need to include the default `ssh` entry. If you don't it will be removed.

Nginx includes warrant their own extra solution for one main reason: they are templates which can't just be represented in a YAML data structure like these fail2ban services can.

This seems like it would solve your use case; can you confirm that? If so, I'd prefer just using these default features of Ansible in this case instead of adding an extra feature.

raph-topo:
> The one very minor downside here is you need to include the default ssh entry. If you don't it will be removed.
Indeed, but people might miss or forget that and inadvertently deactivate the `ssh` filter. So I believe it is safer to make it extensible by default instead of just replaceable.

> This seems like it would solve your use case; can you confirm that? 
For the services part yes, for the filters part no. They still have to be copied over to the servers, which I introduce in the second commit https://github.com/roots/trellis/pull/1177/commits/d7f8733dd65b84f653a0606c866a225df1f5bf8c

For the WP Fail2ban, [3 filters](https://docs.wp-fail2ban.com/en/4.2/filters.html) and [3 services](https://docs.wp-fail2ban.com/en/4.2/configuration/fail2ban.html) have to be added.

If this convinces you, should I also edit the [Fail2ban Readme](https://github.com/roots/trellis/blob/master/roles/fail2ban/README.md) and add this to the docs in Digging deeper or Security?

raph-topo:
By the forums, it seems [people bumped into this already](https://discourse.roots.io/t/prevent-brute-force-attacks-on-wp-login-php-with-fail2ban/15758) without being able to configure Fail2ban as expected.

swalkinshaw:
Ah thank you, that makes sense! I didn't know about the filters.

Yes updating the README + docs would be great. Thank you 😄

raph-topo:
This PR should be ready, but I am still working on [roots/docs](https://github.com/roots/docs). But I am blocked by an unexpected behavior described in https://github.com/roots/trellis/issues/1181.

raph-topo:
Rebased to HEAD.
OK for merge of this PR.

swalkinshaw:
🎉 thank you this is great. Tested it out and works as expected.

PR #1188, opened 2020-05-27
Silence Ansible warning

raph-topo:
```
TASK [mariadb : Delete anonymous MySQL server users] **************************************************************************************************************************************
ok: [staging_host] => (item=localhost)
ok: [staging_host] => (item=staging_host)
ok: [staging_host] => (item=******)
[WARNING]: Module did not set no_log for update_password
```

### DIFF
--- a/roles/mariadb/tasks/main.yml
+++ b/roles/mariadb/tasks/main.yml
@@ -59,6 +59,7 @@
       user: ""
       host: "{{ item }}"
       state: absent
+    no_log: true
     with_items:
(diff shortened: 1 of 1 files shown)


raph-topo:
I do not actually know why Ansible complains here. I guess it is just yet another "It's not a bug, it's a feature".

swalkinshaw:
Yeah it really shouldn't complain, but also fine to use `no_log` anyway 👍

Issue #1181, opened 2020-05-07
`ip_whitelist` does not whitelist correct IP on development 


raph-topo:
* [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)
* [x] This request isn't a duplicate of an existing issue
* [x] I've read the [docs](https://roots.io/documentation/) and followed them (if applicable)
* [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/) community

## Description

In `group_vars/all/main.yml`:
```yaml
ip_whitelist:
  - "{{ (env == 'development') | ternary(ansible_default_ipv4.gateway, ipify_public_ip | default('')) }}"
```
When setting up development, this does not recognize the developer's local IP.

I detected this when testing Fail2ban ban actions, which do `Ignore` the correct IP on staging and production, but not on development. Yet, the above statement shows Trellis' intention to actually whitelist the developer's local IP.

## Steps to reproduce

1. Ensure `hosts/development` contains:
```ini
[development]
192.168.50.5 ansible_connection=local

[web]
192.168.50.5 ansible_connection=local
```
2. Add the following to `roles/fail2ban/tasks/main.yml`:
```yaml
- name: Debug
  debug:
    var: ip_whitelist
```
3. Run `ansible-playbook dev.yml --tags fail2ban`
4. Read:
```
TASK [fail2ban : Debug] ************************************************************************************************
ok: [development_host] => {
    "ip_whitelist": [
        "xxx.xxx.xxx.xxx"
    ]
}
````

**Expected behavior:**
The debug statement should read:

```
"ip_whitelist": [
        "192.168.50.1"
    ]
```

This is confirmed by Fail2ban reading `Ban 192.168.50.1` when I try to brute-force on development.

**Actual behavior:**
In my case, the debug statement instead reads:
```
"ip_whitelist": [
        "10.0.2.2"
    ]
```
`10.0.2.2` does not correspond to any IP range on my network.
**Reproduces how often:** always when setting up development.
## Versions
HEAD (1.4.0+) on Windows WSL

## Additional information
This might be an Ansible bug, but either way this produces an unexpected behavior in Trellis.
Maybe Trellis should use another fact than `ansible_default_ipv4.gateway`?

swalkinshaw:
[...]

PR #1190, opened 2020-05-27
On development, whitelist all local IP ranges in Fail2ban & Ferm 

raph-topo:
Actually fix https://github.com/roots/trellis/issues/1181.
https://github.com/roots/trellis/pull/1183 did not solve this, see https://github.com/roots/trellis/issues/1181#issuecomment-633496169.

Solution explained by https://stackoverflow.com/a/39388565.

Actually,
```yaml
become: no
gather_facts: yes
```
are the defaults, but I made them explicit so future readers understand why it is in a separate play.

Would you prefer this in a separate role, similarly to `python_interpreter` just above?


raph-topo:
Yikes! The IP is correct, but I end up with `ignoreip = 127.0.0.1/8 1 9 2 . 1 6 8 . 5 0 . 1` in `/etc/fail2ban/jail.local` What the hell is Ansible doing with variables?

raph-topo:
Solved.

swalkinshaw:
I get `10.0.2.2` 😅

raph-topo:
Does WP in dev see you on `192.168.50.1` or `10.0.2.2`?

raph-topo:
> this is just outputting ansible_control_node_ip via debug after its setup and running vagrant provision

Sorry, I did not understand what you mean. Here is my `debug` declaration:
```yaml
- name: Get control node IP
  hosts: web:&development
  become: no
  gather_facts: yes
  tasks:
    - name: Get control node IP
      set_fact:
        ansible_control_node_ip: "{{ ansible_env['SSH_CLIENT'].split() | first }}"
      tags: [fail2ban, ferm]
    - debug:
        var: ansible_control_node_ip
```

raph-topo:
With 
```yaml
    - debug:
        var: ansible_env
```
I get
```json
        "SSH_CLIENT": "192.168.50.1 4037 22",
        "SSH_CONNECTION": "192.168.50.1 4037 192.168.50.5 22",
```
which corresponds to reality.

swalkinshaw:
This is mine:

```plain
        "SSH_CLIENT": "10.0.2.2 56830 22",
        "SSH_CONNECTION": "10.0.2.2 56830 10.0.2.15 22",
```

This is on macOS running Trellis/Vagrant normally (I even recently got a new laptop so it's a fresh setup).

swalkinshaw:
```plain
tail -f /srv/www/example.com/logs/access.log
192.168.50.1 - - [28/May/2020:01:52:12 +0000] "GET / HTTP/2.0" 444 0 "-" "curl/7.64.1" "-"
```

raph-topo:
Thanks!
Well… Maybe, after all, this is the wrong approach. I didn't find anything better on the web. Maybe Ansible is just unable to correctly understand local to local connections when Vagrant and/or Virtualbox are in the middle.

raph-topo:
Re-thinking the problem at hand, I found the following: http://www.fail2ban.org/wiki/index.php/Whitelist.
That's by far the simplest solution: it's static.
Now, whatever the development setup, no local to local connection will ever get banned.
This also means any machine on any currently attached local network can now try to brute-force development. I do not think this opens any vulnerability. What do you think?
(If you are ok with this solution, I'll first squash)

swalkinshaw:
Nice find 🎉 agreed this is much better. I like the re-org too 😄

Issue #1189, opened 2020-05-27
Trellis generates a second, self-signed certificate on Staging and Production


raph-topo:
## Description

Trellis generates a second, self-signed certificate for `example.com` on Staging and Production

## Steps to reproduce

Consult the SSL Report of any Trellis-powered server: https://www.ssllabs.com/ssltest/analyze.html?d=roots.io&hideResults=on in section **Certificate #2: RSA 2048 bits (SHA256withRSA)**.

**Expected behavior:**

One valid certificate for `roots.io`.

**Actual behavior:**

Two certificates: `roots.io` (valid) + `example.com` (invalid).

## Versions

Introduced by https://github.com/roots/trellis/pull/888/commits/90ec6c0f88aeb36c167535ecdea18e0e440e4ac6#diff-6a42c926df0abb8be26f6923b00ee7a5

## Additional information

https://github.com/roots/trellis/pull/888#issuecomment-370516859 does state

> Implementation notes:
> if no sites create a self-signed cert, this creates one for throw-away example.com

But, on Staging and Production, self-signed certs should be very rare.

https://web.archive.org/web/20180819104946/https://tech.mendix.com/linux/2014/10/29/nginx-certs-sni/ seems to explain the need for it; but responding 403 Forbidden when it should probably be 495 SSL Certificate Error or 496 SSL Certificate Required seems odd.

https://github.com/h5bp/server-configs-nginx/pull/177#issuecomment-442769233 makes me think this behavior is outdated, but I am not sure I understand all implications.

Is creating an invalid certificate for an extant domain ([example.com](http://example.com) is a real domain after all) really the intended behavior?

swalkinshaw:
🤔 to be honest I had no idea this behaviour existed. I don't entirely understand all of this yet either...

@TangRufus maybe you can help here. You helped reviewed @fullyint's PR from before.

swalkinshaw:
Okay so here's my conclusion/summary:

* the no-default confs are needed. If you have an SSL enabled site, then we do need the SSL no-default too.
* Nginx requires an SSL cert, otherwise it won't work
* a self-signed cert works fine to satisfy this requirement
* however, you should only care about an SSL no-default server *if you have an SSL enabled site*. and in that case, we could re-use the existing cert instead of generating this additional self-signed cert

So if we can find an easy way to reference the real cert, we can skip this self-signed one.

[...]

PR #1191, opened 2020-05-28
Fix #888 & #1189 to comply with RFC2606

raph-topo:
Fix #888 & #1189 to comply with [RFC2606](https://tools.ietf.org/html/rfc2606#section-2), which specifies that:

>       ".invalid" is intended for use in online construction of domain
>       names that are sure to be invalid and which it is obvious at a
>       glance are invalid.

See discussion in #1189.


swalkinshaw:
Thanks 🎉

PR #1195, opened 2020-06-20
Ansible deprecation warning

raph-topo:
(Nearly) no more warnings with Ansible 2.9.9 during a fresh server setup & deploy. 🥳 

Except for a warning concerning Python 3.

### DIFF
--- a/lib/trellis/plugins/vars/version.py
+++ b/lib/trellis/plugins/vars/version.py
@@ -15,7 +15,7 @@ except ImportError:
     display = Display()
 
 version_requirement = '2.8.0'
-version_tested_max = '2.9.6'
+version_tested_max = '2.9.9'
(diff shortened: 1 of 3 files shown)


nextgenthemes:
@raph-topo did you actually test this to be working? Or just the warnings to disappear? Because its NOT working for me. https://github.com/roots/trellis/issues/1199 is just skips the task.

raph-topo:
I set up a new server from scratch since I added all these `| bool`, without any error. I'll check again.

nextgenthemes:
@raph-topo The `server.yml` actually ran thought without any errors. The sudeors entry(s) just get skipped and the actual error happens at the very end of `deploy.yml` when it tries to restart php with the `web` user. Maybe that is why you missed it.

PR #1196, opened 2020-06-22
Ansible version check

raph-topo:


swalkinshaw:
Wow that typo must have been there forever 😂

PR #1197, opened 2020-06-23
Ubuntu 20.04 support

swalkinshaw:
The good news is that Trellis didn't require many updates to support Ubuntu 20.04 (Focal). The main work was getting compatible releases of MariaDB and Nginx which was done in https://github.com/roots/trellis/pull/1212 and https://github.com/roots/trellis/pull/1208.

This PR itself just defaults the Vagrant image to 20.04 and updates our warning for supported Ubuntu versions.

Note: if you want to continue to use 18.04 locally with Vagrant, you can override the `vagrant_box` and `vagrant_box_version` options in your `vagrant.local.yml` config file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Officially support Ubuntu 20.04 (and default Vagrant to it) ([#1197](https://github.com/roots/trellis/pull/1197))
 
 ### 1.6.0: November 5th, 2020
 * Remove prestissimo for Composer 2.0 support ([#1247](https://github.com/roots/trellis/pull/1247))
--- a/roles/common/tasks/main.yml
+++ b/roles/common/tasks/main.yml
@@ -67,7 +67,9 @@
 - name: Validate Ubuntu version
   debug:
     msg: |
-      Trellis is built for Ubuntu 18.04 Bionic as of https://github.com/roots/trellis/pull/992
+      Ubuntu 18.04 Bionic is the minimum supported version of Ubuntu in Trellis 1.0+ (as of https://github.com/roots/trellis/pull/992)
+
+      20.04 Focal is the recommend version for Trellis 1.7+ (as of https://github.com/roots/trellis/pull/1197)
 
       Your Ubuntu version is {{ ansible_distribution_version }} {{ ansible_distribution_release }}
 
@@ -77,8 +79,8 @@
 
       Development via Vagrant: `vagrant destroy && vagrant up`
 
-      Staging/Production: Create a new server with Ubuntu 18.04 and provision
-  when: ansible_distribution_release != 'bionic'
+      Staging/Production: Create a new server with Ubuntu 20.04 and provision
+  when: ansible_distribution_version is version('18.04', '<')
 
 - name: Check whether passlib is needed
   fail:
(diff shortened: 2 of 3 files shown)


roles/nginx/defaults/main.yml
```diff
@@ -1,5 +1,5 @@
 ---
-nginx_ppa: "ppa:nginx/mainline"
+nginx_ppa: "ppa:ondrej/nginx-mainline"
```
TangRufus, line 2: Why not the official one?

See: https://www.nginx.com/resources/wiki/start/topics/tutorials/install/#official-debian-ubuntu-packages and https://www.ubuntuupdates.org/ppa/nginx_from_nginx.org?dist=focal

roles/python_interpreter/tasks/main.yml
```diff
@@ -1,11 +1,11 @@
...
-    register: ubuntu_release
+    raw: lsb_release -rs
+    register: ubuntu_version
     changed_when: false
-  - name: Set ansible_python_interpreter for Ubuntu 18.04 Bionic
+  - name: Set ansible_python_interpreter for Ubuntu >= 18.04
     set_fact:
       ansible_python_interpreter: python3
-    when: ubuntu_release.stdout | trim == 'bionic'
+    when: ubuntu_version.stdout | trim is version('18.04', '>=')
```
TangRufus, line 9: How about we drop support for 18.04?
Then, we don't need to worry about python 2

roles/nginx/defaults/main.yml
swalkinshaw: Well we were using mainline (development) and `ppa:nginx/mainline` doesn't offer focal builds. Do the ones you listed support mainline? I couldn't see any mentions

roles/python_interpreter/tasks/main.yml
swalkinshaw: 🤔 it would simplify this, but we still might want to support python 2 locally? So most of the work to make Trellis python compatible would remain if we want to keep that local support.

roles/nginx/defaults/main.yml
TangRufus: > If you would like to use mainline nginx packages, run the following command instead:
> 
> `echo "deb http://nginx.org/packages/mainline/ubuntu `lsb_release -cs` nginx" \
    | sudo tee /etc/apt/sources.list.d/nginx.list`
>
> — https://nginx.org/en/linux_packages.html

`focal` repo is under http://nginx.org/packages/mainline/ubuntu/dists/focal/

Note: 
- I haven’t tested it on Ubuntu 20.04
- the ansible task will need changes; similar to the mariadb one

swalkinshaw: Oh awesome, thanks. Updated 😄

ajschmidt8:
any updates on the status of this PR? i'm going to being a new deployment over the next few weeks and was hoping to have `20.04` support 🙂

swalkinshaw:
It's getting close since I've merged almost all parts of it separately already. You could even make the change in https://github.com/roots/trellis/pull/1212 manually (or just override that variable in group_vars) and using 20.04 should work 😄

strarsis:
That's great, I want to install Trellis onto a new server which got Ubuntu LTS 20.x.

swalkinshaw:
@strarsis yeah, I'll do it by this weekend 😓

swalkinshaw:
I released v1.6.0 without changing the default 20.04 first: https://github.com/roots/trellis/releases/tag/1.6.0

1.7.0 will *just* be to make 20.04 the default version for Vagrant at least.

Issue #1199, opened 2020-06-23
when: web_sudoers | bool evaluates to false


nextgenthemes:
* [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)
* [x] This request isn't a duplicate of an existing issue
* [x] I've read the [docs](https://roots.io/documentation/) and followed them (if applicable)
* [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/) community

## Description

I am just wanted to deploy trellis on 20.04 with server setup with this PR #1197 but it failed because on the restart php fpm task.

I noticed that the sudoers entry was missing and then found out that the task is actually skipped. This `| bool` changes come because of deprecation warnings like this right?

> [DEPRECATION WARNING]: evaluating 'web_sudoers' as a bare variable, this behaviour will go away and you might need to add |bool to 
the expression in the future. Also see CONDITIONAL_BARE_VARS configuration toggle. This feature will be removed in version 2.12. 
Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.

But it seems that actually doing what they say is **not** working. At least not for 2.9.9. I removed the `| bool` and the sudoers file gets created as expected.

I think there were other changes recently to get rid of the warnings. The one I removed to get it to work was this #1195

## Steps to reproduce

1. [First step]
2. [Second step]
3. [and so on...]

**Expected behavior:** [What you expect to happen]

**Actual behavior:** [What actually happens]

**Reproduces how often:** [What percentage of the time does it reproduce?]

## Versions

ansible 2.9.9
Trellis: #1197

## Additional information

Any additional information, configuration or data that might be necessary to reproduce the issue.

swalkinshaw:
FIxed by #1200

swalkinshaw:
I searched the codebase and couldn't find any other uses of ` | bool` on lists.

raph-topo:
Thanks for catching and correcting this! Surprisingly, I did not get an error on restart PHP FPM… 😯

PR #1200, opened 2020-06-23
Fix `| bool` test for lists

nextgenthemes:
Regarding #1199 but there may be other places where a `list` type got a `| bool` added that needs fixing.

I guess this is the way to do it. Not an Ansible expert though. Could also add another check if the item is actually not empty and a string or something but ...

### DIFF
--- a/roles/users/tasks/main.yml
+++ b/roles/users/tasks/main.yml
@@ -47,7 +47,7 @@
     owner: root
     group: root
     validate: "/usr/sbin/visudo -cf %s"
-  when: web_sudoers | bool
+  when: web_sudoers[0] is defined
 
 - name: Add SSH keys
   authorized_key:


swalkinshaw:
👍 tested this out and works.

`web_sudoers | length > 0 | bool` would also work but this seems fine. Thanks 😄

raph-topo:
Thanks! I wonder if there is any way to do type checking in Ansible… Neither https://github.com/redhat-developer/vscode-yaml nor https://github.com/VSChina/vscode-ansible are of any help for this…

nextgenthemes:
@raph-topo You can do a type check with ` | debug_type` or something like that, I can't find it again right now. I did this to find this type is called `list` in yaml or ansible but its basically like an array. But knowing the type here is not helping. I do not know why yaml or Ansible behaves this way. In php `array( 'somethign' ) == true` probably most languages are treating it similar. I think it makes no sense but I not even know if this is yml or ansible and what the reasoning is for it to behave this way. Maybe it should be files as bug/feature request.

PR #1201, opened 2020-06-25
Refactor: Move creation and copying of .env-file to seperate file

schuhwerk:
I use multiple instances of trellis (for different servers and with different teams) on the same codebase (wordpress).
Every time I boot a local VM with a different url ( a.test -> b.test ) I need to change the .env-file (in the root of the codebase).

With the "change" in this request I can just execute the needed dotenv-tasks (without a trellis-cli dependency) like:

```yml
- name: "Env"
  hosts: web:&development
  become: yes
  remote_user: vagrant

  tasks:
  - import_role:
      name: wordpress-install
      tasks_from: dotenv
```

### DIFF
--- a/roles/wordpress-install/tasks/dotenv.yml
+++ b/roles/wordpress-install/tasks/dotenv.yml
@@ -1,0 +1,16 @@
+---
+- name: Create .env file
+  template:
+    src: "env.j2"
+    dest: "/tmp/{{ item.key }}.env"
+    owner: "{{ web_user }}"
+    group: "{{ web_group }}"
+  with_dict: "{{ wordpress_sites }}"
+
+- name: Copy .env file into web root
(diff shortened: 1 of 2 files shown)


swalkinshaw:
Interesting use case! This change is very minor (basically just adding a tag to two tasks) so I'm good with adding this to help support your case.

swalkinshaw:
@schuhwerk couldn't you also just run this new tag directly without the playbook you showed? ie: `ANSIBLE_TAGS=dotenv vagrant provision`

schuhwerk:
Thank you! True, your suggestion is even better :) I tried `ansible-playbook dev.yml --tags "dotenv"` on the VM (but must have gotten something wrong, because i thought the tasks after were executed, too...)

PR #1202, opened 2020-07-09
Check hooked deploy scripts exist before including

dalepgrant:
Making use of the suggested method to [include deploy scripts per site](https://github.com/roots/trellis/blob/master/roles/deploy/defaults/main.yml#L69) requires you to have a script set up for every site:
```
deploy_build_after:
  - "{{ playbook_dir }}/roles/deploy/hooks/build-after.yml"
  # - "{{ playbook_dir }}/deploy-hooks/sites/{{ site }}-build-after.yml" <-- this line
```
If you set up a deploy hook for domain1.com (e.g. `domain1.com-build-after.yml`), you won't be able to deploy to domain2.com, unless you also create `domain2.com-build-after.yml`. 

This PR checks if the file exists before trying to include it, so that if you only need one site to have additional tasks run, you don't need to create unnecessary files. 

For now, it only covers:
- `deploy_build_before`
- `deploy_build_after`

I wanted to run this by you first before going and updating the rest of them: I'm very happy to do it, but thought it's worth checking if a) this is wanted and b) if there's a better way, first.

### DIFF
--- a/roles/deploy/tasks/build.yml
+++ b/roles/deploy/tasks/build.yml
@@ -1,8 +1,14 @@
 ---
-- include_tasks: "{{ include_path }}"
+- name: Check if deploy_build_before scripts exist
+  local_action: stat path="{{ item }}"
+  register: deploy_build_before_paths
   with_items: "{{ deploy_build_before | default([]) }}"
+
+- include_tasks: "{{ include_path.item }}"
+  with_items: "{{ deploy_build_before_paths.results }}"
   loop_control:
     loop_var: include_path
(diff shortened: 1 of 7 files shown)


swalkinshaw:
Thanks for bringing this up; agreed it makes sense. Unfortunately it's a lot more verbose with more tasks... but I can't think of another way around this. Feel free to do this for all include tasks 👍

dalepgrant:
@swalkinshaw Done!

swalkinshaw:
Thank you! 👏

PR #1203, opened 2020-07-10
Ansible `file` module expects a string for `mode`

raph-topo:
https://docs.ansible.com/ansible/latest/modules/file_module.html#parameter-mode
Comments say that octal numbers are tolerated, but parameter asks for string only, and so do linters…


swalkinshaw:
Yeah this is better 👍 thanks

PR #1207, opened 2020-07-21
Reloading Nginx even if some Let's Encrypt certificates renewals has failed

brenoalvs:
Sometimes you can have just one or a few of your sites failing in the Let's Encrypt certificate renewal process.
When this occurs, if one of the others sites have its certificate renewed Nginx will not be reloaded which may break things.

This PR is a proposal to fix this issue by reloading Nginx anyways.

Another option could be modifying how the renewal script will terminate in these cases.
But I can't see any huge difference by doing this instead.

### DIFF
--- a/roles/letsencrypt/tasks/main.yml
+++ b/roles/letsencrypt/tasks/main.yml
@@ -8,7 +8,7 @@
     cron_file: letsencrypt-certificate-renewal
     name: letsencrypt certificate renewal
     user: root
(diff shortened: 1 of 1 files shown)


swalkinshaw:
Good idea 👍  Honestly I just default to using `&&` so mostly an oversight I'm guessing.

brenoalvs:
Leaving some credits to @viniciusrezende
We've fixed this in our Trellis fork a long time ago, now I'm upgrading to the latest version and I've remembered to share.

PR #1208, opened 2020-07-23
Use official Nginx package

swalkinshaw:
Extracted from https://github.com/roots/trellis/pull/1197

Note: this will *not* re-install Nginx if you re-provision an already provisioned server since it's the same apt package name.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+### HEAD
+* Switch to official Nginx Ubuntu package ([#1208](https://github.com/roots/trellis/pull/1208))
+
 ### 1.5.0: August 5th, 2020
 * Improve Nginx reloading for failed Let's Encrypt certificates ([#1207](https://github.com/roots/trellis/pull/1207))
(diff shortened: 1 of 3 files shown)


TangRufus:
Question: What is the procedure to reinstall Nginx with the official ppa?

swalkinshaw:
You'd have to manually uninstall via `apt-get remove nginx` *after* running these changes; then provision again to install from the new package. I don't know of a better/different way.

strarsis:
Would a force-reinstall make sense? https://stackoverflow.com/questions/34260658/how-do-i-force-reinstall-a-package-with-ansible

swalkinshaw:
@strarsis probably, but I'm wary of coding that by default into Trellis. Especially since it involves an uninstall + reinstall which means downtime.

mockey:
Setting `nginx_package_state: latest` worked for me in re-provisioning.

strarsis:
@mockey: Even better, when a specific nginx release is used that is from the official PPA, so the nginx version is locked in, without installing just the latest available one.

swalkinshaw:
`latest` works in this specific case... but I think it could lead to unintended upgrades or breaking changes?

PR #1209, opened 2020-07-23
Update local_action to delegate_to

swalkinshaw:
fyi @dalepgrant 

Not sure why CI didn't notice this before during linting


dalepgrant:
Thanks for the heads up @swalkinshaw, will look out for that in future 👍

PR #1212, opened 2020-08-07
Update MariaDB to 10.5

swalkinshaw:
This bumps the MariaDB PPA from 10.2 to 10.5.

Note that this will *not* actually upgrade the version of MariaDB already installed on a server.

Upgrading is a manual process that involves:

1. stopping the service
2. removing the installed package
3. re-installing the package

The full process is detailed at https://mariadb.com/kb/en/upgrading-from-mariadb-104-to-mariadb-105/

Though it's not required, backing up your database before the upgrade is recommended.

Note: this is split out from the [Ubuntu 20.04](https://github.com/roots/trellis/pull/1197) PR. Upgrading to 10.5 means we can keep the same versions across 18.04 and 20.04.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Update MariaDB package to 10.5 ([#1212](https://github.com/roots/trellis/pull/1212))
 * Switch to official Nginx Ubuntu package ([#1208](https://github.com/roots/trellis/pull/1208))
 
 ### 1.5.0: August 5th, 2020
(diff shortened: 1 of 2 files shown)


swalkinshaw:
@TangRufus this is the same as the Nginx PPA switch in terms of manual upgrades. I don't love *not* upgrading people, but I don't really think that's an option either since there's many steps and downtime.

TangRufus (approved):

swalkinshaw:
CI failure unrelated. Will fix on master

PR #1214, opened 2020-08-15
Drop version from PHP package names

jeffbyrnes:
Excepting for the OPcache package, all of the other PHP packages on Ubuntu can be installed without specifying the PHP minor version. This will always install the latest version of PHP available from the provided APT repositories, at time of provisioning.

### DIFF
--- a/roles/php/defaults/main.yml
+++ b/roles/php/defaults/main.yml
@@ -2,18 +2,18 @@ disable_default_pool: true
 memcached_sessions: false
 
 php_extensions_default:
-  php7.4-cli: "{{ apt_package_state }}"
-  php7.4-common: "{{ apt_package_state }}"
-  php7.4-curl: "{{ apt_package_state }}"
-  php7.4-dev: "{{ apt_package_state }}"
-  php7.4-fpm: "{{ apt_package_state }}"
-  php7.4-gd: "{{ apt_package_state }}"
-  php7.4-mbstring: "{{ apt_package_state }}"
-  php7.4-mysql: "{{ apt_package_state }}"
+  php-cli: "{{ apt_package_state }}"
+  php-common: "{{ apt_package_state }}"
+  php-curl: "{{ apt_package_state }}"
+  php-dev: "{{ apt_package_state }}"
+  php-fpm: "{{ apt_package_state }}"
+  php-gd: "{{ apt_package_state }}"
+  php-mbstring: "{{ apt_package_state }}"
+  php-mysql: "{{ apt_package_state }}"
   php7.4-opcache: "{{ apt_package_state }}"
-  php7.4-xml: "{{ apt_package_state }}"
-  php7.4-xmlrpc: "{{ apt_package_state }}"
-  php7.4-zip: "{{ apt_package_state }}"
+  php-xml: "{{ apt_package_state }}"
+  php-xmlrpc: "{{ apt_package_state }}"
+  php-zip: "{{ apt_package_state }}"
 
 php_extensions_custom: {}
 php_extensions: "{{ php_extensions_default | combine(php_extensions_custom) }}"


TangRufus:
The version in the PHP package names are needed to archive some level of idempotence.

For example, the server was first provisioned with PHP 5.6. When PHP 7.0 is out and become apt package alias, we don't want Trellis to update PHP without notices. Same goes to minor version upgrades as well.

What do you think @swalkinshaw ?

jeffbyrnes:
tl;dr Package versions may be a better way to ensure versions, instead of package names.

* * *

Using Trellis, the server would be provisioned with PHP 7.4 no matter what. That’s ensured by the repo used for PHP packages (Ubuntu 18.04’s repos dictate PHP 7.3, officially).

Currently the “unversioned” PHP packages refer to whatever the repo maintainers decide to use, as virtual packages (i.e., aliases).

Rather than use the package names as a gate for the version, it would seem more beneficial to me to use the package _version_ to ensure what is installed.

This also has the effect of avoiding drift if you install these packages at different times on different servers.

Right now, there’s nothing that prevents us from provisioning a server today, getting PHP 7.4.1, and later, provisioning another one, and while the first server remains at PHP 7.4.1, the new one gets PHP 7.4.5 if that’s the current version, b/c:

https://github.com/roots/trellis/blob/14e26cd6010c52c78a6eace122499d97f887091c/group_vars/all/main.yml#L5

That `present` means once we’ve installed these packages, they remain at those versions until manually updated via `apt upgrade`.

swalkinshaw:
@jeffbyrnes do you mean to use `php-cli=7.4.0` (or whatever) to specify the version?

> Using Trellis, the server would be provisioned with PHP 7.4 no matter what. That’s ensured by the repo used for PHP packages (Ubuntu 18.04’s repos dictate PHP 7.3, officially).

🤔 we use a 3rd party PPA though (`ondrej/php`) though off the top of my head I don't know if there's different guarantees based on the Ubuntu version?

jeffbyrnes:
I do, yes, mean specifying in that manner, since it’s more guaranteed, and also ensures there’s no “creep” of the version if you provision a new box later (i.e., you know you’re running the _exact_ same version across your fleet).

So yeah, the PPA is why we can get PHP 7.4 on Ubuntu 18.04 (b/c Ubuntu only provides 7.3 on 18.04, and 7.4 on 20.04).

I’m not sure what their policy is w/r/t which PHP version they provide via the repo, though I expect it’s mildly conservative.

If the PPA doesn’t provide unsuffixed packages, this is obviously a non-starter, though my point about using versions, not names, stands.

jeffbyrnes:
A gentle bump to consider this further. No hard feelings if it’s not how y’all wanna manage this & simply want to close it!

swalkinshaw:
Thanks for the bump @jeffbyrnes. Honestly, I still don't entirely understand the scope and effects of this change well enough and haven't had the time to explore it more. I'll try to give it some more time over the holidays.

jeffbyrnes:
@swalkinshaw all good! Yeah, just off the top of my head, some potential impacts/interactions:

* This may sidestep the PPA and use the Ubuntu 20.04 PHP packages, which are now meta-packages w/o versions
* If it does still use the PPA, it may mean skipping to PHP 8.x, which was released recently, and is unlikely to be something we want just yet

When I wrote this, I didn’t realize PHP 8.0.0 was so close to release, so it made more sense. It’s probably simpler to close this & leave it as-is.

strarsis:
Is this related?: https://github.com/roots/trellis/pull/1284

Though I am more for locking the PHP version as there could be subtle differences and bugs one may want to avoid.

swalkinshaw:
Somewhat related I guess, but doesn't solve the main issue brought up in this PR. It just makes the versions _more_ obvious in code, but doesn't change the behaviour.

PR #1215, opened 2020-08-29
Remove default Vagrant SMB credentials

LachlanArthur:
This PR removes the fallback values for the SMB username and password, allowing Vagrant to prompt for the credentials during `vagrant up`.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -70,8 +70,8 @@ Vagrant.configure('2') do |config|
 
   extra_options = if vagrant_mount_type == 'smb'
     {
-      smb_username: vconfig.fetch('vagrant_smb_username', 'vagrant'),
-      smb_password: vconfig.fetch('vagrant_smb_password', 'vagrant'),
+      smb_username: vconfig.fetch('vagrant_smb_username', nil),
+      smb_password: vconfig.fetch('vagrant_smb_password', nil),
     }
   else
     {}


swalkinshaw:
Can you explain a bit more about the use case and problem this is solving?

You can specify those config values in https://github.com/roots/trellis/blob/master/vagrant.default.yml (or a local version of it) if you want to override them.

LachlanArthur:
The default experience after setting the synced folder mode to SMB is poor: Vagrant doesn't ask for credentials (because the username and password have already been set to the fallback value `vagrant`) and so the mount command fails with a generic "permission denied" error message:

<details>
<summary>Log</summary>

```
$ vagrant reload
==> default: Attempting graceful shutdown of VM...
    default: Configuring the VM...
    default: Setting VM Integration Services
==> default: time_synchronization is enabled
    default: Setting VM Enhanced session transport type to disabled/default (VMBus)
==> default: Starting the machine...
==> default: Waiting for the machine to report its IP address...
    default: Timeout: 120 seconds
    default: IP: 172.29.96.221
==> default: Waiting for machine to boot. This may take a few minutes...
    default: SSH address: 172.29.96.221:22
    default: SSH username: vagrant
    default: SSH auth method: private key
==> default: Machine booted and ready!

Vagrant requires administrator access for pruning SMB shares and
may request access to complete removal of stale shares.
==> default: Preparing SMB shared folders...

Vagrant requires administrator access to create SMB shares and
may request access to complete setup of configured shares.
==> default: Setting hostname...
==> default: Mounting SMB shared folders...
    default: C:/Users/Lachlan/Source/wordpress => /srv/www/wordpress/current
Failed to mount folders in Linux guest. This is usually because
the "vboxsf" file system is not available. Please verify that
the guest additions are properly installed in the guest and
can work properly. The command attempted was:

mount -t cifs -o vers=3.02,credentials=/etc/smb_creds_vgt-908aead4cd9bdec2cbb079cf486c232c-a50889c11b9a802b31b270d15406cd26,uid=1000,gid=33,mfsymlinks,dir_mode=0776,file_mode=0775,sec=ntlmssp //10.0.75.1/vgt-908aead4cd9bdec2cbb079cf486c232c-a50889c11b9a802b31b270d15406cd26 /srv/www/wordpress/current

The error output from the last command was:

mount error(13): Permission denied
Refer to the mount.cifs(8) manual page (e.g. man mount.cifs)
```

</details>

It's true that the SMB username and password can be set in `vagrant.local.yml` using `vagrant_smb_username` and `vagrant_smb_password`, but this is not documented anywhere:

https://github.com/roots/trellis/search?q=vagrant_smb_password&unscoped_q=vagrant_smb_password
https://github.com/roots/docs/search?q=vagrant_smb_password&unscoped_q=vagrant_smb_password

I don't see a good reason for having fallback values for the credentials, as they will always be incorrect and their existence prevents Vagrant from asking the user to provide them.

swalkinshaw:
I might be missing something here... but what should the credentials be? How do you know what they are?

I guess on Windows its just your user name + password? 🤔 maybe these default values are more useful on macOS. Though I'm assuming most smb users are on Windows anyway... it might make sense to optimize for that.

So we could merge this PR, and then document setting the credentials in the git ignored `vagrant.local.yml`.

LachlanArthur:
You are correct, the credentials are the username and password of the host machine. They are required for the guest to authenticate to the host.

I don't use macOS, so I'm not sure if credentials are required (or prompted for) to connect to a share on a macOS host. I'll test it out on one of the macs at work tomorrow.

swalkinshaw:
I tried changing the mount type to `smb` on macOS and got this:

```plain
==> default: Preparing SMB shared folders...
Vagrant SMB synced folders require the account password to be stored
in an NT compatible format. Please update your sharing settings to
enable a Windows compatible password and try again.
```

So... I'll go ahead and merge this. We can make tweaks after if someone is using smb on Mac and has issues.

PR #1220, opened 2020-09-11
Move site_env to defaults

### LINKED ISSUE #1218: site_env in deploy/vars/main.yml overwrites provided vars

## Description

In `roles/deploy/vars/main.yml`, `site_env` is defined is basically the exact same way it's defined in `deploy.yml`. However, because it's defined in `vars`, it "wins" when creating the `.env` file.

I'm using Trellis alongside a Gatsby site and am borrowing the `deploy` role to do so. And it mostly works! The role gets all the files into place and the hooks do the app-specific logic (in my case, running the build, hence needing the .env file–yes, I've installed node on the server to do this). But the wrong .env file is created cuz I've set a custom `site_env` in my playbook that gets overriden.

## Steps to reproduce

1. Create a new playbook using the `deploy` role and a custom `site_env`.
2. Run the deployment
3. Note that the `.env` file created uses the `site_env` from the role.

**Expected behavior:** [What you expect to happen]

`site_env` from my playbook should overwrite the role's `site_env`.

**Actual behavior:** [What actually happens]

`site_env` from the role is used.

**Reproduces how often:** [What percentage of the time does it reproduce?]

Always.

## Versions

I'm running Trellis 1.3.0 but it's def present in master.

## Additional information

The fix is fairly easy: remove or move to `defaults`. Happy to open a PR but wondering which you'd prefer. There's also a bit of duplication in this area (`wordpress_env_defaults` is also present in box the `deploy.yml` & `deploy` roles, and overridden in the same way).

mAAdhaTTah:
This allows consumers of the deploy role to provide their own site_env.

Fixes #1218,

### DIFF
--- a/roles/deploy/defaults/main.yml
+++ b/roles/deploy/defaults/main.yml
@@ -57,7 +57,7 @@ composer_classmap_authoritative: true
 project: "{{ wordpress_sites[site] }}"
 project_root: "{{ www_root }}/{{ site }}"
 project_local_path: "{{ (lookup('env', 'USER') == 'vagrant') | ternary(project_root + '/' + project_current_path, project.local_path) }}"
-
+site_env: "{{ wordpress_env_defaults | combine(vault_wordpress_env_defaults | default({}), project.env | default({}), vault_wordpress_sites[site].env) }}"
 
 # Deploy hooks
 # For list of hooks and explanation, see https://roots.io/trellis/docs/deploys/#hooks
--- a/roles/deploy/vars/main.yml
+++ b/roles/deploy/vars/main.yml
@@ -10,5 +10,3 @@ wordpress_env_defaults:
   git_sha: "{{ git_clone.after }}"
   release_version: "{{ deploy_helper.new_release }}"
   wp_debug_log: "{{ project_root }}/logs/debug.log"
-
-site_env: "{{ wordpress_env_defaults | combine(vault_wordpress_env_defaults | default({}), project.env | default({}), vault_wordpress_sites[site].env) }}"


swalkinshaw:
🤔 so the original PR mentions it was put in `vars` instead of `defaults` due to variable precedence: https://github.com/roots/trellis/pull/820

We might have to dig into this more or do some testing

mAAdhaTTah:
Ahhh, I didn't see `site_env` in `helpers.yml`. Looking there, I see it takes advantage of `item.key` & `item.value`. What loop is responsible for setting this? My thought is that loop could set the `loop_vars` so the values used in `helpers.yml` are consistent.

swalkinshaw:
It is pretty weird... I'm sure they are initialized/defined at boot time (or thereabouts) but since they contain Jinja templates those will get executed in the context of the task. Otherwise Trellis wouldn't work at all 😄

mAAdhaTTah:
@swalkinshaw that is fascinating. I hadn't really thought about it but that makes way more sense.

Thought: Move `group_vars/all/helpers.yml` into `wordpress-setup/defaults`, similar to the helper vars moved to `deploy/defaults` in this PR? Then all of the helper vars are associated with the role that uses them and can be overridden as needed.

swalkinshaw:
@mAAdhaTTah it sounds fine but we'd have to test to make sure all the precedents make sense

mAAdhaTTah:
If this isn't useful to anyone else, then I'm going to close this. I moved the static frontend to DigitalOcean's new App Platform and keep everything a bit more separated.

swalkinshaw:
👍 probably not high priority enough to deal with the unknown side effects of this

PR #1227, opened 2020-10-07
Separate 'Enable or disable Nginx sites' task into two 

### LINKED ISSUE #1225: "Enable or disable Nginx sites" task doesn't work with ansible 2.10.1

## Description

"Enable or disable Nginx sites" task in wordpress-setup role doesn't work with ansible 2.10.1

## Steps to reproduce

1. Provision

**Expected behavior:** Finish provisioning

**Actual behavior:** Error: src option requires state to be 'link' or 'hard'

**Reproduces how often:** Always

## Versions
trellis: HEAD
OS: Linux Mint 20 Cinnamon
ansible 2.10.1
vagrant 2.2.10

The problem seems to be: 
`state: "{{ item.enabled | default(true) | ternary('link', 'absent') }}"`
According to the docs:
https://docs.ansible.com/ansible/latest/collections/ansible/builtin/file_module.html#parameter-src
only `link` or `hard` are allowed as state when a src parameter is defined.

I created two separate tasks for enable and disable using `when: item.enabled | default(true)` for enable and `when: not item.enabled | default(false)` for disable. Seems to work but I'm not sure if that's the best solution. I can make a PR if you want.

TangRufus:
Fix #1225

Revert #1226

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
-ansible>=2.8.0,<2.10.0
+ansible>=2.8.0,<3.0
 passlib
--- a/roles/wordpress-setup/tasks/nginx.yml
+++ b/roles/wordpress-setup/tasks/nginx.yml
@@ -28,12 +28,22 @@
   notify: reload nginx
   tags: nginx-sites
 
-- name: Enable or disable Nginx sites
+- name: Enable Nginx sites
   file:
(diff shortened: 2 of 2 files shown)


swalkinshaw:
Makes sense to me 👍

TangRufus:
Removed `src: xxx` and `force: yes` from the `Disable Nginx sites` task.
Please re-review @swalkinshaw

swalkinshaw (approved):

PR #1235, opened 2020-10-14
Fix ANSI color escape sequence

strarsis:
Edit: Or is the actual issue that I used python v2 instead v3 for running ansible?
Ansible complains about an incorrect sequence "bright gray".
Apparently there is no "bright gray", but "bright black" for greyish colors.
After replacing "bright gray" with "bright black", the error goes away.

### DIFF
--- a/lib/trellis/utils/output.py
+++ b/lib/trellis/utils/output.py
@@ -102,17 +102,17 @@ def display(obj, result):
     hr = '-' * int(wrap_width*.67)
 
     if obj.task_failed and first:
-        display(system(obj.vagrant_version), 'bright gray')
-        display(hr, 'bright gray')
+        display(system(obj.vagrant_version), 'bright black')
+        display(hr, 'bright black')
 
     if msg == '':
         if obj.task_failed and not first:
-            display(hr, 'bright gray')
+            display(hr, 'bright black')
         else:
             return
     else:
         if not first:
-            display(hr, 'bright gray')
+            display(hr, 'bright black')
         display(msg, 'red' if obj.task_failed else 'bright purple')
 
 def display_host(obj, result):


swalkinshaw:
🤔 

I found the defined colours at https://github.com/ansible/ansible/blob/3e9943bc5e7a9cd393757aa8100d7fed80bd316e/lib/ansible/utils/color.py#L57-L68

```plain
❯ trellis exec python
Python 3.7.7 (default, Mar 10 2020, 15:43:33)
[Clang 11.0.0 (clang-1100.0.33.17)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from ansible.utils.color import stringc
>>> stringc("test", "blue")
'\x1b[0;34mtest\x1b[0m'
>>> stringc("test", "bright gray")
'\x1b[0;37mtest\x1b[0m'
>>> stringc("test", "bright black")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.7/site-packages/ansible/utils/color.py", line 92, in stringc
    color_code = parsecolor(color)
  File "/usr/local/lib/python3.7/site-packages/ansible/utils/color.py", line 77, in parsecolor
    return codeCodes[color]
KeyError: 'bright black'
```

I don't see `bright black` on any Ansible version (2.8-2.10).

swalkinshaw:
What's the actual output you get with bright gray?

strarsis:
@swalkinshaw: Actually it was a different issue, the renew certificate script printed an error message that is confusing for the current trellis code.

strarsis:
@swalkinshaw: Underlying issue: https://github.com/roots/trellis/issues/1236

strarsis:
For some reason the color codes used by Trellis sometimes cause an exception within Ansible/Python.
Ansible defines a set of colors that can be used for logging, and shows a color-code related error otherwise, 
without showing the actual logged error.

(Ansible `v2.9.10` and `3.8.5` used)
https://github.com/ansible/ansible/blob/v2.9.10/lib/ansible/config/base.yml#L227
https://github.com/ansible/ansible/blob/v2.9.10/lib/ansible/utils/display.py#L214

Ansible apparently doesn't like the terminal color `bright gray`:
`Invalid color supplied to display: bright gray`

swalkinshaw:
That makes sense it looks like. We'll have to pick another colour that Ansible actually maps to a log level.

PR #1246, opened 2020-10-24
Set universal_newlines for renew-certs script

swalkinshaw:
Using `universal_newlines` is better for Python 2/3 compatibility since it will return a string and not bytes to ensure it's the same regardless of the version.


TangRufus:
Should we move to `run` instead?
See: https://stackoverflow.com/a/4760517

swalkinshaw:
Looks like that's python 3 only? Unfortunately this script is supposed to run in python 2 only (though apparently it may not, hence this PR).

 See https://github.com/roots/trellis/pull/1060 but it's confusing 😕

TangRufus (approved):

PR #1256, opened 2020-12-21
Fix security issue with empty password

arusa:
If no password is defined for user "web", insert "!" to disable authentication without a password.

### DIFF
--- a/roles/users/tasks/main.yml
+++ b/roles/users/tasks/main.yml
@@ -33,7 +33,7 @@
     name: "{{ item.name }}"
     group: "{{ item.groups[0] }}"
     groups: "{{ item.groups | join(',') }}"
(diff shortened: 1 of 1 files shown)


swalkinshaw:
🤔 Can you describe what the problem is? Or how you can take advantage of this security issue?

```plain
vagrant@example:~$ su web
No passwd entry for user 'web'
```

arusa:
According to the shadow man page it could be that some applications may allow access without a password.
I'm not sure how to exactly take advantage of this, but "man shadow" says:

> If the password field contains some string that is not a valid result of crypt(3), for instance ! or *, the user will not be able to use a unix password to log in (but the user may log in the system by other means).

> This field may be empty, in which case no passwords are required to authenticate as the specified login name. However, some applications which read the /etc/shadow file may decide not to permit any access at all if the password field is empty.

> A password field which starts with an exclamation mark means that the password is locked. The remaining characters on the line represent the password field before the password was locked.

swalkinshaw:
Thanks @arusa, makes sense to me 👍

PR #1260, opened 2021-01-01
Migrate to Xdebug v3

### LINKED ISSUE #1255: Upgrading from Xdebug 2 to 3
* [x] This request isn't a duplicate of an existing request

## Summary

Thanks for all the great work on Trellis.

Xdebug 3 is out and includes major changes in functionality and configuration compared to Xdebug 2. Trellis ships with default support for Xdebug 2 style options, no default support for Xdebug 3 style config options yet.

## Motivation

Why are we doing this? 

Provide default support for Xdebug 3 config options.

What use cases does it support? 

Debugging with Xdebug.
 
What is the expected outcome?

Default Xdebug 3 config options set in `roles/xdebug/defaults/main.yml`

## Additional context

[Xdebug 3 Upgrade Guide](https://xdebug.org/docs/upgrade_guide)

adambrgmn:
This PR will migrate the Xdebug config to work with Xdebug v3 that [was released in november 2020](https://xdebug.org/announcements/2020-11-25).

It's a pretty naive migration where I simply went thru the [upgrade guide](https://xdebug.org/docs/upgrade_guide) and removed the config options that's been deprecated and updated those that have been changed.

Though there is one significant change:

With Xdebug v3 they've changed the default connection port from `9000` to `9003`. I decided to follow the defaults and this means the new port to connect with trellis+xdebug will be `9003`.

I would love some help with this one. I'm not in any way experienced in configuring Xdebug. The changes made here does work well with the php debugger in VScode. But I haven't tested it with other IDE's. So if someone with more experience in configuring Xdebug can take a look I would warmly welcome some feedback.

## Considerations

- Does this migration in any way affect the xdebug-tunnel role? I've never used it, but maybe changes need to happen there as well?

Fixes #1255

### DIFF
--- a/roles/xdebug/tasks/main.yml
+++ b/roles/xdebug/tasks/main.yml
@@ -19,13 +19,13 @@
       state: link
     notify: reload php-fpm
 
-  when: xdebug_remote_enable | bool
+  when: xdebug_mode is not match("off")
 
 - name: Disable Xdebug
   file:
     path: /etc/php/7.4/fpm/conf.d/20-xdebug.ini
     state: absent
-  when: not xdebug_remote_enable | bool
+  when: xdebug_mode is match("off")
   notify: reload php-fpm
 
 - name: Disable Xdebug CLI
--- a/roles/xdebug/tasks/main.yml
+++ b/roles/xdebug/tasks/main.yml
@@ -19,15 +19,6 @@
       state: link
     notify: reload php-fpm
 
-  when: xdebug_mode is not match("off")
-
-- name: Disable Xdebug
-  file:
-    path: /etc/php/7.4/fpm/conf.d/20-xdebug.ini
-    state: absent
-  when: xdebug_mode is match("off")
-  notify: reload php-fpm
-
 - name: Disable Xdebug CLI
   file:
     path: /etc/php/7.4/cli/conf.d/20-xdebug.ini
--- a/group_vars/development/php.yml
+++ b/group_vars/development/php.yml
@@ -5,5 +5,6 @@ php_track_errors: 'On'
 php_mysqlnd_collect_memory_statistics: 'On'
 php_opcache_enable: 0
 
-xdebug_remote_enable: 1
-xdebug_remote_connect_back: 1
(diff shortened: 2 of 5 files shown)


swalkinshaw:
I'm guessing at a minimum the ports for the tunnel need updating as well: https://github.com/roots/trellis/blob/master/roles/xdebug-tunnel/defaults/main.yml

Want to update those to be `9003`? I don't that much about the tunnel feature, but from a quick look at the rest the role's code, there shouldn't be anything to update.

adambrgmn:
@swalkinshaw sorry for the delay. But I updated the ports on the xdebug-tunnel role as well. Please tell me if you think something more needs to be handled – otherwise, from end, it's good to go 👍

roles/xdebug/tasks/main.yml
```diff
@@ -19,13 +19,13 @@
       state: link
     notify: reload php-fpm
 
-  when: xdebug_remote_enable | bool
+  when: xdebug_mode is not match("off")
```
swalkinshaw, line 22: 🤔  their docs for `off` says:

> Nothing is enabled. Xdebug does no work besides checking whether functionality is enabled. Use this setting if you want close to 0 overhead.

Maybe we can just *always* template the config file? If it exists but the mode is `off`, then nothing is enabled anyway. Seems much easier than removing/adding the file.

adambrgmn: That is probably a good idea. In 04d4bc3 I also skipped removing the conf file if `mode == "off"`. Not sure what impact that will have on prod environments?

adambrgmn:
Sorry if this is just noise @swalkinshaw, but just wanted to make sure you've seen my updates? There is no rush from my end!

roles/xdebug/tasks/main.yml
swalkinshaw: The xdebug role is only included in `dev.yml` and not the production `server.yml` so we should be fine here 😄

swalkinshaw:
@adambrgmn thanks so much for this 🎉

PR #1261, opened 2021-01-10
Add config for PHP CLI

akovalyov:
Currently PHP CLI configuration is not handled by Trellis, this PR aims to fix it. 

Usecase: emails, which are being processed via cronjobs are getting lost as they are being sent from background tasks, where default sendmail is configured instead of ssmtp.

Web:
image
CLI:
image


swalkinshaw (approved):
Good idea, thanks 🎉

roles/php/tasks/main.yml
```diff
@@ -45,8 +45,13 @@
...
+- name: Copy PHP-FPM configuration file
   template:
-    src: php.ini.j2
+    src: php-fpm.ini.j2
     dest: /etc/php/7.4/fpm/php.ini
   notify: reload php-fpm
+
+- name: Copy PHP CLI configuration file
+  template:
+    src: php-fpm.ini.j2
```
TangRufus, line 56: Should this be?

```diff
-     src: php-fpm.ini.j2 
+     src: php-cli.ini.j2 
```

akovalyov: @TangRufus you're right. I've missed it. Should I submit another PR to fix that? /cc @swalkinshaw ?

TangRufus: Yes, please

akovalyov: @TangRufus done, https://github.com/roots/trellis/pull/1262

PR #1265, opened 2021-01-21
Extend failed connection to remote repo msg

ChrisScrumping:
Due to no_log for git clone on deploy you are unable to see the exact
output error message, adding the branch being used to the fail message
may help when incorrect branch name is used.


swalkinshaw:
Never thought of that, but makes sense 👍 Thanks for the improvement

ChrisScrumping:
No problem, thanks

PR #1266, opened 2021-01-23
Reduce output for letsencrypt tasks

biinari:
Avoid printing out the whole `wordpress_sites` config for letsencrypt tasks.


biinari:
Sorry, opened in the wrong place (hence the unclean changeset). Closing

PR #1270, opened 2021-02-06
Set file permissions explicitly

swalkinshaw:
Fixes ansible-lint violations

PR #1271, opened 2021-02-09
File permissions: Use strings instead of octal numbers

TangRufus:
Follow up #1270
Fix: https://discourse.roots.io/t/mode-must-be-in-octal-or-symbolic-form/20038

See: https://github.com/ansible/ansible/issues/31952#issuecomment-338042101


swalkinshaw (commented):
Computers are fun 🙃

PR #1276, opened 2021-02-16
Update required Ubuntu version in README.md

lukasbesch:


swalkinshaw:
🤔  I believe that was left on purpose since 20.04 isn't technically *required*. Did you run into any issues with 18.04?

lukasbesch:
That makes sense. I updated the README again to clarify that both versions are still supported.

swalkinshaw:
Works for me, thanks 👍

PR #1278, opened 2021-03-06
Allow customizing PHP CLI `memory_limit`

### LINKED ISSUE #1277: PHP CLI memory limit setting

## Description

(Following up on https://discourse.roots.io/t/cant-use-wp-cli-due-to-php-memory-limit/20257/5)

When installing a WP-CLI package by adding it to the wp_cli_packages ansible variable (but the issue could potentially happen with any other php command), I get a `Fatal error: Allowed memory size of 134217728 bytes exhausted`.

I can get rid of the problem by setting the memory_limit to a higher value :
`php -d memory_limit=512M "$(which wp)" package install git@github.com:10up/wp-vulnerability-scanner.git`

But it's not an ideal solution.

I also tried setting php cli memory limit to -1 in `/etc/php/7.4/cli/php.ini`.

```
[PHP]
error_reporting = E_ALL
sendmail_path = /opt/mailhog/mhsendmail
expose_php = Off
date.timezone = Europe/Paris
memory_limit = -1
```
This works, but shouldn't this be set by default during provisioning ?

## Steps to reproduce

Run a command in the development VM , for example :
`wp package install git@github.com:10up/wp-vulnerability-scanner.git`

**Expected behavior:** 
Command works.

**Actual behavior:**
Returns a PHP memory limit fatal error.

## Versions

Trellis version : 1.7.1
Ubuntu 20.04

TangRufus:
Close #1277

---

⚠️  Requires testing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Fix #1277 - Disable PHP CLI memory limit ([#1278](https://github.com/roots/trellis/pull/1278))
 * Fix #1285 - Improve handling of WP-CLI failed verification ([#1295](https://github.com/roots/trellis/pull/1295))
 * Fix #1284 - Update logrotate postrotate Nginx command ([#1293](https://github.com/roots/trellis/pull/1293))
 * Replace php-gd with php-imagick ([#1292](https://github.com/roots/trellis/pull/1292))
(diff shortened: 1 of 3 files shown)


swalkinshaw (approved):

strarsis:
@TangRufus: Because of https://github.com/roots/trellis/issues/1277#issuecomment-853281400, could you hardcode the memory value to `-1` as the default?

PR #1282, opened 2021-03-27
Update to PHP 8.0

clintonb:
This adds support for PHP 8.0, which requires WordPress 5.6+.

### DIFF
--- a/roles/php/tasks/main.yml
+++ b/roles/php/tasks/main.yml
@@ -20,7 +20,7 @@
 - name: Check for existing php7.4-fpm service
   stat:
     path: /etc/init.d/php7.4-fpm
-  register: php73_status
+  register: php74_status
 
 - name: Stop php7.4-fpm service if it exists
   service:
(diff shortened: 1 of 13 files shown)


roles/php/tasks/main.yml
```diff
@@ -1,22 +1,36 @@
...
   service:
-    name: php7.4-fpm
+    name: php8.0-fpm
     state: started
     enabled: true
 
+- name: Check for existing php7.4-fpm service
+  stat:
+    path: /etc/init.d/php7.4-fpm
+  register: php73_status
```
swalkinshaw, line 23: This needs to be updated to `php74` to match the the task below

swalkinshaw (commented):
Thanks for this! Just one minor thing then we should be good

roles/php/tasks/main.yml
clintonb: Fixed.

swalkinshaw:
This made me put some thought into how Trellis handles PHP versions because it's brittle right now and not too flexible.

https://github.com/roots/trellis/pull/1284 is an attempt at improving this and refactoring version support. Assuming it's good, then setting the default to `8.0` will be quite simple.

I'm a little hesitant to bump Trellis to `8.0` by default right now, without easily letting people change back to `7.4`, since I think it might be common for some plugins to have compatibility issues?

clintonb:
See #1284.

PR #1283, opened 2021-03-31
Change default ansible-galaxy --force behaviour

jaisgit:
Only download roles when installed role versions differ from those specified in galaxy.yml


swalkinshaw:
We did this previously and reverted it for reasons. See https://github.com/roots/trellis/pull/1044 and https://github.com/roots/trellis/pull/1064.

🤔 trellis-cli actually offers a *much* better `trellis galaxy install` command: https://github.com/roots/trellis-cli/blob/8e48eb7f48c9e572a8ab2ece88df07a748871896/cmd/galaxy_install.go#L106-L109

We could potentially use that command instead of the normal galaxy one within Vagrant if people have trellis-cli installed. I'd have to think more about that.

PR #1284, opened 2021-04-02
Improve handling of PHP versions

swalkinshaw:
Refactors how PHP and its extensions are installed per version. Previously all version references were hardcoded and updating to a new version (ie from `7.4` to `8.0`) meant replacing a bunch of version numbers across lots of files which made it difficult to use a different version than Trellis' default.

Now to switch to another PHP version that Trellis supports, only the `php_version` variable needs to be changed. And to support a new version, only a single version specific vars file needs to be created (example: `roles/php/vars/8.0.yml`).

### DIFF
--- a/roles/common/tasks/main.yml
+++ b/roles/common/tasks/main.yml
@@ -34,6 +34,18 @@
   when: item.value.site_hosts | rejectattr('canonical', 'defined') | list | count
   tags: [letsencrypt, wordpress]
 
+- name: Import PHP version specific vars
+  include_vars: "{{ lookup('first_found', params) }}"
+  vars:
+    params:
+      files:
+        - '{{ php_version }}.yml'
+        - '7.4.yml'
+      paths:
+        - "{{ playbook_dir }}/roles/php/vars/"
+
+  tags: [php, memcached]
+
 - name: Verify dict format for apt package component variables
   fail:
     msg: "{{ lookup('template', 'package_vars_wrong_format_msg.j2') }}"
--- a/group_vars/all/main.yml
+++ b/group_vars/all/main.yml
@@ -1,8 +1,9 @@
-composer_keep_updated: true
 apt_cache_valid_time: 3600
 apt_package_state: present
 apt_security_package_state: latest
 apt_dev_package_state: latest
+composer_keep_updated: true
+php_version: "7.4"
 ntp_timezone: Etc/UTC
 ntp_manage_config: true
 www_root: /srv/www
--- a/group_vars/all/users.yml
+++ b/group_vars/all/users.yml
@@ -19,4 +19,4 @@ users:
 web_user: web
 web_group: www-data
 web_sudoers:
-  - "/usr/sbin/service php7.4-fpm *"
+  - "/usr/sbin/service php{{ php_version }}-fpm *"
--- a/roles/common/handlers/main.yml
+++ b/roles/common/handlers/main.yml
@@ -9,7 +9,7 @@
 
 - name: reload php-fpm
   service:
-    name: php7.4-fpm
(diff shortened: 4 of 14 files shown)


TangRufus:
This approach looks good to me. 

I foresee [php-xxx.ini](https://github.com/roots/trellis/tree/31fb95030debc8f8a13cde5d61e578d848cf46d3/roles/php/templates) would need refactoring soon or later. 

Discussion:

Whenever a new PHP version is released, there are lots of enquires about when will Trellis support them. I think this is the chance we define our PHP version support policy. 

Things to consider:
- https://www.php.net/supported-versions.php 
    - Do we support versions out of “active support”? <- I say yes to reduce PHP versions we need to support
- https://getcomposer.org/doc/00-intro.md#system-requirements 
    - Not much an issue because composer supports a wide range of PHP versions
- https://wordpress.org/about/requirements/
    -  Do we support older PHP versions not recommended/supported by WordPress?
    - Do we support PHP versions WordPress not yet supported? <— I say yes so that we could test on it.

TangRufus (approved):

swalkinshaw:
> I foresee php-xxx.ini would need refactoring soon or later.

Yeah, though they are pretty stable across versions thankfully. Good news is we can just copy this similar pattern an have version specific .ini templates if need be.

I realized this approach works better for the extensions too since they don't _only_ differ by version numbers. There's been once or twice where the extensions have changed so this makes it easier to support.

> Whenever a new PHP version is released, there are lots of enquires about when will Trellis support them. I think this is the chance we define our PHP version support policy.

Your heuristics make sense to me overall. The one thing we've avoided in the past is defaulting to a brand new PHP version when WP core itself still has compatibility issues (or if lots of plugins do).

This PR adds support for 8.0 without defaulting it which we also couldn't easily do before 🎉

TangRufus:
Note: I made a typo

- https://php.net/supported-versions.php
    * Do we support versions ~~out of~~ **only in** “active support”? <- I say yes to reduce PHP versions we need to support

clintonb (approved):

roles/common/tasks/main.yml
```diff
@@ -34,6 +34,18 @@
...
 
+- name: Import PHP version specific vars
+  include_vars: "{{ lookup('first_found', params) }}"
+  vars:
+    params:
+      files:
+        - '{{ php_version }}.yml'
+        - '7.4.yml'
+      paths:
+        - "{{ playbook_dir }}/roles/php/vars/"
```
swalkinshaw, line 45: Added this as a further fallback/default in case people don't get the update to `group_vars/all/main.yml`.

strarsis:
@swalkinshaw: Though this isn't a big deal, the PHP versions are still hardcoded in some debug output of Trellis:
https://github.com/roots/trellis/blob/master/server.yml#L19
https://github.com/roots/trellis/blob/master/dev.yml#L9

swalkinshaw:
Oh good catch. We can just remove the versions 👍  Do you want to do a PR for that @strarsis ?

strarsis:
@swalkinshaw: Sure. Or should the PHP version variable added to the message?

swalkinshaw:
I think remove them? We output the version later on anyway

strarsis:
@swalkinshaw: PR: https://github.com/roots/trellis/pull/1294

vesper8:
This is great! Is it confirmed anywhere which versions of PHP are supported? For example can I use this with PHP 7.1 and or even 7.0 ?

It would be fantastic if it was possible to do the same thing but when setting whether to use mysql 5.6 or mysql 8 or mariadb. Currently picking one of the 3 is not so difficult but nowhere near as easy as picking the PHP version after this enhancement PR

PR #1108, opened 2019-09-12
Add imagick and bcmath extensions.

nextgenthemes:
The https://wordpress.org/plugins/health-check/ plugin recommends it even thought the page it links to https://make.wordpress.org/hosting/handbook/handbook/server-environment/#php-extensions listing what is recommended by the WP hosting team has bcmath not listed, so not sure what its needed or recommenced for.

PR #1291, opened 2021-05-20
Add php-intl extension

joshuafredrickson:
Per the WordPress Hosting Team recommendation: https://make.wordpress.org/hosting/2021/05/20/why-hosters-should-install-the-php-intl-extension/


retlehs (approved):

retlehs:
nice, thank you!

swalkinshaw:
We should revisit https://github.com/roots/trellis/pull/1108 too

PR #1292, opened 2021-05-21
Add ImageMagick, ghostscript, bc_math and PDF thumbnails

joshuafredrickson:
Addresses #1108


swalkinshaw:
Thank you 🎉  tested and site health all passes

roles/php/vars/7.4.yml
```diff
@@ -3,7 +3,7 @@ php_extensions_default:
   php7.4-curl: "{{ apt_package_state }}"
   php7.4-dev: "{{ apt_package_state }}"
   php7.4-fpm: "{{ apt_package_state }}"
-  php7.4-gd: "{{ apt_package_state }}"
```
TangRufus, line 6: Note: Removing this line doesn't actually uninstall `php7.4-gd`. `php7.4-gd` would linger around if a server is provisioned before this PR.

`php7.4-gd: absent` is required to ensure it is being uninstalled.

Although I don't see any damage if having both `php7.4-gd` and `php7.4-imagick`, hard-to-reproduce tickets might come up in the future.

cc @swalkinshaw

joshuafredrickson: That's a good point. This would only keep `php-gd` from being added during provisioning. I agree that there shouldn't be issues with having `php-gd` and `php-imagick` installed concurrently on existing deployments.

As of 3.5, [WordPress will default to Imagick if available](https://github.com/WordPress/WordPress/blob/086edc3d12b381b129ec906819a5dd4c40b19eee/wp-includes/media.php#L3733) and fallback to GD and adds a filter for overriding this behavior. 

Would having both extensions always available be the more preferable route here?

swalkinshaw: I think this is kind of expected with Trellis by now and is _mostly_ consistent with how we do other package changes. We don't uninstall PHP version packages either for example.

I'm not against adding `php7.4-gd: absent` though.

strarsis:
What about adding the PHP SOAP extension? Some plugins like WooCommerce use/need it.
You may find this PHP extensions table interesting:
https://github.com/strarsis/docker-php-bedrock/blob/main/README.md#php-extensions

strarsis:
@joshuafredrickson: Pardon my ignorance, but I can't find where in this PR the `bc_math` PHP extension is added. 😼

Issue #1285, opened 2021-04-19
Logrotate script gives error; `service nginx rotate` command doesn't exist.


robrecord:
* [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)
* [x] This request isn't a duplicate of an existing issue
* [x] I've read the [docs](https://roots.io/documentation/) and followed them (if applicable)
* [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/) community

## Description

I am getting the following error in my cron emails:

```
/etc/cron.daily/logrotate:
Usage: /etc/init.d/nginx {start|stop|status|restart|reload|force-reload|upgrade|configtest|check-reload}
error: error running shared postrotate script for '"/srv/www/**/logs/*.log"
'
run-parts: /etc/cron.daily/logrotate exited with return code 1
```

## Steps to reproduce

1. Provision server with trellis
2. wait for logrotate

**Expected behavior:** [What you expect to happen]

Logs rotate with no error

**Actual behavior:** [What actually happens]

The error email given above

**Reproduces how often:** [What percentage of the time does it reproduce?]

Unknown, presumed always.

## Versions

Trellis 1.8.0

## Additional information

I dug around, and found that the command `rotate` was removed from the nginx init.d script, as per the following discussion:
https://bugs.launchpad.net/nginx/+bug/1450770/comments/7

This change appears to have been implemented incorrectly: e8d1b4a

I believe the command on line 20 of `group_vars/all/logrotate.yml` should be `service nginx reload >/dev/null 2>&1 `as per this fix: https://bugs.launchpad.net/nginx/+bug/1450770/comments/8

Not sure how to invoke to test completely other than waiting to see if it works. I presume there is a way to force a logrotate?

(I have commented the same report here: https://github.com/roots/trellis/pull/377#commitcomment-49694275 )

I am happy to make a pull request if I'm not barking up the wrong tree.

swalkinshaw:
https://bugs.launchpad.net/nginx/+bug/1450770/comments/8 seems to make sense. A PR would be welcome 👍

swalkinshaw:
Closed by https://github.com/roots/trellis/pull/1293

PR #1293, opened 2021-05-28
Correct logrotate error using reload over rotate

dalepgrant:
Relates to #1285 

Credit really to @robrecord for investigating the issue finding a solution [here](https://bugs.launchpad.net/nginx/+bug/1450770/comments/8).


swalkinshaw:
Thank you 🎉

PR #1294, opened 2021-06-01
Remove PHP version from messages

strarsis:
This PR removes the PHP versions from the messages.

Complements https://github.com/roots/trellis/pull/1284 (see https://github.com/roots/trellis/pull/1284#issuecomment-852086830).


swalkinshaw:
🎉  thank you

PR #1299, opened 2021-06-05
Add php-bcmath extension

joshuafredrickson:
This commit was missing from #1292. It adds the PHP bc_math extension. Thanks, @strarsis!


swalkinshaw:
Thanks!

PR #1309, opened 2021-08-12
Update Deploy Hook example's folder to public

jdebuchy:
Can we update folder name to `public` so it follows Sage 10 convention?


retlehs:
thank you!

PR #1313, opened 2021-10-17
Using variable for location instead of hard-coded

hemangajmera:
Currently `web` is hard-coded location path for nginx. This is true for bedrock. For sites which are not bedrock specific, it will be helpful to keep it configurable.
I know trellis only supports wordpress, but I am using this for even setting up laravel projects where the location path is `public` instead of `web`. I manually have to modify the nginx config, whenever I run ansible playbook for server.yml.


swalkinshaw:
I'm not against this in theory, but there are more references to `web` throughout Trellis. Ideally they'd all support this as well and I have a hunch in some cases it might not be so trivial since you want the variable to be site specific.

TangRufus:
I suggest we change the `deploy` tasks to use the `location_path` variable as well because it would help other developers to modify the `deploy` tasks for non-bedrock usages.

hemangajmera:
Have added another commit. Not sure if that is correct way.

PR #1320, opened 2021-11-11
Change MariaDB mirror source

swalkinshaw:
Digitalocean has proven to be an unreliable mirror so let's switch to another one (from the official MariaDB site).


jonlow:
Yep, this fixed a `apt cache update failed` error I was getting when provisioning a server today.

PR #1322, opened 2021-11-22
Update default PHP version to 8.0

swalkinshaw:


swalkinshaw:
PHP 8.1 is coming out in a few days so might as well wait for that

TangRufus (approved):

retlehs:
should 8.0 be the default though over 8.1 just because it's so new? we should definitely add 8.1 support at least tho, i think just by adding a dir to https://github.com/roots/trellis/tree/master/roles/php/vars?

swalkinshaw:
Yeah, actually there should probably be an between release with 8.0 at least.

PR #1324, opened 2021-11-27
Add GitHub SSH ed25519 key to known hosts

swalkinshaw:
Ref: https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/githubs-ssh-key-fingerprints

Got the full known host by running `ssh -T git@github.com` locally


cspicuzza:
This commit fixed an issue where I wasn't able to install composer packages while deploying to a
DO droplet

PR #1326, opened 2021-11-30
Remove ssl_dhparam and Diffie-Hellman group

swalkinshaw:
This was needed to prevent Logjam attacks but those only applied to DHE cyphers which haven't been supported in Trellis for 2 years.


swalkinshaw:
Confirmed we can remove these thanks to SSL Labs: 
image

Generating the DH params is one of the slowest parts of provisioning a server with Trellis (and even more annoying in dev when you want to test local SSL) so removing it should be noticeable.

PR #1332, opened 2021-12-10
Require Vagrant < 2.2.19

swalkinshaw:
2.2.19 has a bug affecting macOS Big Sur and should be avoided.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -16,7 +16,7 @@ ensure_plugins(vconfig.fetch('vagrant_plugins')) if vconfig.fetch('vagrant_insta
 
 trellis_config = Trellis::Config.new(root_path: ANSIBLE_PATH)
 
-Vagrant.require_version '>= 2.1.0'
+Vagrant.require_version '>= 2.1.0', '< 2.2.19'
 
 Vagrant.configure('2') do |config|
   config.vm.box = vconfig.fetch('vagrant_box')


TangRufus (approved):

tomektomczuk:
I'm on Vagrant 2.2.19 as this is necessary to run Trellis on MacOS Monterey without any problems.

swalkinshaw:
@tomektomczuk I've been following https://github.com/hashicorp/vagrant/issues/12557 but is there anything specific in 2.2.19 that's needed for Trellis?

I also noticed that VB 6.1.30 is out now which I assume fixes a few of these issues.

tomektomczuk:
Yes 6.1.30 and 2.2.19 fixed all the issues from that post. All you need is networks.conf file inside etc/vbox folder with this entry `* 0.0.0.0/0 ::/0`

swalkinshaw:
```plain
The IP address configured for the host-only network is not within the
allowed ranges. Please update the address used to be within the allowed
ranges and run the command again.

  Address: 192.168.50.5
  Ranges: 192.168.56.0/21

Valid ranges can be modified in the /etc/vbox/networks.conf file. For
more information including valid format see:

  https://www.virtualbox.org/manual/ch06.html#network_hostonly
exit status 1
```

Should Trellis just change the default IP to avoid this?

swalkinshaw:
2.2.19 still seems broken even on Monterey unfortunately 😞 

I get the same error message that others do on Big Sur:

> /opt/vagrant/embedded/gems/2.2.19/gems/vagrant-2.2.19/plugins/hosts/darwin/cap/path.rb:20:in `resolve_host_path': uninitialized constant VagrantPlugins::HostDarwin::Cap::Version (NameError)

tomektomczuk:
The fix is here
https://github.com/hashicorp/vagrant/issues/12583#issuecomment-964002289 -
it's one line of code to change

The reason you get that is because you haven't uninstalled completely your
previous version.

sob., 11 gru 2021 o 21:01 Scott Walkinshaw ***@***.***>
napisał(a):

swalkinshaw:
I understand there's a "fix", but unfortunately there's no version that's completely working right now for either Big Sur or Monterey.

I ran `brew reinstall vagrant` 🤷‍♂️

tomektomczuk:
I'm using Trellis on Monterey without any problems. My friend who also
upgraded from Big Sur to Monterey also has no problems.

I have downloaded Vagrant from Oracle and use this to Uninstall it. Them
after restart I used brew to install new version and everything works well
for me. The fix is already merged upstream
https://github.com/hashicorp/vagrant/pull/12581/commits/2bc1e2927bf34246e8f762d2c667b22b46d409df
.

sob., 11 gru 2021 o 21:14 Scott Walkinshaw ***@***.***>
napisał(a):

PR #1339, opened 2021-12-17
Stop old php-fpm services before starting new one

swalkinshaw:
The new version service can fail to start if there's already one running.


swalkinshaw:
Closing this since it was an issue only ever reported by one user. We can re-evaluate it if it ever comes up more but I'd be worried this might have other consequences.

PR #1343, opened 2021-12-25
Using variable for location instead of hard-coded

hemangajmera:
I closed https://github.com/roots/trellis/pull/1313 by mistake, so created another one.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -19,7 +19,13 @@ trellis_config = Trellis::Config.new(root_path: ANSIBLE_PATH)
 Vagrant.require_version '>= 2.1.0', '< 2.2.19'
 
 Vagrant.configure('2') do |config|
-  config.vm.box = vconfig.fetch('vagrant_box')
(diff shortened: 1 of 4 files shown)


swalkinshaw:
https://github.com/roots/trellis/blob/7a85b293df2e7aefaee73488be42e70a008b7ed3/roles/wordpress-install/tasks/directories.yml#L4 needs to be updated as well.

Also not sure about the generic `location_path` name; it's not very descriptive. What about `public_path`? And `public_uploads_path`

hemangajmera:
`public_path` and `public_upload_path` sounds better. Will create another request...

hemangajmera:
Also updated the code at https://github.com/roots/trellis/blob/7a85b293df2e7aefaee73488be42e70a008b7ed3/roles/wordpress-install/tasks/directories.yml#L4

hemangajmera:
Created clean pull request https://github.com/roots/trellis/pull/1378

hemangajmera:
Closing it.

PR #1344, opened 2021-12-29
Improve support for adding public keys for deploys

swalkinshaw:
Adding a deploy specific SSH public key to a Trellis server is a common task to enable CI/CD deploys (such as GitHub Actions).

This creates a standard folder (`public_keys`) for them. Any public SSH keys in that folder (ending in `.pub`) will be automatically added to the `web_user` as an authorized key.

### DIFF
--- a/lib/trellis/utils/output.py
+++ b/lib/trellis/utils/output.py
@@ -55,7 +55,7 @@ def replace_item_with_key(obj, result):
     )
 
     if should_replace:
-        if 'key' in result._result[item]:
+        if type(result._result[item]) is dict and 'key' in result._result[item]:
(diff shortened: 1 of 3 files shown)


lib/trellis/utils/output.py
```diff
@@ -55,7 +55,7 @@ def replace_item_with_key(obj, result):
     )
 
     if should_replace:
-        if 'key' in result._result[item]:
```
swalkinshaw, line 58: This PR exposed a bug in this code since `result._result[item]` was a string containing `"key"` (in `public_keys`) and Python's `in` operator is flexible for dict keys or substrings.

retlehs (approved):

PR #1352, opened 2022-01-09
Remove bin scripts

swalkinshaw:
These scripts were created as better interfaces to running the deploy and xdebug tunnel Ansible playbooks (since the Ansible CLI developer experience isn't the best when arguments are needed).

They've been superceded by the much better trellis-cli now. If you don't want to use trellis-cli, then the `ansible-playbook` commands that these scripts ran can always be used.

trellis-cli offers better validation, autocompletion, dependency management, and more.

## Deploy
Script:
```bash
./bin/deploy.sh production example.com
```

trellis-cli:
```bash
trellis production example.com
```

## xdebug-tunnel
Script:
```bash
./bin/xdebug-tunnel.sh open 12.34.56.78
```

trellis-cli:
```bash
trellis xdebug-tunnel open 12.34.56.78
```


TangRufus (approved):

swalkinshaw:
I need to remove some mounts from the Vagrantfile as well first, which is a nice side-effect of this change.

swalkinshaw:
Docs update: https://github.com/roots/docs/pull/342

PR #1359, opened 2022-01-28
Add integration test

swalkinshaw:
Uses trellis-cli to create a new Trellis project from the current SHA. The project is provisioned and deployed all using trellis-cli.

Finally the site is verified that WordPress was actually installed and working.

The steps are a little more complex than someone provisioning a normal remote server for a few reasons:

1. GitHub's Ubuntu image has a lot of software installed by default including mysql-server which needs to be purged completely.
2. there's a few workarounds to force Ansible to use a local connection since we aren't actually SSHing into a remote server.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -1,0 +1,51 @@
+name: Integration
+
+on:
+  push:
+    branches:
+      - master
+  workflow_dispatch:
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - run: mkdir $HOME/.ssh
+      - name: Remove and cleanup mysql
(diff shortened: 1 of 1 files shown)


roles/mariadb/tasks/main.yml
```diff
@@ -33,34 +33,32 @@
     when: mysql_binary_logging_disabled | bool
     notify: restart mysql server
 
+  - name: Copy .my.cnf file with root password credentials.
```
swalkinshaw, line 36: Going to separate this change into its own PR

swalkinshaw: https://github.com/roots/trellis/pull/1360

PR #1367, opened 2022-02-15
Tune default PHP opcache settings

swalkinshaw:
Based on https://tideways.com/profiler/blog/fine-tune-your-opcache-configuration-to-avoid-caching-suprises


swalkinshaw:
This needs benchmarks

PR #1370, opened 2022-03-02
Allow for more projects (similar) being provisioned in same server

davidmpaz:
Hi,

thanks for the great job with this repository. Could you consider to include this in the main stream please? My use case is the following:

When deploying several (similar) projects to same server, letsencrypt logic is overriden with latest project being provisioned in server, leading to issues when updating certificates.

This introduce a new variable derived (could be off course anything else) from wordpress_sites and rename the script generated based on this new "project_name" variable.

thanks in advance and cheers,
David


swalkinshaw:
> When deploying several (similar) projects to same server, letsencrypt logic is overriden with latest project being provisioned in server, leading to issues when updating certificates.

Not sure I understand the use case exactly yet. "deploying several (similar) projects" does this mean you have multiple Trellis projects _each_ with different `wordpress_sites` defined?

A bit more detail on your setup might help.

davidmpaz:
Hi,

thanks for answering and sorry for the late reply. 

> does this mean you have multiple Trellis projects each with different wordpress_sites defined?

Yes, it is the case.

Regards,
David

TangRufus:
Question: Any advantages to **not** putting all sites under the same Trellis project?

mikaelwedemeyer:
I don't think there are any advantages to separate Trellis projects for each site/server. We have approx 100 sites across around 20 servers in our single trellis project and it seems ideal.

charlestroluxe:
> I don't think there are any advantages to separate Trellis projects for each site/server. We have approx 100 sites across around 20 servers in our single trellis project and it seems ideal.

Out of interest, with this setup, how do you tell which site which server to deploy to? Or are you just scaling horizontally, with all 100 sites on all 20 servers?

davidmpaz:
> I don't think there are any advantages to separate Trellis projects for each site/server. We have approx 100 sites across around 20 servers in our single trellis project and it seems ideal.

I will take a look into this possibility, at the time, did seems like an overkill to use same trellis setup for just an extra static html website :-)

Thanks for answers, i will close this then.
Cheers,
David

mikaelwedemeyer:
> Out of interest, with this setup, how do you tell which site which server to deploy to? Or are you just scaling horizontally, with all 100 sites on all 20 servers?

Adding additional named hosts in group_vars, from scratch you would copy the production directory and rename it to the server name, eg `alpha`

This should work with current Trellis, I assumed it is in the docs somewhere, has been our process since 2015.

You then duplicate the host file for `production` in the `hosts` directory and then rename the filename to the host name and also change the group name `[production] `to `[alpha]`, for example, eg. Notew you still need it in the web group:

```
# Add each host to the [production] group and to a "type" group such as [web] or [db].
# List each machine only once per [group], even if it will host multiple sites.

[alpha]
000.000.000.00

[web]
000.000.000.00
```
These are the ansible commands rather than trellis-cli, haven't got around to that yet.

You can then reprovision/deploy to the host with the host name, eg:

`ansible-playbook server.yml -e env=alpha`

We deploy like this (could be outdated now, there is an equivalent in trellis-cli I believe):

`./bin/deploy.sh alpha sitename


Also apologies if this is an inappropriate place for this discussion, happy to move over to Discourse.

swalkinshaw:
> We deploy like this (could be outdated now, there is an equivalent in trellis-cli I believe):

Yeah, it's just `trellis deploy alpha sitename` 😄  (and `trellis provision alpha`)

charlestroluxe:
Thank you!

PR #1375, opened 2022-03-19
Add built-in fail2ban filters

swalkinshaw:
Trellis supported default fail2ban services previously but they were restricted to filters built into fail2ban itself (like `sshd`).

This adds filters defined by Trellis as well now by automatically creating the filter configuration files from templates.

Importantly, these filters will be _disabled_ by default. Any time a new filter is added, it will also be added to `fail2ban_services_custom` with enabled set to `false.`

This achieves a few goals:

1. makes it very easy to enable the built-in filters, which
2. brings more awareness to them by adding them to `group_vars/all/security.yml`
3. hopefully encourages more fail2ban filters to be created and used

There's two initial filters:
* wp-login
* xmlrpc

Which are both designed to prevent common DDoS attack vectors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### HEAD
+* Add built-in fail2ban filters ([#1375](https://github.com/roots/trellis/pull/1375))
 * Support Ansible >= 2.10 (tested up to 5.4.0) ([#1373](https://github.com/roots/trellis/pull/1373))
 * Remove Python 2 support ([#1361](https://github.com/roots/trellis/pull/1361))
 
--- a/group_vars/all/security.yml
+++ b/group_vars/all/security.yml
@@ -1,3 +1,5 @@
+# Documentation: https://roots.io/trellis/docs/security/
+
 ferm_input_list:
   - type: dport_accept
     dport: [http, https]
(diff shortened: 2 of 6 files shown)


swalkinshaw:
@TangRufus I think we "recommend" (or at least link to) https://github.com/ItinerisLtd/trellis-disable-xml-rpc but the fail2ban based solution should be better since it will prevent it at the iptables level so requests don't even reach Nginx.

Any thoughts?

TangRufus (approved):

TangRufus (approved):

TangRufus:
> the fail2ban based solution should be better since it will prevent it at the iptables level so requests don't even reach Nginx.

Agree.

> I think we "recommend" (or at least link to) [ItinerisLtd/trellis-disable-xml-rpc](https://github.com/ItinerisLtd/trellis-disable-xml-rpc)

Agree. They block at different levels. It should be okay to have both enabled at the same time.

---

We should update https://docs.roots.io, especially:

1. how to determine whether an IP is being banned
1. how to use `fail2ban_ignoreip`
1. how to unban IPs
1. warn about legitimate use cases of xmlrpc (e.g: [jetpack](https://jetpack.com/support/getting-started-with-jetpack/troubleshooting-tips/))
1. warn about _false_ IPs, Nginx might logged proxy IPs / VPN IPs

---

Question: When an IP is banned because of one of the fail2ban services (e.g: `wordpress-xmlrpc`), does fail2ban ban all its access (to different URLs) via http and https?

If yes, we should warn the users.

TangRufus (approved):
Looks good to me

swalkinshaw:
> Question: When an IP is banned because of one of the fail2ban services (e.g: wordpress-xmlrpc), does fail2ban ban all its access (to different URLs) via http and https?

I believe so since the ban is done at the iptables level which doesn't know about URLs. So the initial _detection_ is URL based, but the ban isn't.

PR #1382, opened 2022-05-14
Do not disallow up to date Vagrant vers. on Linux

nextgenthemes:
When this version restriction was introduced, I found out from the changelog or an issue that this was specifically because of a mac OS issue.

I have successfully ran later Vagrant versions that ship with Open SUSE Tumbleweed (Not sure), Ubuntu 21.10 and Ubuntu 22.04. AFAIK there is no reason to shut Linux users out of up-to-date Vagrant versions.


swalkinshaw:
👍  yeah this is probably an improvement. The current version constraint is more macOS specific

PR #1383, opened 2022-05-22
fix: set mjs as application/javascript in h5bp

kellymears:
Problem: nginx serves `mjs` as `application/octet-stream` and results in errors in the browser:

```
Failed to load module script: Expected a JavaScript module script but the server responded with a MIME type of "application/octet-stream".
```

This change sets `mjs` as `application/javascript` so that these files are served in accordance with the HTML specification.

### DIFF
--- a/roles/nginx/templates/h5bp/mime.types
+++ b/roles/nginx/templates/h5bp/mime.types
@@ -14,7 +14,7 @@ types {
 
     # Normalize to standard type.
     # https://tools.ietf.org/html/rfc4329#section-7.2
-    application/javascript                js;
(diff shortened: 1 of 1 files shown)


swalkinshaw (approved):
Oh nice one 🎉

TangRufus (approved):
Looks good to me

Related: https://github.com/h5bp/server-configs-nginx/pull/187/files#diff-132a8679cdda32834a76d9ec05ed2824193d3b95af21c5031a53972e68cd949aR17

retlehs:
@kellymears no, probably fixed by #1384

Issue #1385, opened 2022-05-25
Bug: Autodeploy failures with WordPress 6.0


SteelWagstaff:
### Terms

- [X] I have read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)
- [X] This request is not a duplicate of an existing issue
- [X] I have read the [docs](https://roots.io/docs/) and followed them (if applicable)
- [X] I have seached the [Roots Discourse](https://discourse.roots.io/) for answers and followed them (if applicable)
- [X] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/) community

#### What's wrong?

Our Trellis-based autodeploy pipeline is failing to create a WordPress multisite installation with the latest WordPress 6.0 release. In particular, this command is failing: https://github.com/roots/trellis/blob/f1f8ff1c7d10dba602916aff4c04d0132c524da3/roles/deploy/hooks/finalize-before.yml#L7-L8

We think it's because the WP CLI command does like WPMU_PLUGIN_DIR being defined as `null`: https://github.com/roots/trellis/blob/f1f8ff1c7d10dba602916aff4c04d0132c524da3/roles/deploy/files/tmp_multisite_constants.php#L5

We don't know what might be different about WordPress 6.0 that might be producing this problem/issue.

#### What have you tried?
We can complete the deploy by removing the `--require` parameter in https://github.com/roots/trellis/blob/f1f8ff1c7d10dba602916aff4c04d0132c524da3/roles/deploy/hooks/finalize-before.yml#L7-L8

We can also complete the deploy by commenting out `define('WPMU_PLUGIN_DIR', null);` here https://github.com/roots/trellis/blob/f1f8ff1c7d10dba602916aff4c04d0132c524da3/roles/deploy/files/tmp_multisite_constants.php#L5

However, neither of these seem like desirable solutions, considering the rationale provided by the author of this PR: https://github.com/roots/trellis/pull/801. Not sure if others are experiencing similar problems or have more insight into how to resolve them.

#### What insights have you gained?
This WordPress multisite stuff is complicated :joy: 

#### Possible solutions
See the `What have you tried?` section

#### Temporary workarounds
See the `What have you tried?` section

### Steps To Reproduce

1. Attempt to deploy WordPress multisite using the latest Trellis repo and roots/bedrock
1. Observe the output of the 'TASK [deploy: WordPress Installed?]` step
[...]

PR #1386, opened 2022-05-29
Remove `WPMU_PLUGIN_DIR` constant to `null` in `tmp_multisite_constants.php`

strarsis:
(Variant A) This PR removes the line that sets the `WPMU_PLUGIN_DIR` constant to `null` in `tmp_multisite_constants.php` so the Ansible check that uses `wp core is-installed [...]` passes (failing since the WordPress `6.0` release).

Before merging, it may be good to think about possible issues that may arise with not setting the `WPMU_PLUGIN_DIR` constant to null in that (smoke) test during deployment. Can this result in passing deploys for multisite setups that would have issues, hence should not have passed this check?
Previous discussions (https://github.com/roots/trellis/issues/1385) indicate that this shouldn't be the case and that setting that constant to `null` isn't really required anymore with recent WordPress versions.

Fixes issue: https://github.com/roots/trellis/issues/1385

### DIFF
--- a/roles/deploy/files/tmp_multisite_constants.php
+++ b/roles/deploy/files/tmp_multisite_constants.php
@@ -2,6 +2,5 @@
 error_reporting(E_ALL & ~E_NOTICE);
 define('MULTISITE', false);
 define('SUBDOMAIN_INSTALL', false);
-define('WPMU_PLUGIN_DIR', null);
 define('WP_PLUGIN_DIR', null);
 define('WP_USE_THEMES', false);


swalkinshaw:
If someone else could help test with a multisite setup that would be helpful. I'm on vacation but I'm happy to get this merged; I just can't really help test.

I'll give this ~24 hours anyway since non-multisite setups are way more common so it's better to fix those at least.

strarsis:
@swalkinshaw: An alternative PR is also in preparation that will let the Ansible check just ignore that particular PHP warning.
This would be an alternative if setting that `WPMU_PLUGIN_DIR` constant to `null` somehow proves to be important for multisite.
And the check could be split into two different checks, one for non-multisite and one specifically for multisites.

strarsis:
@swalkinshaw: Alternative PR that makes the `WordPress installed?` check ignore this particular PHP warning: https://github.com/roots/trellis/pull/1387

strarsis:
Closed as Variant C PR (https://github.com/roots/trellis/pull/1388) was merged.

PR #1391, opened 2022-06-11
Exempt from `is-installed` check the DB error dump PHP warning for not yet set up multisite sites

strarsis:
This PR changes the [recently added `stderr` exemption](https://github.com/roots/trellis/pull/1388) of a specific PHP warning from the WP CLI `wp core is-installed` command for multisite sites to the DB error dump that occurs with not yet set up multisite sites.

This eliminates the need for letting the WP CLI require a temporary PHP file with constants that disable the multisite feature, which itself caused fatal PHP errors in recent WordPress/Bedrock/Trellis setups where those multisite constants where redefined in the Bedrock application config.

Initial deployment of multisite sites will successfully complete and allow finalization of the site setup (by using the WP CLI, web GUI setup process, importing a database dump/transfer uploads/data files, etc).

- As an improvement the [Trellis Multisite documentation](https://docs.roots.io/trellis/master/multisite/#:~:text=After%20provisioning%20your%20remote%20server%20and%20deploying%20your%20sites) could be updated so that it includes the web GUI URL for the site setup (if someone prefers or needs that over the WP CLI approach): `http://<a canonical host or domain_current_site>/wp/wp-admin/install.php` as frontend request redirection to the setup web GUI page may not work until the site setup is completed.
- ~~Should Ansible clean up an existing `tmp_multisite_constants.php` file? It would be useless anyway and its purpose become forgotten on existing Trellis systems.~~ (added: https://github.com/roots/trellis/pull/1391/files#diff-da02775cb0a887a7b8c5f5183fffe57a8840d53f47065f919dff77ad2accd4a0R2-R6)

### DIFF
--- a/roles/deploy/hooks/finalize-before.yml
+++ b/roles/deploy/hooks/finalize-before.yml
@@ -1,4 +1,9 @@
 ---
+- name: Clean up unused, temporary PHP file with multisite constants that had been used for WordPress Installed check.
+  file:
+    state: absent
+    path: "{{ deploy_helper.shared_path }}/tmp_multisite_constants.php"
+
 - name: WordPress Installed (non-multisite)?
   block:
     - name: "Invoke 'wp core is-installed' command"
--- a/roles/deploy/hooks/finalize-before.yml
+++ b/roles/deploy/hooks/finalize-before.yml
@@ -1,5 +1,5 @@
 ---
-- name: Clean up unused, temporary PHP file with multisite constants that had been used for WordPress Installed check.
+- name: Clean up unused, temporary PHP file with multisite constants that had been used for WordPress Installed checks.
   file:
     state: absent
     path: "{{ deploy_helper.shared_path }}/tmp_multisite_constants.php"
--- a/roles/deploy/hooks/finalize-before.yml
+++ b/roles/deploy/hooks/finalize-before.yml
@@ -17,22 +17,17 @@
 
 - name: WordPress Installed (multisite)?
   block:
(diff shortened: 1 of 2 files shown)


roles/deploy/hooks/finalize-before.yml
```diff
@@ -17,22 +17,17 @@
 
 - name: WordPress Installed (multisite)?
   block:
-    - name: Create file with multisite constants defined as false
-      copy:
-        src: "tmp_multisite_constants.php"
```
swalkinshaw, line 22: This file can be deleted now too

strarsis: https://github.com/roots/trellis/pull/1391/commits/b58ca90e5d99617c713462952cc8c02904178557
This way to clean up the file by Ansible looks save enough I think (`file state: absent`).

swalkinshaw: Sorry I meant https://github.com/roots/trellis/blob/17430191bb7211545eb63ba3ba989ee95c262c5f/roles/deploy/files/tmp_multisite_constants.php can be deleted 😅 

I'd be okay without the clean up task 🤷‍♂️ but it's fine either way

strarsis: @swalkinshaw: https://github.com/roots/trellis/pull/1391/files#diff-051d620484530c4b6b44daf807be7915a2c6b14046c3d0fd24f325b51ea353ad

swalkinshaw (approved):
🚀

Issue #1393, opened 2022-06-22
Bug: Incompatibility with Ansible 6


codepuncher:
### Terms

- [X] I have read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)
- [X] This request is not a duplicate of an existing issue
- [X] I have read the [docs](https://roots.io/docs/) and followed them (if applicable)
- [X] I have seached the [Roots Discourse](https://discourse.roots.io/) for answers and followed them (if applicable)
- [X] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/) community

#### What's wrong?
Ansible 6 was released yesterday alongside ansible-core 2.13.1
This causes deployments to fail.

#### What have you tried?
Adding a version constraint to `requirements.txt`
```diff
- ansible>=2.10.0
+ ansible>=2.10.0,<6
passlib
```
This works

#### What insights have you gained?
I needed a quick way to apply the above patch to all projects, so a helpful bash script was made as a result of this.

#### Possible solutions
Either use the above version constraint or update Trellis to support Ansible 6

#### Temporary workarounds
As mentioned above.

### Steps To Reproduce

1. `rm -rf trellis/.trellis/virtualenv`
2. `trellis init`
3. `trellis exec ansible --version` # Verify ansible-core 2.13.1 is in use
4. `trellis deploy staging`

### Expected Behavior

To deploy without fail

### Actual Behavior

The deployment fails

### Relevant Log Output

```shell
TASK [connection : Check whether Ansible can connect as chiuni] ****************
System info:
  Ansible 2.13.1; Darwin
  Trellis version (per changelog): "Add built-in fail2ban filters"
---------------------------------------------------
The field 'environment' has an invalid value, which includes an undefined
variable. The error was: 'ssh_args_default' is undefined

The error appears to be in '/Users/lee-
work/Code/wordpress/chi/trellis/roles/connection/tasks/main.yml': line 18,
column 3, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:


- name: Check whether Ansible can connect as {{ dynamic_user | default(true)
| ternary('root', web_user) }}
  ^ here
We could be wrong, but this one looks like it might be an issue with
missing quotes. Always quote template expression brackets when they
start a value. For instance:

    with_items:
      - {{ foo }}

Should be written as:

    with_items:
      - "{{ foo }}"
[...]

PR #1396, opened 2022-06-24
Limit ansible version range in `>=2.10.0,<6`

TangRufus:
See: https://github.com/roots/trellis/issues/1393


swalkinshaw:
I'm good with this for now, thanks @TangRufus

PR #1397, opened 2022-06-24
Fix ansible incompatibility with `Failed connection to remote repo`

TangRufus:
Newer versions of `ansible.builtin.git` puts error message in `msg` instead of `stderr`.


swalkinshaw:
Should be backwards compatible too 👍

PR #1403, opened 2022-07-10
Fix shared folder permissions

### LINKED ISSUE #1400: Bug: Wrong 0777 permission on web/app folder on every deploy
### Terms

#### What's wrong?

Every time I deploy, either to Kinsta, custom server or Trellis managed server, the web/app folder from bedrock gets permissions 0777.

If I override the **project_shared_children** var in my trellis/group_vars/production/main.yml file to share some files and folders across deploys, then I get the web/app folder chmod 0777  and the web folder also with 0777. 

Maybe I'm wrong, but I think applying 0777 to a folder is too permisive and this should be changed to 0755

This is the disourse topic related to this issue I've opened almost a year ago. I don't know how no one have seen this before. Maybe I'm the only one that has the problem.

https://discourse.roots.io/t/deploy-places-web-and-app-folder-777/21087

#### What have you tried?

I think this line is the culprit, as if I change it in my trellis install then the web and web/app folders get 0755 as I think they should...

https://github.com/roots/trellis/blob/e497cfe002fa00fc33447131c0eb2570634453af/roles/deploy/tasks/share.yml#L43

If I don't override the **project_shared_children** var then only the **web/app** folder has 0777.
If I override the **project_shared_children** var then **web** and **web/app** folder both two have 0777.

Using the mode option in project_shared_children doesn't work as this is related to the parent folder:
https://github.com/roots/trellis/blob/e497cfe002fa00fc33447131c0eb2570634453af/roles/deploy/defaults/main.yml#L32

#### What insights have you gained?

Maybe all our websites are a bit exposed to attackers.

#### Possible solutions

Change this line

https://github.com/roots/trellis/blob/e497cfe002fa00fc33447131c0eb2570634453af/roles/deploy/tasks/share.yml#L43

to 0755

#### Temporary workarounds

Change this line

https://github.com/roots/trellis/blob/e497cfe002fa00fc33447131c0eb2570634453af/roles/deploy/tasks/share.yml#L43

to 0755

### Steps To Reproduce

Just deploy using trellis and see the permissions of web/app on bedrock.

Captura de pantalla 2022-07-01 a las 14 44 15

I you override the **project_shared_children** var in your trellis/group_vars/production/main.yml file with something like this:

```
project_shared_children:
  - path: web/app/uploads
    src: uploads
  - path: web/robots.txt
    src: robots.txt
```
...then you'll get web and web/app folder with 0777.

### Expected Behavior

web/app folder to be 0755

### Actual Behavior

web/app folder is 0777

### Relevant Log Output

_No response_

### Versions

Trellis v1.15.0 - OSX 10.15.7

swalkinshaw:
Fixes #1400

This should match a related task for shared paths in https://github.com/roots/trellis/blob/5411c18bc8963710f4f5423ac5819cfb1f0dbbea/roles/deploy/tasks/share.yml#L28

### DIFF
--- a/roles/deploy/tasks/share.yml
+++ b/roles/deploy/tasks/share.yml
@@ -40,7 +40,7 @@
 - name: Ensure parent directories for shared paths are present
   file:
     path: "{{ deploy_helper.new_release_path }}/{{ item.path | dirname }}"
-    mode: '0777'
+    mode: '0755'
     state: directory
   with_items: "{{ project.project_shared_children | default(project_shared_children) }}"
 


TangRufus:
Looks good to me

PR #1411, opened 2022-07-21
Run composer check-platform-reqs during deploy

### LINKED ISSUE #1244: Feature request: Perform composer platform check during deploy
> To avoid an unexpected white page of death with some obscure PHP extension warning in production, you can run composer check-platform-reqs --no-dev as part of your deployment/build and if that returns a non-0 code you should abort.
>
> -- https://getcomposer.org/doc/07-runtime.md#platform-check

Requires composer v2.

swalkinshaw:
Fixes #1244

Runs `composer check-platform-reqs` before installing composer dependencies to verify that platform requirements (i.e. php and php extensions) are fulfilled by the PHP process currently running. This could potentially prevent some "white page of death" errors before a deploy.

Ref https://getcomposer.org/doc/07-runtime.md#platform-check

### DIFF
--- a/roles/deploy/defaults/main.yml
+++ b/roles/deploy/defaults/main.yml
@@ -51,6 +51,8 @@ update_db_on_deploy: true
 
 # Most scripts are used in development instead of remote servers. Use with caution.
 composer_no_scripts: true
+# Whether to run `composer check-platform-reqs`.
+composer_platform_requirements_check: true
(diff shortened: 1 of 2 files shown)


TangRufus:
Can we add a `when` conditional to allow skipping?

swalkinshaw:
Good idea 👍  added

TangRufus (approved):

PR #1412, opened 2022-07-22
Improve composer authentications password default

swalkinshaw:
Removes the validation to enforce `password` and sets the default to `""` which is quoted to ensure `composer` is properly passed `""` as the last argument value.

### DIFF
--- a/roles/deploy/hooks/build-after.yml
+++ b/roles/deploy/hooks/build-after.yml
@@ -12,14 +12,13 @@
 - name: Setup composer authentications
   composer:
     command: config
-    arguments: --auth http-basic.{{ composer_authentication.hostname }} {{ composer_authentication.username }} {{ composer_authentication.password }}
+    arguments: --auth http-basic.{{ composer_authentication.hostname }} {{ composer_authentication.username }} {{ composer_authentication.password | default("") | quote }}
     working_dir: "{{ deploy_helper.new_release_path }}"
   no_log: true
   changed_when: false
   when:
     - composer_authentication.hostname is defined and composer_authentication.hostname != ""
     - composer_authentication.username is defined and composer_authentication.username != ""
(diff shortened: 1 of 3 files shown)


roles/deploy/hooks/build-after.yml
```diff
@@ -12,14 +12,13 @@
 - name: Setup composer authentications
   composer:
     command: config
-    arguments: --auth http-basic.{{ composer_authentication.hostname }} {{ composer_authentication.username }} {{ composer_authentication.password }}
+    arguments: --auth http-basic.{{ composer_authentication.hostname }} {{ composer_authentication.username }} {{ composer_authentication.password | default("") | quote }}
```
TangRufus, line 15:
```suggestion
    arguments: --auth http-basic.{{ composer_authentication.hostname | quote }} {{ composer_authentication.username | quote }} {{ composer_authentication.password | default("") | quote }}
```

Although `hostname` should never contains whitespaces, quoting it anyway for security.

roles/wordpress-install/tasks/composer-authentications.yml
```diff
@@ -2,14 +2,13 @@
 - name: "Setup composer authentications - {{ site }}"
   composer:
     command: config
-    arguments: --auth http-basic.{{ composer_authentication.hostname }} {{ composer_authentication.username }} {{ composer_authentication.password }}
+    arguments: --auth http-basic.{{ composer_authentication.hostname }} {{ composer_authentication.username }} {{ composer_authentication.password | default("") | quote }}
```
TangRufus, line 5:
```suggestion
    arguments: --auth http-basic.{{ composer_authentication.hostname | quote }} {{ composer_authentication.username | quote }} {{ composer_authentication.password | default("") | quote }}
```

PR #1414, opened 2022-07-27
Simplify Nginx no-default sites for HTTPS

swalkinshaw:
Nginx has a newer `ssl_reject_handshake` directive that lets us simplify the "no default" site conf. Before we had to generate a self-signed certificate just to satisfy Nginx and create a server that listened on port 443.

Now with `ssl_reject_handshake`, SSL handshakes with server names that are *not* part of the real WordPress server hosts will be rejected up front.

This means we can skip self-signed certificate generation in default cases. The two separate server configs (HTTP and HTTPS) have now been merged into one as well since they are simpler overall.

This was extracted from https://github.com/roots/trellis/pull/1310/ since it should be merged on its own.


dalepgrant:
[Related Discourse thread](https://discourse.roots.io/t/trellis-duplicate-listen-options-for-443-in-etc-nginx-sites-enabled-ssl-no-default-conf-11/23660) for anyone ending up here looking for 
```
non-zero return code
nginx: [emerg] duplicate listen options for [::]:443 in /etc/nginx/sites-enabled/ssl.no-default.conf:11
nginx: configuration file /etc/nginx/nginx.conf test failed
```

swalkinshaw:
Oh because the old conf/site was left behind and active 😓 Thanks for posting @dalepgrant

swalkinshaw:
https://github.com/roots/trellis/pull/1415 should prevent the backwards compat issue once merged. Turned out to be simple and better ✨

PR #1424, opened 2022-09-13
Nginx gzip javascript files

pacotole:
Add "text/javascript" to gzip_types to enable js files compression.


swalkinshaw (approved):
Thank you! Looks like we're missing a few others, I'll add them separately.

PR #1431, opened 2022-10-12
Automate arm64 Vagrant box support for Apple Silicon

swalkinshaw:
Adds built-in support to automate host architecture detection and setting of arm64 specific boxes for Apple Silicon.

Previously the `vagrant_box` config setting would have to be manually overwritten. Now Vagrant will detect the OS architecture of the host and _append_ `-arm64` (by convention) when three conditions apply:

1. `vagrant_box_auto_arch` is enabled (default: true)
2. `vagrant_box` does not already end up with `-arm64`
3. the OS is running on Apple Silicon

### DIFF
--- a/vagrant.default.yml
+++ b/vagrant.default.yml
@@ -3,7 +3,8 @@ vagrant_ip: '192.168.56.5'
 vagrant_cpus: 1
 vagrant_memory: 1024 # in MB
 vagrant_box: 'bento/ubuntu-20.04'
-vagrant_box_version: '>= 202012.23.0'
+vagrant_box_version: '>= 0'
+vagrant_box_auto_arch: true
 vagrant_ansible_version: '2.10.7'
 vagrant_skip_galaxy: false
 vagrant_mount_type: 'nfs'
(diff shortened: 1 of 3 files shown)


vagrant.default.yml
```diff
@@ -3,7 +3,8 @@ vagrant_ip: '192.168.56.5'
 vagrant_cpus: 1
 vagrant_memory: 1024 # in MB
 vagrant_box: 'bento/ubuntu-20.04'
-vagrant_box_version: '>= 202012.23.0'
+vagrant_box_version: '>= 0'
```
TangRufus, line 6: Why `>= 0`?

swalkinshaw: I guess it doesn't matter as much now that we're using the main `bento` boxes which all have consistent date based version numbers. But before if you override `vagrant_box` to an arm64 one from a different provider, this version constraint would prevent you from using it so you'd have to overwrite it as well.

`'>= 0` just means "give me the latest box" which honestly isn't too different than what we had since we never actually bump this.

TangRufus (approved):

retlehs (approved):

PR #1432, opened 2022-10-13
Support ARM64 arch in Mailhog role

swalkinshaw:
`geerlingguy.mailhog` doesn't support arm64 (Apple Silicon) because the underlying binaries (`mailhog` and `mhsendmail`) don't have official `arm64` builds.

To support this by default, we need to override the url variables to point to forks that have these builds.

This adds a new "shim loader" `mailhog` role which simply loads the proper vars and then passes them to the included `geerlingguy.mailhog` role.

### DIFF
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -16,6 +16,5 @@ roles:
     src: oefenweb.swapfile
     version: v2.0.32
 
-  - name: mailhog
-    src: geerlingguy.mailhog
+  - src: geerlingguy.mailhog
     version: 2.3.0
--- a/roles/mailhog/defaults/main.yml
+++ b/roles/mailhog/defaults/main.yml
@@ -1,0 +1,1 @@
+mailhog_install_dir: /opt/mailhog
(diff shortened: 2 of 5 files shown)


TangRufus:
Does https://github.com/geerlingguy/ansible-role-mailhog/pull/29/ solves the issue?

swalkinshaw:
Unfortunately not because that still points to the official repos which don't contain builds for arm64; only 32bit ARM.

retlehs (changes requested):
Currently breaks provisioning at this step:

```
TASK [php : Copy PHP-FPM configuration file] ***********************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ansible.errors.AnsibleUndefinedVariable: {{ mailhog_install_dir }}/mhsendmail: 'mailhog_install_dir' is undefined
fatal: [default]: FAILED! => {"changed": false, "msg": "AnsibleUndefinedVariable: {{ mailhog_install_dir }}/mhsendmail: 'mailhog_install_dir' is undefined"}
```

retlehs (approved):

PR #1433, opened 2022-10-13
Refactor Vagrant plugins

swalkinshaw:
Removes the custom method of installing plugins and uses Vagrant's built-in `config.vagrant.plugins` setting.


swalkinshaw:
This still needs some work.

PR #1435, opened 2022-10-14
Only allow caching for GET and HEAD requests

swalkinshaw:
ref https://github.com/roots/trellis/pull/1434#issuecomment-1279556417

Thanks for the reminder @PDowney 

cc @craigpearson

### DIFF
--- a/roles/wordpress-setup/templates/wordpress-site.conf.j2
+++ b/roles/wordpress-setup/templates/wordpress-site.conf.j2
@@ -37,13 +37,8 @@ server {
   # Fastcgi cache conditions
   set $skip_cache 0;
 
-  # Prevent POST data being held in cache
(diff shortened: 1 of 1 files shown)


craigpearson:
Thanks for the input and attention all. Based on the original suggestion [here](https://github.com/roots/trellis/pull/1434#issuecomment-1278981408) and comments [here](https://github.com/roots/trellis/pull/1434#issuecomment-1279556417), would you see any issue tackling this with:

```
if ($request_method !~ ^(GET|HEAD)$) {
    set $skip_cache 1;
}
```

I remember reading something many moons ago that this type of if statement is "evil" but based on the list of HTTP methods there isn't much room for any unknowns?

swalkinshaw:
@craigpearson yeah works for me 👍

PR #1430, opened 2022-10-07
Adding server tuning adjustments for MariaDB and PHP FPM

paulbrzeski:
Sharing a couple of changes we've made to our Trellis so we can fine tune MariaDB and PHP settings for high traffic websites.

Changes include,
- MariaDB/innodb optimisation
- PHP-FPM crash prevention and optimisation
- Small improvements to make things more semantic
- Various configs to support above

This PR only provides the defaults, you can override these settings like any other using your group_vars configs.

Example settings from our production servers below,

MariaDB for a server with 4GB of RAM and sharing environment with PHP and Nginx
New variables to add to ```group_vars/%ENV%/mariadb.yml```,
```
mariadb_set_innodb_buffer_pool_size: true
mariadb_innodb_buffer_pool_size: 512M
mariadb_set_innodb_log_file_size: true
mariadb_innodb_log_file_size: 128M
```
More info - https://mariadb.com/kb/en/innodb-system-variables/#innodb_buffer_pool_size

PHP FPM for a server sharing environment with MariaDB.
New variables to add to ```group_vars/%ENV%/php.yml```,
```
# Allow a max of 100 children, based on 16GB of ram and 164mb per child.
php_fpm_pm_max_children: 100
# Start and go up in small incremenets as we have the CPU to spare.
php_fpm_pm_start_servers: 5
php_fpm_pm_min_spare_servers: 5
php_fpm_pm_max_spare_servers: 35
# Allow lots of requests as we have the resources to spare.
php_fpm_pm_max_requests: 1000

# Emergency restart settings.
# Restart after 60 seconds if 10 PHP instances have not responded for 5 min

# Number of child processes not responding
php_fpm_set_emergency_restart_threshold: true
php_fpm_emergency_restart_threshold: 10

# Length of time children not responding.
php_fpm_set_emergency_restart_interval: true
php_fpm_emergency_restart_interval: 5m

# Time to wait before restarting after interval passes.
php_fpm_set_process_control_timeout: true
php_fpm_process_control_timeout: 60s
```
More info - https://geekflare.com/php-fpm-optimization/

PR #1444, opened 2022-10-25
Adding server tuning adjustments for PHP FPM

paulbrzeski:
Sharing a couple of changes we've made to our Trellis so we can fine tune PHP settings for high traffic websites.

Changes include,
- PHP-FPM crash prevention and optimisation
- Small improvements to make things more semantic
- Various configs to support above

This PR only provides the defaults, you can override these settings like any other using your group_vars configs.

PHP FPM for a server sharing environment with MariaDB.
New variables to add to ```group_vars/%ENV%/php.yml```,
```
# Allow a max of 100 children, based on 16GB of ram and 164mb per child.
php_fpm_pm_max_children: 100
# Start and go up in small incremenets as we have the CPU to spare.
php_fpm_pm_start_servers: 5
php_fpm_pm_min_spare_servers: 5
php_fpm_pm_max_spare_servers: 35
# Allow lots of requests as we have the resources to spare.
php_fpm_pm_max_requests: 1000

# Emergency restart settings.
# Restart after 60 seconds if 10 PHP instances have not responded for 5 min

# Number of child processes not responding
php_fpm_set_emergency_restart_threshold: true
php_fpm_emergency_restart_threshold: 10

# Length of time children not responding.
php_fpm_set_emergency_restart_interval: true
php_fpm_emergency_restart_interval: 5m

# Time to wait before restarting after interval passes.
php_fpm_set_process_control_timeout: true
php_fpm_process_control_timeout: 60s
```
More info - https://geekflare.com/php-fpm-optimization/

(previously roots/trellis#1430)


roles/php/templates/php-fpm.conf.j2
```diff
@@ -0,0 +1,151 @@
+;;;;;;;;;;;;;;;;;;;;;
```
swalkinshaw, line 1: Can you add `; {{ ansible_managed }}` here?

swalkinshaw (commented):
Looks good. Would you be able to rebase this to fix the conflict?

paulbrzeski:
This should be good to go now @swalkinshaw

swalkinshaw:
See https://github.com/roots/trellis/pull/1496

Issue #1392, opened 2022-06-18
Feature Request: Add support for ubuntu 22.04 LTS


hemangajmera:
### Terms

- [X] I have read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)
- [X] This request is not a duplicate of an existing issue
- [X] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/) community

### Summary

Upgrading to ubuntu 22.04.

#### Why are we doing this?
Moving to higher version

#### What use cases does it support?
Better long term support

#### What is the expected outcome?
default version can be ubuntu 22.04

#### Potential conflicts / foreseeable issues
Not sure

swalkinshaw:
Do you want to try it out as a start and see if anything doesn't work? That's always the first step in this process.

It might be as simple as updating the vagrant box: https://github.com/roots/trellis/blob/6338bd21425906a62f90480b1b0bd8b80655bc56/vagrant.default.yml#L5

sangemaru:
@swalkinshaw most likely, the Python 3.10 that comes with Ubuntu 22.04 will break anything ansible-related.

swalkinshaw:
I attempted to try this out on Apple Silicon with Parallels using a few different boxes: https://app.vagrantup.com/boxes/search?utf8=%E2%9C%93&sort=downloads&provider=parallels&q=ubuntu+22+arm

I ran into the SSH issues described starting here: https://github.com/chef/bento/issues/1405#issuecomment-1148977615

Even with the latest Vagrant 2.3.0 I couldn't get passed it so this remains the main hurdle to getting 22.04 support.

lukasz-gorski:
Any updates on it? Have pushing from DevOps to jump into 22.04. I know there are work arounds, but I would prefer native support.

swalkinshaw:
No, no updates. As I said earlier, it would be helpful for people to _try_ upgrading their vagrant box version and see what works and what doesn't.

clintonb:
I am trying this now with a fresh DigitalOcean droplet. Note that I am using a remote DB, and not installing MariaDB on the machine.

https://github.com/roots/trellis/pull/1453 has been merged, and fixes an issue. I also had to change to upgrade to MariaDB 10.6, which was done by simply changing the version number in the URL.

swalkinshaw:
MariaDB version is bumped to 10.6 as of https://github.com/roots/trellis/pull/1460

[...]

PR #1453, opened 2022-12-11
Fix sshd config - set AcceptEnv conditionally

swalkinshaw:
GH workflows switching to Ubuntu 22.04 exposed this bug. https://github.com/roots/trellis/pull/1454 is the proper fix for now.

This might be needed for https://github.com/roots/trellis/issues/1392


clintonb (commented):
This is definitely needed for Ubuntu 22.04.

swalkinshaw:
Merging now since it's backward compatible anyway 👍

PR #1458, opened 2022-12-15
Tweak multisite cron job to fix standard output

E-VANCE:
With our multisite-setup I have been seeing hourly attempts within our Mailgun-logs to send out an email notification regarding the cron job that is set up there:

`cd {{ www_root }}/{{ item.key }}/{{ item.value.current_path | default('current') }} && wp site list --field=url | xargs -n1 -I \\% wp --url=\\% cron event run --due-now > /dev/null 2>&1`

Seeing that `> /dev/null 2>&1` is appended, this shouldn't be happening...

Wrapping the two joint commands in brackets (`(wp site list --field=url | xargs -n1 -I \\% wp --url=\\% cron event run --due-now) > /dev/null 2>&1`) fixes the standard output and I am not seeing any attempts to send out email notifications in our Mailgun logs any more.

Have successfully tested running the new command in full on our server and it performs as expected.

### DIFF
--- a/roles/wordpress-setup/tasks/main.yml
+++ b/roles/wordpress-setup/tasks/main.yml
@@ -62,7 +62,7 @@
     name: "{{ item.key }} WordPress network cron"
     minute: "{{ item.value.cron_interval_multisite | default('*/30') }}"
     user: "{{ web_user }}"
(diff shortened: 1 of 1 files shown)


swalkinshaw:
I couldn't really replicate this locally with different commands, but it makes sense in theory and there shouldn't be any downside regardless 👍

swalkinshaw:
@E-VANCE could you rebase to get rid of the other two commits please?

E-VANCE:
@swalkinshaw Sorry for the overhead / mess... Should be clean now.

PR #1463, opened 2022-12-18
Set SendEnv conditionally

clintonb:
Empty values break Git+SSH. If `ssh_send_env` is empty, `SendEnv` should not be set in SSH config.

This is especially needed for Ubuntu 22.04 (#1392)


swalkinshaw:
Thank you!

PR #1467, opened 2023-01-24
🔒️ Disable xmlrpc by default

retlehs:
This PR disables `xmlrpc.php` by default

### DIFF
--- a/group_vars/development/wordpress_sites.yml
+++ b/group_vars/development/wordpress_sites.yml
@@ -17,3 +17,5 @@ wordpress_sites:
       provider: self-signed
     cache:
       enabled: false
+    xmlrpc:
+      enabled: false
--- a/group_vars/production/wordpress_sites.yml
+++ b/group_vars/production/wordpress_sites.yml
@@ -19,3 +19,5 @@ wordpress_sites:
       provider: letsencrypt
     cache:
       enabled: false
+    xmlrpc:
+      enabled: false
--- a/group_vars/staging/wordpress_sites.yml
+++ b/group_vars/staging/wordpress_sites.yml
@@ -19,3 +19,5 @@ wordpress_sites:
       provider: letsencrypt
     cache:
       enabled: false
+    xmlrpc:
+      enabled: false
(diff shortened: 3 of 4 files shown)


dalepgrant:
Funny. I spent yesterday checking this out after seeing increased activity on a few of our servers. There is also https://github.com/ItinerisLtd/trellis-disable-xml-rpc. I do like being able to set this per site though as this PR implements

retlehs:
There's also some new-ish fail2ban rules you can enable FYI:

https://github.com/roots/trellis/blob/5e47bd94d82ded8ccc579dc15f836cf4501011c8/group_vars/all/security.yml#L16-L22

dalepgrant:
There are, thanks Ben 🙏 that's where we ended up yesterday with a `todo` to look into completely disabling it today/next week. Gotta check all the sites aren't using it first though, AFAIK JetPack still uses it. Top of my head I think we're good as it's not a plugin we'd normally use but a blanket ban without checking is probably a bad idea. Probably.

swalkinshaw:
Jetpack using this makes me slightly iffy about disabling it by default, but probably in favour anyway. Regardless, better to have the option built in.

PDowney:
Jetpack does still use this, but you can whitelist their IP address ranges:

https://jetpack.com/support/how-to-add-jetpack-ips-allowlist/
It would look something like this:

```location = /xmlrpc.php {
    # Whitelist Jetpack IP ranges
    allow 122.248.245.244/32;
    allow 54.217.201.243/32;
    allow 54.232.116.4/32;
    allow 192.0.80.0/20;
    allow 192.0.96.0/20;
    allow 192.0.112.0/20;
    allow 195.234.108.0/22;

    # Deny all other requests
    deny all;
   }

strarsis:
@PDowney: Thanks! So when Jetpack is used, `xmlrpc.enabled` is set to `true` and then these `fail2ban` rules are added instead, correct?

swalkinshaw:
That seems correct to me 👍

PR #1470, opened 2023-02-07
Install wp as the default user

### LINKED ISSUE #1446: Bug: uploads subfolders ownership
### Terms

#### What's wrong?
In a very specific use case (installing a fresh wp instance in a circleCI environment), I run the wordpress-install role.
WP is installed with the `--allow-root` flag, which leads to the creation of a empty subfolder in `uploads`, corresponding to the current month (eg: `uploads/2022/11`). This folder is owned by root because of the `--allow-root` flag.

In my use case I want to do some changes in this folder afterwards, being the default user (`circleci`), and I can't because the folder is owned by `root`.

#### What have you tried?
Running the role again, with  `--tags=wordpress-install-directories` flag enabled, solves the issue. The `uploads/2022/11` is then owned by the default user.

#### Possible solutions
In the wordpress-install role, we could run `- import_tasks: directories.yml` after the `wp core install` part. This would prevent such issues.

### Steps To Reproduce

Run the wordpress-install role in another environment than the development VM.

### Expected Behavior

Wordpress files and folder are owned by the default user.

### Actual Behavior

uploads subfolders are owned by root.

### Versions

trellis 1.9.1

LucasDemea:
Fixes #1446

### DIFF
--- a/roles/wordpress-install/tasks/main.yml
+++ b/roles/wordpress-install/tasks/main.yml
@@ -50,6 +50,7 @@
            --admin_user="{{ item.value.admin_user | default('admin') }}"
            --admin_password="{{ vault_wordpress_sites[item.key].admin_password }}"
            --admin_email="{{ item.value.admin_email }}"
+  become: no
   args:
     chdir: "{{ www_root }}/{{ item.key }}/{{ item.value.current_path | default('current') }}/"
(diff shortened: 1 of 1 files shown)


swalkinshaw:
Thanks Lucas!

PR #1478, opened 2023-02-22
fix: trim strict-transport-security header

### LINKED ISSUE #1477: Bug: latest curl version fails to retrieve urls on trellis managed hosts
### Terms

### Description

I faced an issue on my satispress server hosted on a trellis managed host.

`curl https://satis.mydomain.com` was failing with this error:
```console
curl: (92) HTTP/2 stream 0 was not closed cleanly: PROTOCOL_ERROR (err 1)
```

This was preventing composer (which uses curl) from fetching updates from this server when it was run from a client with curl 7.85.0 installed (ubuntu 22.10 default).

The culprit in trellis is this line:
https://github.com/roots/trellis/blob/c9fa84141c6bca35515bdeddaa71b64e1a04bce5/roles/wordpress-setup/templates/wordpress-site.conf.j2#L86-L88

It results in a response header containing a trailing whitespace, which is now considered a RFC violation.
image

> This is nghttp2 1.49.0 that started to reject header fields with trailing whitespace, as that is an RFC violation. Also discussed here: #9479
> 
> The next nghttp2 release (1.50.0) has an option to switch off that behavior and curl will use that if available. But nghttp2 1.50.0 has not been released yet and neither has a curl version with that added logic.
> 
> This is not a curl bug. This is changed behavior in nghttp2. Our advice is to either downgrade nghttp2 to 1.48.0 again for the time being, or to apply a patch to nghttp2 to make it go back to its previous behaviors.

https://github.com/curl/curl/issues/9526#issuecomment-1250088023

### Steps To Reproduce

1. install curl 7.85.0 locally
2. `curl https://example.com` (server must be managed by trellis and have an ssl provider != 'self-signed') 

### Expected Behavior

curl returns page content.

### Actual Behavior

`curl: (92) HTTP/2 stream 0 was not closed cleanly: PROTOCOL_ERROR (err 1)`

### Relevant Log Output

_No response_

### Versions

1.20.0

LucasDemea:
Fixes #1477.

### DIFF
--- a/roles/wordpress-setup/templates/wordpress-site.conf.j2
+++ b/roles/wordpress-setup/templates/wordpress-site.conf.j2
@@ -84,7 +84,7 @@ server {
   ssl_buffer_size 1400; # 1400 bytes to fit in one MTU
 
   {% if item.value.ssl.provider | default('manual') != 'self-signed' -%}
-  add_header Strict-Transport-Security "max-age={{ [hsts_max_age, hsts_include_subdomains, hsts_preload] | reject('none') | join('; ') }}";
+  add_header Strict-Transport-Security "max-age={{ [hsts_max_age, hsts_include_subdomains, hsts_preload] | reject('none') | join('; ') | trim }}";
   {% endif -%}
 
   {% if item.value.ssl.client_cert_url is defined -%}


roles/wordpress-setup/templates/wordpress-site.conf.j2
```diff
@@ -84,7 +84,7 @@ server {
   ssl_buffer_size 1400; # 1400 bytes to fit in one MTU
 
   {% if item.value.ssl.provider | default('manual') != 'self-signed' -%}
-  add_header Strict-Transport-Security "max-age={{ [hsts_max_age, hsts_include_subdomains, hsts_preload] | reject('none') | join('; ') }}";
+  add_header Strict-Transport-Security "max-age={{ [hsts_max_age, hsts_include_subdomains, hsts_preload] | reject('none') | join('; ') | trim }}";
```
swalkinshaw, line 87: This might work as well if you want to try:

```
add_header Strict-Transport-Security "max-age={{ [hsts_max_age, hsts_include_subdomains, hsts_preload] | reject | join('; ')
```

LucasDemea: I did try, but it wasn't working. It was returning an empty list.

swalkinshaw:
Thanks @LucasDemea

mariusmateoc:
Hi @swalkinshaw @LucasDemea,

If I didn't set `hsts_include_subdomains: true` I get invalid header `strict-transport-security: max-age=31536000; ;` _(note double "; ;")_.

I don't know why `reject('none')` doesn't reject false/empty. I fix it with the following code: 
```
add_header Strict-Transport-Security "max-age={{ [hsts_max_age, hsts_include_subdomains, hsts_preload] | reject('none') | reject('equalto','') | join('; ') | trim }}";
```

PR #1483, opened 2023-04-04
Add conditional dry_run

MWDelaney:
Sometimes I want to be able to check if Trellis can deploy, including whether it can run its build hooks, install Composer dependencies, etc, without actually deploying. A use case example is a check using a GitHub workflow.

Test with the following:
```
$ trellis deploy --extra-vars "dry_run=true" production
```

I was able to get this working using the following GitHub workflow(s)
```yml
# .github/workflows/dry-run.yml
name: Dry-run deploy to target branch
run-name: Dry-run deploy to target branch

on:
  pull_request:
    branches: [staging, production]

jobs:
  dry-run:
    uses: ./.github/workflows/trellis-cli.yml
    secrets: inherit
    with:
      extra-vars: '"dry_run=true"'
```

```yml
# .github/workflows/trellis-cli.yml

name: trellis-cli
run-name: 'Trellis CLI'

on:
  workflow_call:
    inputs:
      extra-vars:
        required: false
        type: string

jobs:
  trellis-cli:
    runs-on: ubuntu-latest

    steps:
    - uses: actions/checkout@v2
    - uses: andstor/file-existence-action@v2
      id: check_files
      with:
        files: "trellis/deploy.yml"
    - uses: shimataro/ssh-key-action@v2
      with:
        key: ${{ secrets.TRELLIS_DEPLOY_SSH_PRIVATE_KEY }}
        known_hosts: ${{ secrets.TRELLIS_DEPLOY_SSH_KNOWN_HOSTS }}
    - uses: webfactory/ssh-agent@v0.5.4
      with:
        ssh-private-key: ${{ secrets.TRELLIS_DEPLOY_SSH_PRIVATE_KEY }}
    - uses: roots/setup-trellis-cli@v1
      with:
        repo-token: ${{ secrets.GITHUB_TOKEN }}
        ansible-vault-password: ${{ secrets.ANSIBLE_VAULT_PASSWORD }}
    - name: Deploy
      if: steps.check_files.outputs.files_exists == 'true'
      run: trellis deploy --extra-vars ${{ inputs.extra-vars }} ${{ github.base_ref }}
```

### DIFF
--- a/roles/deploy/tasks/main.yml
+++ b/roles/deploy/tasks/main.yml
@@ -17,6 +17,11 @@
 - import_tasks: update.yml
 - import_tasks: prepare.yml
 - import_tasks: build.yml
+
+- name: "Stop here if this is a dry-run"
+  meta: end_play
+  when: dry_run|default(False)
+
 - import_tasks: share.yml
 - import_tasks: finalize.yml
 
--- a/roles/deploy/tasks/main.yml
+++ b/roles/deploy/tasks/main.yml
@@ -18,14 +18,7 @@
 - import_tasks: prepare.yml
 - import_tasks: build.yml
 
-- name: "Delete the deploy if this is a dry-run"
-  ansible.builtin.file:
-    state: absent
-    path: /tmp/trellis
-  when: dry_run|default(False)
-
-- name: "Stop here if this is a dry-run"
-  meta: end_play
+- include_tasks: dryrun.yml
   when: dry_run|default(False)
 
 - import_tasks: share.yml
--- a/roles/deploy/tasks/main.yml
+++ b/roles/deploy/tasks/main.yml
@@ -18,6 +18,12 @@
 - import_tasks: prepare.yml
 - import_tasks: build.yml
 
+- name: "Delete the deploy if this is a dry-run"
+  ansible.builtin.file:
+    state: absent
+    path: /tmp/trellis
+  when: dry_run|default(False)
+
 - name: "Stop here if this is a dry-run"
   meta: end_play
   when: dry_run|default(False)
--- a/roles/deploy/tasks/dryrun.yml
+++ b/roles/deploy/tasks/dryrun.yml
@@ -1,0 +1,12 @@
+- name: "Delete the deploy if this is a dry-run"
+  ansible.builtin.file:
+    state: absent
+    path: "{{ deploy_helper.new_release_path }}"
+
+- name: "Clean up if this is a dry-run"
+  ansible.builtin.file:
+    state: absent
+    path: "{{ project_root }}/shared"
+
+- name: "Stop here if this is a dry-run"
+  meta: end_play
--- a/roles/deploy/tasks/dryrun.yml
+++ b/roles/deploy/tasks/dryrun.yml
@@ -3,10 +3,5 @@
     state: absent
     path: "{{ deploy_helper.new_release_path }}"
 
-- name: "Clean up if this is a dry-run"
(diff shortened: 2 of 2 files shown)


roles/deploy/tasks/main.yml
```diff
@@ -17,6 +17,11 @@
 - import_tasks: update.yml
 - import_tasks: prepare.yml
 - import_tasks: build.yml
+
+- name: "Stop here if this is a dry-run"
+  meta: end_play
```
swalkinshaw, line 22: TIL! Had no idea about the `meta` module 🆒

swalkinshaw (commented):
My main question is naming. "dry run" can mean a lot, but in some other software I know about (like rsync) it only shows you what _would_ have happened.

In this case, we're still creating a deploy, just not finalizing it. So one option is `trellis deploy --without-finalize production` or `--no-finalize`. That way it's consistent with the terms we use in Trellis?

roles/deploy/tasks/main.yml
swalkinshaw: Can you let me know what the output of this looks like as well? I'm wondering if it's useful to have a separate more descriptive message or if this is good enough.

MWDelaney:
Would there be a practical way to divert the deploy directory to a temporary location so that it doesn't interrupt the chain of deploys?

roles/deploy/tasks/main.yml
MWDelaney: GitHub's output ends after the previous task:
image

But locally it shows the following:
```
TASK [deploy : include_tasks] **************************************************
included: /path/to/example.com/trellis/roles/deploy/hooks/build-after.yml for example.com => (item=/path/to/example.com/trellis/roles/deploy/hooks/build-after.yml)

TASK [deploy : Check for composer.json in project root or project_subtree_path] ***
ok: [example.com]

TASK [deploy : Fail if composer.json not found] ********************************
skipping: [example.com]

TASK [deploy : Setup composer authentications] *********************************
ok: [example.com] => (item=None)
ok: [example.com]

TASK [deploy : Run composer check] *********************************************
changed: [example.com]

TASK [deploy : Install Composer dependencies] **********************************
changed: [example.com]

TASK [deploy : Stop here if this is a dry-run] *********************************

PLAY RECAP *********************************************************************
example.com : ok=26   changed=12   unreachable=0    failed=0    skipped=26   rescued=0    ignored=0   
localhost                  : ok=0    changed=0    unreachable=0    failed=0    skipped=1    rescued=0    ignored=0   


```

swalkinshaw:
Overriding the `project_root` variable should theoretically do that. If could be set to a dir in `/tmp`?

MWDelaney:
Seems to work...

```sh
$ trellis deploy --extra-vars "dry_run=true project_root=/tmp/trellis" staging
```

```sh
root@example.com:/tmp/trellis# ls
releases  shared
```

swalkinshaw:
> Is there an inventory that's being updated for rollback somewhere that this going to interfere with?

I don't think so because it's in a different path, so the normal deploy help wouldn't even know about it.

Deploys that aren't finalized get an `unfinished` file written to them, but again, this would be in a different path so it's outside of the deploy cleanup process. One small risk is that they wouldn't be cleaned up by default?

swalkinshaw:
Deleting it right after works 👍 (maybe with a variable to optionally skip that)

MWDelaney:
This hard-codes a path that I'm setting in `--extra-vars` though

PR #1487, opened 2023-05-11
Update deploy.yml

E-VANCE:
Fix doc link in "Deploy WP site"-step message. URL should be redirected IMO additionally.


swalkinshaw:
Agreed, thanks.

PR #1490, opened 2023-05-26
Add Libvirt to the Vagrant config file

Simbaclaws:
Gives the ability for users who have the libvirt plugin installed in vagrant on linux to use libvirt for creating the virtual machine. This change will make sure that the RAM and CPU count will be properly set on each trellis up. Libvirt is working properly so far, but is causing a kernel panic on trellis up because the RAM is set to 512. Currently I have to manually set the RAM to a higher number in order to be able to boot the machine.

I haven't tested this change yet, I'm about to do that right now.

Please add this when people have confirmed it working (not just me).

Basically for this to work, all you have to do is follow the setup instructions for your distribution here: https://vagrant-libvirt.github.io/vagrant-libvirt/

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -208,4 +208,11 @@ Vagrant.configure('2') do |config|
     h.enable_virtualization_extensions = true
     h.linked_clone = true
   end
+
+  # Libvirt/KVM settings
+  config.vm.provider 'libvirt' do [lv]
+    lv.vmname = config.vm.hostname
+    lv.cpus = vconfig.fetch('vagrant_cpus')
+    lv.memory = vconfig.fetch('vagrant_memory')
+  end
 end


swalkinshaw:
This is a good addition, thanks @Simbaclaws

Simbaclaws:
one more thing I thought I'd mention is that you might need to set the ```VAGRANT_DEFAULT_PROVIDER``` env variable to libvirt in order for trellis to pickup libvirt as the provider

Vagrantfile
```diff
@@ -208,4 +208,11 @@ Vagrant.configure('2') do |config|
     h.enable_virtualization_extensions = true
     h.linked_clone = true
   end
+
+  # Libvirt/KVM settings
+  config.vm.provider 'libvirt' do [lv]
```
talss89, line 213: Just at a glance, shouldn't this be `config.vm.provider 'libvirt' do |lv|`? (pipes, not squares)

Simbaclaws: that might very well be the case, I might've miss-typed or didn't notice it needed pipes... Like I said, wait with adding this until it has been tested :)

Simbaclaws: Yeah, I think you are right here, I might've broken the vagrant file with this change. I'll make a quick update...

Simbaclaws: Please see #1491 for a fix @swalkinshaw

Simbaclaws: Thank you for finding this @talss89 :+1:

PR #1491, opened 2023-05-27
@simbaclaws made a mistake using square brackets instead of pipes in the vagrantfile...

Simbaclaws:
I made a mistake where I had used square brackets instead of pipes, potentially breaking the Vagrantfile. I'm sorry....

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -211,7 +211,7 @@ Vagrant.configure('2') do |config|
 
   # Libvirt/KVM settings
   config.vm.provider 'libvirt' do |lv|
-    lv.vmname = config.vm.hostname
+    lv.title = config.vm.hostname
     lv.cpus = vconfig.fetch('vagrant_cpus')
     lv.memory = vconfig.fetch('vagrant_memory')
   end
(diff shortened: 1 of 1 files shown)


Simbaclaws:
I did an actual test this time (after making some changes) to see whether it is working. Here are some screenshots of libvirt working:

Changed the memory

Running site

Simbaclaws:
There seems to be a slight ttfb of 1-2 seconds, which I'm trying to fix but not really sure where it is coming from in the VM...

swalkinshaw:
Thanks for the quick fix (+ @talss89) Also my fault for not noticing that.

The TTFB issue could be a lot of things but I'll merge this for now and if there's any Vagrant setting that improves that a follow-up fix is good.

Simbaclaws:
I just added the latest commit to change the accelerator for the VM to KVM, this should make the virtual machine a lot faster, I'm hoping that fixes the TTFB issue. I'm testing that right now. This does mean that whoever uses libvirt needs kvm and qemu both installed on their machine.

PR #1492, opened 2023-05-29
Update Bitbucket Cloud SSH Host Keys

rkaalma:
Since encrypted copies of Bitbucket’s SSH host keys were included in a data breach of a third-party credential management vendor, Bitbucket issued two new SSH host keys and will be replacing the current host keys on June 20, 2023.

More about the topic: https://bitbucket.org/blog/ssh-host-key-changes


swalkinshaw:
Thank you!

PR #1509, opened 2023-10-27
Move from .test domain to .local

peebeebee:
Chrome can't handle .test domains well. 
See: 
- https://discourse.roots.io/t/chrome-wont-load-test-local-site/11628/3
- https://apple.stackexchange.com/q/458722

### DIFF
--- a/group_vars/development/wordpress_sites.yml
+++ b/group_vars/development/wordpress_sites.yml
@@ -5,11 +5,11 @@
 wordpress_sites:
   example.com:
     site_hosts:
-      - canonical: example.test
+      - canonical: example.local
         redirects:
-          - www.example.test
+          - www.example.local
     local_path: ../site # path targeting local Bedrock site directory (relative to Ansible root)
-    admin_email: admin@example.test
+    admin_email: admin@example.local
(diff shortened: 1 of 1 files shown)


retlehs:
The links you've provided don't really provide much to support your claim regarding `.test` having issues in Chrome?

`.test` is specifically a reserved TLD per [RFC-6761](https://www.rfc-editor.org/rfc/rfc6761.html)

`.local` is reserved for multicast DNS per [RFC-6762](https://datatracker.ietf.org/doc/html/rfc6762) and has been known to cause delay issues for folks (more info can be found by searching about `.local` TLD delays with macOS)

peebeebee:
Ok, but I did have issues with .test here.
Going to chrome://settings/security and not using "Use secure DNS" (which was on by default) also did the trick to have the site loading again.
Did not need to do this with a .local domain though.

So maybe the error is with my current service provider?

swalkinshaw:
I'm not sure what was happening in your case but so far it appears specific to your machine since we haven't heard any other reports about it.

Closing for this now unless we hear more.

PR #1511, opened 2024-01-17
Refactor PHP role to support future versions by default

dalepgrant:
Refactors php version vars to be dynamic by default whilst allowing overrides.

### Motivation
The current approach requires a file to be duplicated and renamed for every php version release. At time of writing, the contents of said files covering 7.4 - 8.2 (4 files total) are identical, except for the php version number itself. ~To add support for php 8.3, a fifth file would need to be added.~ _Edit to add: since #1514, this fifth file has now been added._ 

### Proposal
By making the default file dynamic, we can avoid duplicating files each time a new php version is released. If something changes from one release to the next, an override file can be created using the php version as the filename (i.e. we can fallback to the existing method). 

### Considerations
I initially started to put these vars in the `php/defaults/main.yml` file, as these are essentially defaults. However, ultimately I settled on `php/vars/version-specific-defaults.yml` which a) is closer to what developers are currently familiar with, b) separates php version specific defaults from other defaults and c) allows for easy duplicate & rename to override.

I've only tested this on local so far and everything provisions as expected using php 8.3.

Comments have been written with ~8.1~ 8.2 as that is the current default version for Trellis.


swalkinshaw:
🤔  interesting!

I have a few initial concerns:

* this works right now because the list of default packages has remained consistent across versions. Say for example 8.4 required a new package, then we'd have to include `8.4.yml` file right?
* it might be less obvious when Trellis "officially" supports a new PHP version since there'd be code changes

But on the other hand, requiring no code changes for people running Trellis projects is a benefit too and that's probably your main reason for proposing this change.

dalepgrant:
>Say for example 8.4 required a new package, then we'd have to include 8.4.yml file right?

Correct. As it stands though, we have to add a 8.4.yml file even if there are no package changes.

>it might be less obvious when Trellis "officially" supports a new PHP version since there'd be code changes

A suitable PR (and changelog entry) can still change the ['out of the box' php version](https://github.com/roots/trellis/blob/36c8491cdd813bba19853e8d2976cb4690d2c116/group_vars/all/main.yml#L6) to highlight official support, like #1484. We could also add a line to the readme and/or the docs stating something along the lines of "Trellis officially supports PHP 7.4 - 8.2, more recent versions may also work but have not yet been tested by the community."

dalepgrant:
>As it stands though, we have to add a 8.4.yml file even if there are no package changes.

8.4 added in #1560

swalkinshaw:
@dalepgrant want to rebase this? I think it's still an improvement over all

dalepgrant:
@swalkinshaw done, and thank you 🙏  I was looking for somewhere in the docs to put a note about this and how to change it, but there's nowhere obvious. Best place I could come up with was [here](https://roots.io/trellis/docs/ansible/#variables) but you'd only find that by accident. Would you be happy with a new PHP page (I can PR)? Or is there somewhere better to put this/don't think documentation is needed?

swalkinshaw:
A new page for configuring PHP sounds good actually. We can expand it with the other options after too 👍

dalepgrant:
📖 https://github.com/roots/docs/pull/536

swalkinshaw:
Thanks @dalepgrant

PR #1515, opened 2024-04-18
🐛 Fix MariaDB apt-key is deprecated failure

retlehs:
Fixes the issue where MariaDB installation fails on the "Add MariaDB APT key" task

Ref https://discourse.roots.io/t/trellis-task-mariadb-add-maria-apt-key-apt-key-is-deprecated/26706
Ref https://discourse.roots.io/t/apt-key-deprecated-should-playbooks-be-updated/26761

### DIFF
--- a/roles/mariadb/tasks/main.yml
+++ b/roles/mariadb/tasks/main.yml
@@ -11,21 +11,10 @@
   args:
     removes: "/tmp/mariadb_release_signing_key.asc"
 
-- name: Setup MariaDB repository
-  ansible.builtin.copy:
-    dest: "/etc/apt/sources.list.d/mariadb.list"
-    content: |
-      # Ansible managed
-      {{ mariadb_ppa }}
-    mode: "0644"
-
-- name: Install necessary packages for repository management
-  ansible.builtin.apt:
-    name:
-      - apt-transport-https
-      - gnupg
-    state: present
-    update_cache: true
+- name: Add MariaDB PPA
+  apt_repository:
+    repo: "{{ mariadb_ppa }}"
+    update_cache: yes
 
 - name: Install MySQL client
   ansible.builtin.apt:
--- a/roles/mariadb/tasks/main.yml
+++ b/roles/mariadb/tasks/main.yml
@@ -1,24 +1,41 @@
 ---
-- block:
-  - name: Add MariaDB APT key
-    apt_key:
-      keyserver: "{{ mariadb_keyserver }}"
-      id: "{{ mariadb_keyserver_id }}"
+- name: Download MariaDB GPG key
+  ansible.builtin.get_url:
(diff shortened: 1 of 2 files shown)


roles/mariadb/tasks/main.yml
```diff
@@ -1,24 +1,41 @@
...
+  ansible.builtin.get_url:
+    url: "https://mariadb.org/mariadb_release_signing_key.asc"
+    dest: "/tmp/mariadb_release_signing_key.asc"
+    mode: '0644'
+
+- name: Add the MariaDB GPG key to apt
+  ansible.builtin.shell:
+    cmd: "apt-key add /tmp/mariadb_release_signing_key.asc"
+  args:
+    removes: "/tmp/mariadb_release_signing_key.asc"
```
swalkinshaw, line 12: Can we use the built-in url feature of `apt_key`? https://docs.ansible.com/ansible/latest/collections/ansible/builtin/apt_key_module.html

```yaml
- name: Add an Apt signing key, uses whichever key is at the URL
  ansible.builtin.apt_key:
    url: https://ftp-master.debian.org/keys/archive-key-6.0.asc
    state: present
```

roles/mariadb/tasks/main.yml
```diff
@@ -1,24 +1,41 @@
...
+
+- name: Setup MariaDB repository
+  ansible.builtin.copy:
+    dest: "/etc/apt/sources.list.d/mariadb.list"
+    content: |
+      # Ansible managed
+      {{ mariadb_ppa }}
+    mode: "0644"
```
swalkinshaw, line 20: I'm curious if the old task still works here:

```yaml
  - name: Add MariaDB PPA
    apt_repository:
      repo: "{{ mariadb_ppa }}"
      update_cache: yes
```

roles/mariadb/tasks/main.yml
retlehs:
```
fatal: [default]: FAILED! => {"after": ["D94AA3F0EFE21092", "871920D1991BC93C"], "before": ["D94AA3F0EFE21092", "871920D1991BC93C"], "changed": true, "fp": "AED4B06F473041FA", "id": "AED4B06F473041FA", "key_id": "AED4B06F473041FA", "msg": "apt-key did not return an error, but failed to add the key (check that the id is correct and *not* a subkey)", "short_id": "473041FA"}
```

tangrufus: Try https://mariadb.com/kb/en/installing-mariadb-deb-files-with-ansible/#importing-mariadb-gpg-key


```yaml
- name: Add an apt key by id from a keyserver
  ansible.builtin.apt_key:
    keyserver: hkp://keyserver.ubuntu.com:80
    id: 0xF1656F24C74CD1D8
```

swalkinshaw (approved):

PR #1518, opened 2024-05-30
Fix Nginx apt-key is deprecated failure

Twansparant:
🐛 Fix NginxDB apt-key is deprecated failure

* Use apt_repository
* Use apt_key module

### DIFF
--- a/roles/nginx/defaults/main.yml
+++ b/roles/nginx/defaults/main.yml
@@ -1,6 +1,4 @@
 ---
-nginx_keyserver: "https://nginx.org/keys/nginx_signing.key"
-nginx_keyserver_id: "ABF5BD827BD9BF62"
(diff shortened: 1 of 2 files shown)


swalkinshaw (approved):

paulbrzeski:
Hey Trellis devs, I think I've spotted a race conditions issue with this fix.

I just tried to reprovision a box using a version of Trellis containing the fix but it still failed due on apt update due to deprecated nginx keys. Ran the fix manually on the box and it worked, i.e. `wget https://nginx.org/keys/nginx_signing.key -O - | sudo apt-key add -`

Checked out how `server.yml` is structured. The `[common]` task runs first and that contains the `apt update`. If anyone has invalid keys blocking the update then they won't get up to the `[nginx]` task to do the key update. Checked the issues list and it's not been rerecorded so I'm just gonna logging it here

Not sure what the fix would be... Maybe if there's an option to add an error handling step to Ansible.

Cheers!

PR #1521, opened 2024-06-07
Generate list of hosted development sites after `vagrant up`

MWDelaney:


lib/trellis/vagrant.rb
```diff
@@ -79,6 +79,22 @@ def post_up_message
   msg << "\n* You can SSH into the machine with `vagrant ssh`."
   msg << "\n* Then navigate to your WordPress sites at `/srv/www`"
   msg << "\n  or to your Trellis files at `#{ANSIBLE_PATH_ON_VM}`."
+  sites = YAML.load_file(File.join(ANSIBLE_PATH, 'group_vars/development/wordpress_sites.yml'))
+  wordpress_sites = sites['wordpress_sites']
```
swalkinshaw, line 83: This is already available as `trellis_config.wordpress_sites`

lib/trellis/vagrant.rb
```diff
@@ -79,6 +79,22 @@ def post_up_message
...
+  if wordpress_sites.any?
+    msg << "\n\n\e[35m🚀 Development URLs: \e[0m"
+    wordpress_sites.each do |site_name, site|
```
swalkinshaw, line 86: `trellis_config.wordpress_sites.each_pair do |site_name, site|`

lib/trellis/vagrant.rb
MWDelaney: trellis_config wasn't available so I loaded it the same way the Vagrantfile does.

swalkinshaw: Did you actually get an error? This method is only used in the `Vagrantfile` which has `trellis_config` in scope.

If for some reason it doesn't work, you should change the `post_up_message` method to accept an `trellis_config` argument so it can be passed in.

lib/trellis/vagrant.rb
```diff
@@ -72,14 +72,29 @@ def mount_options(mount_type, dmode:, fmode:)
...
   msg << "\n* You can SSH into the machine with `vagrant ssh`."
   msg << "\n* Then navigate to your WordPress sites at `/srv/www`"
   msg << "\n  or to your Trellis files at `#{ANSIBLE_PATH_ON_VM}`."
 
+  if trellis_config.wordpress_sites.any?
+    msg << "\n\n\e[35m🚀 Development URLs: \e[0m"
+    trellis_config.wordpress_sites.each do |site_name, site|
+      site_url = site['site_hosts'].first
+      site_url = site_url.is_a?(Hash) ? site_url['canonical'] : site_url
+      site_url = site['ssl']['enabled'] ? "https://#{site_url}" : "http://#{site_url}"
```
swalkinshaw, line 88: Not sure if its possible, but to be very defensive this could be:

```ruby
site.dig('ssl', 'enabled') ? "https://#{site_url}" : "http://#{site_url}"
```

That way it will gracefully handle either ssl or enabled not existing

MWDelaney: Fixed, and I changed emojis for more consistent spacing in different terminals.

PR #1527, opened 2024-06-29
Do not run Composer as root/super user!

adleviton:
I resolved this warning by adding `become: no` to composer authentications for "BitBucket OAuth" and "Other Tokens". It was already present for "HTTP Basic".


swalkinshaw:
Thank you!

PR #1529, opened 2024-07-23
Use `http2` and `http3` `nginx` directives

strarsis:
This PR uses the `http2` and `http3` directives instead the deprecated `http2` keyword in `listen` directive.
`HTTP 2` and `HTTP 3` support can be controlled using the ansible `http2_enabled` and `http3_enabled` options.

Using the [`h5bp` `nginx` configuration](https://github.com/h5bp/server-configs-nginx/pull/335/files#diff-1fef7bb78ec8a7315a2a8d18eaebf796065a236e3f17d4d1b6bae8b89965ed39) as base.

As some may still be hesitant to enabling HTTP 3 support in nginx, the
option for HTTP 3 is turned off by default and can be turned on manually.

### DIFF
--- a/roles/wordpress-setup/defaults/main.yml
+++ b/roles/wordpress-setup/defaults/main.yml
@@ -7,6 +7,8 @@ nginx_sites_confs:
   - src: ssl.no-default.conf.j2
     enabled: false
 
+http2_enabled: true
+
 # HSTS defaults
 nginx_hsts_max_age: 31536000
 nginx_hsts_include_subdomains: false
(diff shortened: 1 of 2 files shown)


roles/wordpress-setup/defaults/main.yml
```diff
@@ -7,6 +7,9 @@ nginx_sites_confs:
   - src: ssl.no-default.conf.j2
     enabled: false
 
+http2_enabled: true
+http3_enabled: false
```
swalkinshaw, line 11: I think we should prefix with `nginx_` for consistency?

PR #1530, opened 2024-07-29
Conditionally add HTTPS inbound allow firewall rule

strarsis:
This PR lets the firewall rule to allow inbound HTTPS to be added conditionally.
Only when at least one site uses HTTPS, the firewall rule for allowing inbound HTTPS is added.

### DIFF
--- a/roles/ferm/defaults/main.yml
+++ b/roles/ferm/defaults/main.yml
@@ -11,3 +11,5 @@ ferm_default_policy_forward: DROP
 ferm_input_list: []
 ferm_input_group_list: []
 ferm_input_host_list: []
+
+sites_using_ssl: "[{% for name, site in wordpress_sites.items() | list if site.ssl.enabled %}'{{ name }}',{% endfor %}]"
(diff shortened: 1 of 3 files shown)


roles/ferm/defaults/main.yml
```diff
@@ -11,3 +11,5 @@ ferm_default_policy_forward: DROP
 ferm_input_list: []
 ferm_input_group_list: []
 ferm_input_host_list: []
+
+sites_using_ssl: "[{% for name, site in wordpress_sites.items() | list if site.ssl.enabled %}'{{ name }}',{% endfor %}]"
```
swalkinshaw, line 15: You should be able to re-use the existing `sites_use_ssl` here?

strarsis: Oh, I looked for such a variable, but had not found it. Re-using it is much better of course.
After some testing I noticed that just conditionally adding the allow rule would not make ansible clean it up if no site is using HTTPS (in subsequent runs).
Then I noticed the `delete` attribute of the individual ferm rules, which is ideal as it not only toggles the present of the HTTPS allow rule, but also cleans it up if it should not apply anymore (as no sites are using SSL).

strarsis:
Edit: Using the `delete` attribute as this allows toggling _and_ proper clean up of file when HTTPS firewall allow rule is not needed (in subsequent provisions), it also allows for much simpler configuration code.

swalkinshaw:
Thanks @strarsis

PR #1542, opened 2024-08-16
Allow overriding "pm" in PHP-FPM's pool config

gutobenn:
This PR adds support for overriding the value of 'pm' in PHP-FPM's pool config.

This can be done by setting a value to the variable `php_fpm_pm`, just like you can already do to other configs.

Example:
```
# group_vars/all/php.yml
php_fpm_pm: 'static'
```


swalkinshaw:
> allowing different settings for each environment.

The goal is for each environment (eg: production, staging, etc) and **not** per site? If so, this might be confusing since usually parameters under a site are meant to be site specific, but the config created is not site specific.

gutobenn:
@swalkinshaw Thanks for the feedback! I've reformulated my solution.

swalkinshaw:
Thanks @gutobenn

PR #1552, opened 2024-11-23
Use proper `$host` for FastCGI `SERVER_NAME` param

### LINKED ISSUE #1550: PHP `SERVER_NAME` not passed correctly for multiple canonical domains
### Version

1.22.1

### What did you expect to happen?

PHP `$_SERVER['SERVER_NAME']` is assigned the correct domain that is requested, 
not the first (or random?) one in `server_name` in `nginx` site configuration.

This prevents redirection from redirection plugins.

### What actually happens?

PHP `$_SERVER['SERVER_NAME']` is assigned incorrect domain, as the first (or random?) one in `server_name` in `nginx` site configuration.

### Steps to reproduce

1. Set up a site in Trellis with two or more canonical domains (not to be confused with redirections).
2. Request one of the canonical domains.
3. Debug the request (e.g. `print_r($_SERVER);` in `index.php`). 
Note that the `SERVER_NAME` field in `$_SERVER` is not correct for all other requested canonical domains except for one (usually the first listed one).

Note that the `SERVER_NAME` is not always the first canonical domain, apparently this can change, which caused issues with a redirection plugin that redirected from the wrong canonical domain.

### System info

Ubuntu LTS

### Log output

_No response_

### Please confirm this isn't a support request.

Yes

### Context
In order to fix the passed `SERVER_NAME`, `$host` instead of `$server_name` has to be used:
````
fastcgi_param SERVER_NAME         $host;
````

In `fastcgi_params` configuration file included by the sites, `$server_name` is used:
````
fastcgi_param  SERVER_NAME        $server_name;
````

And `$server_name` is the wrong value:
>- $host contains "in this order of precedence: host name from the request line, or host name from the 'Host' request header field, or the server name matching a request"
> - $http_host contains the content of the HTTP "Host" header field, if it was present in the request
> - $server_name contains the server_name of the virtual host which processed the request, as it was defined in the nginx configuration. If a server contains multiple server_names, only the first one will be present in this variable.

(from [Michael Hampton](https://serverfault.com/users/126632/michael-hampton) in https://serverfault.com/a/706439/958731).

As `fastctgi_params` is a default configuration file shipped with `nginx`, it makes sense to 
override the `SERVER_NAME` field after that file was included, in order to override it effectively:
````
include fastcgi_params;
[...]
fastcgi_param SERVER_NAME $host;
````

strarsis:
This PR adds a `fastcgi_param` line to wordpress site configuration to ensure that `SERVER_NAME` FastCGI (PHP) variable is set to `$host`. This is important as the included default `fastcgi_params` set `SERVER_NAME` to `$server_name` which is set (arbitrarily) to one of the matching hostnames of a site, but not the actual matching one.
Fixes #1550.

### DIFF
--- a/roles/wordpress-setup/templates/wordpress-site.conf.j2
+++ b/roles/wordpress-setup/templates/wordpress-site.conf.j2
@@ -255,6 +255,7 @@ server {
 
     {% block fastcgi_basic -%}
     include fastcgi_params;
+    fastcgi_param SERVER_NAME $host;
     fastcgi_param SCRIPT_FILENAME $realpath_root$fastcgi_script_name;
     fastcgi_param DOCUMENT_ROOT $realpath_root;
     fastcgi_pass unix:/var/run/php-fpm-wordpress.sock;


swalkinshaw:
Thanks @strarsis

PR #1556, opened 2025-01-22
Mitigate REST API cache poisoning via Method Override

AltanS:
Assuming a website chooses to cache responses for specific API routes in the WP REST API (like /wp-json/wp/v2/posts). An attacker could use the `X-HTTP-Method-Override` header to request an endpoint (i.e. /wp/v2/posts/1) until the cache is `STALE` and trigger an update of the cache from the response that served the HEAD request instead of the GET request. The endpoint would then return an empty response body.

Normal routes are not affected by this, but since WP allows method overrides against the REST API, this could become an issue for headless WP instances.


retlehs (approved):
Appreciate both the PR and the background. This is a nice update — thank you!

tangrufus (approved):

swalkinshaw:
Thank you!

PR #1560, opened 2025-03-06
🔧 PHP 8.4 support

retlehs:
⚠️ WP-CLI doesn't support PHP 8.4 at this time — expect lots of deprecated spam if you're wanting to use PHP 8.4 right now


retlehs:
WP-CLI now has PHP 8.4 compatibility as of v2.12.0, see https://make.wordpress.org/cli/2025/05/07/wp-cli-v2-12-0-release-notes/

PR #1564, opened 2025-03-17
Replace link for creating CSP

Dyras:
Hello!

I woke up this morning and decided to add CSP to the company website. Turns out that the *checks changelog* 7 year old link for this project takes you to a blog now.

So I looked around online for a similar site that is free to use. I found nothing.

Eventually I found a browser extension that allows you to create CSP by browsing around your website. I just tried it and so far it works great.

If you guys have a better site or extension, feel free using that instead :) 

PS. What's going on with the community forum at the moment? I don't have Discord access and the message on the forum stating it's read only temporarily is really bare bones.


retlehs:
We should update this file based off the upstream source (https://github.com/h5bp/server-configs-nginx)

swalkinshaw:
Sorry we weren't 100% clear before. We can't replace this entire file because there's other non-CSP settings. Specifically we should preserve the file as is, but replace the CSP related comment in this one only.

Dyras:
Ah yes, seems to also have been an error from my side. So ready for the Friday to be honest.

swalkinshaw:
Thanks!

PR #1566, opened 2025-03-27
Remove "production" from asset compilation titles

joshuafredrickson:
Clarifies that these steps run in all environments, not just production.

I've had a few heart-stopping moments when deploying to staging and seeing "production" flash on the screen, worried that I accidentally full-sent to the wrong environment. 😅


retlehs (approved):
Thank you!

PR #1567, opened 2025-04-01
Flush rewrite rules at the end of each deploy

joshuafredrickson:
Permalinks aren’t automatically flushed on deployment, so I usually handle it in a deploy hook. Since rewrites function like a cache—similar to Acorn, ACF, or Redis—it makes sense to clear them during deployment.

For example, adding a new CPT requires flushing rewrites manually for the archive page to work. Automating this would help ensure rewrite *just work* after deployment.


swalkinshaw:
Thank you!

PR #1573, opened 2025-05-01
Filter Nginx includes by sites present on target server

dalepgrant:
One for those of us using multiple remote servers with Trellis.

Currently, Trellis templates out anything you put into `nginx-includes/` to the remote servers. Take the following:

```
trellis/
└── group_vars/         
    └── production/
        ├── main.yml
        └── vault.yml
    └── staging/
└── host_vars
    └── server1/
        └── wordpress_sites.yml
    └── server2/
        └── wordpress_sites.yml
    └── staging1/
    └── staging2/
```

Assume server1 has a different site(s) to server2, and perhaps staging1 and staging 2 have additional sites. Your Nginx includes folder might look like this:

```
trellis/
└── nginx-includes/         
    └── all/
        ├── rewrites.conf.j2
        └── proxy.conf.j2
    └── site1/
        └── rewrites.conf.j2
    └── site2/
        └── rewrites.conf.j2
    └── site3/
        └── rewrites.conf.j2
...etc
```

 **All** servers will end up with **all** Nginx includes files, even if the sites don't exist on that server:

```
/etc/nginx/includes.d/
└── all/  
    ├── rewrites.conf
    └── proxy.conf
└── site1/  
    └── rewrites.conf
└── site2/  
    └── rewrites.conf
└── site3/  
    └── rewrites.conf
└── site4/  
    └── rewrites.conf
```

This PR filters the Nginx includes that are to be 'included' by combining the sites that exist on the target server (taken from  wordpress_sites keys) with a default 'all'. This follows with the default setup [shown in the docs](https://roots.io/trellis/docs/nginx-includes/#default). 

~Also included is the ability to override the default 'all' folder. I've only included this because it is possible to override the [block in the Nginx conf template](https://github.com/roots/trellis/blob/d4b6f29648ae0759f319eee8a9e4d4ace684224d/roles/wordpress-setup/templates/wordpress-site.conf.j2#L116-L120) and if someone did that, they may also want to determine their own Nginx includes locations.~

~To override or otherwise add to the Nginx includes locations, in e.g. `group_vars/all/main.yml` you can do~

```yml
# Edit: this change was removed 
```

Note: in testing, the folders for sites not present on a remote are not cleaned up if they already exist but the files inside them are. That's to do with what happens when setting `nginx_includes_d_cleanup: true` which is a separate issue.

### DIFF
--- a/roles/wordpress-setup/tasks/nginx-includes.yml
+++ b/roles/wordpress-setup/tasks/nginx-includes.yml
@@ -19,14 +19,20 @@
     set_fact:
       nginx_include_contexts: "{{ wordpress_sites.keys() | list + nginx_includes_extra_folders_list }}"
 
-  - name: Build list of Nginx includes templates
+  - name: Loop through site keys to build a list of Nginx includes templates per folder
     find:
-      paths: "{{ nginx_includes_templates_path }}"
+      paths: "{{ nginx_includes_templates_path }}/{{ item }}"
+      patterns: "*.conf.j2"
       recurse: yes
-      patterns: "{{ nginx_include_contexts | map('regex_replace', '^(.*)$', '\\1/*.conf.j2') | list }}"
+    loop: "{{ nginx_include_contexts }}"
     become: no
     delegate_to: localhost
-    register: nginx_includes_templates
+    register: nginx_includes_folder_results
+
+  - name: Flatten include template files into one list
+    set_fact:
+      nginx_includes_templates:
+        files: "{{ nginx_includes_folder_results.results | map(attribute='files') | flatten }}"
 
   - name: Create includes.d directories
     file:
--- a/roles/wordpress-setup/tasks/nginx-includes.yml
+++ b/roles/wordpress-setup/tasks/nginx-includes.yml
@@ -13,7 +13,30 @@
       recurse: yes
     become: no
     delegate_to: localhost
-    register: nginx_includes_templates
+    register: nginx_includes_templates_all
+
+  - name: Combine site keys with extra include folders
+    set_fact:
+      nginx_include_contexts: >-
+        {{
+          (wordpress_sites.keys() | list)
+          + (
+              (nginx_includes_extra_folders
+                if nginx_includes_extra_folders is iterable and not nginx_includes_extra_folders is string
+                else [nginx_includes_extra_folders | default('all')]
+              )
+            )
+        }}
+
+  - name: Filter Nginx includes templates by existing sites and global folders
+    set_fact:
+      nginx_includes_templates:
+        files: >-
+          {{
+            nginx_includes_templates_all.files
+            | selectattr('path', 'search', '/(' + nginx_include_contexts | join('|') + ')/')
+            | list
+          }}
 
   - name: Create includes.d directories
     file:
--- a/roles/wordpress-setup/tasks/nginx-includes.yml
+++ b/roles/wordpress-setup/tasks/nginx-includes.yml
@@ -6,15 +6,6 @@
   register: nginx_includes_local_directory
 
 - block:
-  - name: Build list of Nginx includes templates
-    find:
-      paths: "{{ nginx_includes_templates_path }}"
-      pattern: "*.conf.j2"
-      recurse: yes
-    become: no
-    delegate_to: localhost
-    register: nginx_includes_templates_all
-
   - name: Normalize extra include folders to a list
     set_fact:
       nginx_includes_extra_folders_list: >-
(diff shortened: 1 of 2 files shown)


roles/wordpress-setup/tasks/nginx-includes.yml
```diff
@@ -13,7 +13,30 @@
...
+
+  - name: Filter Nginx includes templates by existing sites and extra folders
+    set_fact:
+      nginx_includes_templates:
+        files: >-
+          {{
+            nginx_includes_templates_all.files
+            | selectattr('path', 'search', '/(' + nginx_include_contexts | join('|') + ')/')
+            | list
+          }}
```
swalkinshaw, line 40: Could we use `nginx_include_contexts` to build the patterns for use in the `find`? Basically, move the "Build list of Nginx includes templates" task _after_ this one, and make it look something like this:

```yml
- name: Build list of Nginx includes templates
  find:
    paths: "{{ nginx_includes_templates_path }}"
    patterns: "{{ [item + '/*.conf.j2' for item in nginx_include_contexts] }}"
    recurse: yes
  become: no
  delegate_to: localhost
```

dalepgrant: Yeah that'd be cleaner. Thanks! Will review and push changes up 👍

dalepgrant: >```yaml
>patterns: "{{ [item + '/*.conf.j2' for item in nginx_include_contexts] }}"
>```

This caused an error in testing:

```bash
TASK [wordpress-setup : Build list of Nginx includes templates] *******************************************************************************************************************
fatal: [staging1 -> localhost]: FAILED! => {"msg": "template error while templating string: expected token ',', got 'for'. String: {{ [item + '/*.conf.j2' for item in nginx_include_contexts] }}. expected token ',', got 'for'"}
```

Swapped for 
```yaml
patterns: "{{ nginx_include_contexts | map('regex_replace', '^(.*)$', '\\1/*.conf.j2') | list }}"
```

dalepgrant:
Docs added: https://github.com/roots/docs/pull/544

swalkinshaw:
> Also included is the ability to override the default 'all' folder. I've only included this because it is possible to override the [block in the Nginx conf template](https://github.com/roots/trellis/blob/d4b6f29648ae0759f319eee8a9e4d4ace684224d/roles/wordpress-setup/templates/wordpress-site.conf.j2#L116-L120) and if someone did that, they may also want to determine their own Nginx includes locations.

Is there really a point to this though? Even if someone customized that block, I'm not sure why they'd _change_ the "all" folder. And I'm not sure I see a use case for having more than one when they do the same thing?

Personally I'd rather just omit this extra functionality since I don't see any use case.

dalepgrant:
I have to admit I didn't question the "why" someone might want to, only that they _can_ elsewhere and so that should probably be catered for. Otherwise someone else in a few month's time might stumble across it and raise it as a bug for you guys to spend time reviewing & maybe fixing. 

Doesn't affect me personally so I'm happy to take it out if you say so but given it's already there and keeps that functionality complete, I'd vote for it to stay in. 

Let me know what you decide 👍

swalkinshaw:
Let's remove it and keep this simple ✨

roles/wordpress-setup/tasks/nginx-includes.yml
swalkinshaw: Looks good 👍

dalepgrant:
Ability to override the default 'all' folder has been removed (https://github.com/roots/trellis/pull/1573/commits/2d4741077a63426a6ad0002212c3052072fb4bf0)

Tested the key motivation for this PR:
- ✅ `nginx-includes/example2.com/*.conf.j2` is NOT templated to `/etc/nginx/includes.d/example2.com/*.conf` (where `example2.com` is NOT present in `wordpress_sites.yml`)

Tested existing functionality hasn't been affected: 
- ✅ `nginx-includes/all/*.conf.j2` is templated to `/etc/nginx/includes.d/all/*.conf`
- ✅ `nginx-includes/all/subdir/*.conf.j2` is templated to `/etc/nginx/includes.d/all/subdir/*.conf`
- ✅ `nginx-includes/example1.com/*.conf.j2` is templated to `/etc/nginx/includes.d/example1.com/*.conf` (where `example1.com` is the site key in `wordpress_sites.yml`)
- ✅ Removing a file e.g. `nginx-includes/example3.com/rewrites.conf.j2` removes that file from `/etc/nginx/includes.d/example3.com/rewrites.conf`. It does not remove empty folders.

Issue #1087, opened 2019-06-18
Removing keys from authorized_keys when re-provisioning


jan-muller:
* [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)
* [x] This request isn't a duplicate of an existing request
* [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/) community

## Feature request
When re-provisioning a server with a users tag `trellis provision --tags users env` it does not remove the keys from `authorized_keys`. It would be nice if it could remove keys.

Working with a team of developers, we've put all the keys of all the team members in the `/group_vars/all/users.yml`. Removing a key in `users.yml` and re-provision the server does not remove any keys.

This is something to be aware of, might want to modify the docs with a warning.

Related discourse topic: https://discourse.roots.io/t/re-provisioning-with-tag-users-does-not-removing-ssh-keys-from-authorized-keys/15857

swalkinshaw:
I looked into this and unfortunately it would be very complex to implement. Ansible's `authorized_key` module helpfully [provides](https://docs.ansible.com/ansible/latest/collections/ansible/posix/authorized_key_module.html) a `exclusive` option which will remove old keys automatically, but the way Trellis adds keys makes this very hard (if not impossible) to use.

To use the `exclusive` option, **all** keys need to be provided as a single string blob. But we have two different tasks: https://github.com/roots/trellis/blob/048e5682ddeaf00c0f24b7d82a927d5b00a50b5b/roles/users/tasks/main.yml#L52-L64

Trying to get them all together would also mean not being able to support the `https://github.com/username.keys` URL feature.

If someone wants to try and implement this please feel free. It would be a great addition but I don't see a way to do it.

dalepgrant:
Spitballing an idea, we could do something like this: 

`trellis/roles/users/tasks/main.yml`
```diff
+ - name: Overwrite user SSH keys with primary user
+  authorized_key:
+    user: "{{ item.name }}"
+    key: "{{ item['keys'] | first }}"  # get first SSH key listed in group_vars/all/users.yml
+    exclusive: true                    # remove all other keys
+  loop: "{{ users | default([]) }}"
+ loop_control:
+    label: "{{ item.name }}"

+ - name: Add other user SSH keys
- - name: Add user SSH keys 
   authorized_key: 
     user: "{{ item.0.name }}" 
     key: "{{ item.1 }}" 
   with_subelements: 
[...]

PR #1576, opened 2025-05-14
Allow removal of old SSH keys on provision

dalepgrant:
Rel: roots/trellis#1087

This PR adds the ability to enable removing old SSH keys from the server. 

The added task replaces **all SSH keys on the server** for the users defined in Trellis with the _first found_ SSH key set in `group_vars/all/users.yml`. This allows for cases where the Trellis defaults `~/.ssh/id_rsa.pub` and `~/.ssh/id_ed25519.pub` don't exist. The existing task "Add user SSH keys" then adds the remaining keys as usual. 

Given this has the power to lock a user out of their server, a flag has been added to allow making this an optional task to run. To enable, set `reset_user_ssh_keys: true`. I've placed this in `group_vars/all/security.yml` but arguments can be made for this in `users.yml` or `main.yml`.

### Scenarios
Given the default Trellis keys:

```yaml
  keys:
      - "{{ lookup('file', '~/.ssh/id_rsa.pub', errors='ignore') }}"
      - "{{ lookup('file', '~/.ssh/id_ed25519.pub', errors='ignore') }}"
      # - https://github.com/username.keys
```

If...
- `~/.ssh/id_rsa.pub` does not exist, `~/.ssh/id_ed25519.pub` does exist
  - `keys` becomes `["", "ssh-ed25519 AAAAC3NzaC1.."]` (example)
  - `select('truthy')` filters out empty strings, `keys` becomes `["ssh-ed25519 AAAAC3NzaC1.."]`
  - The value of `~/.ssh/id_ed25519.pub` is used (e.g. `ssh-ed25519 AAAAC3NzaC1..`)
  - The task is run. Note: Until the subsequent (existing) task is run, access is temporarily only available to `~/.ssh/id_ed25519.pub`.
- Neither `~/.ssh/id_rsa.pub` nor `~/.ssh/id_ed25519.pub` exist
  - `keys` becomes `["", ""]`
  - `select('truthy')` filters out empty strings, `keys` becomes `[]`
  - The `when` clause `(item['keys'] | select('truthy') | list | length) > 0` evaluates to false
  - The task is skipped - no SSH keys removed from the server.

### DIFF
--- a/group_vars/all/security.yml
+++ b/group_vars/all/security.yml
@@ -38,8 +38,3 @@ sshd_password_authentication: false
 ip_whitelist:
   - 127.0.0.0/8
   - "{{ ipify_public_ip | default('') }}"
-
-# Only enable this when you want to reset the SSH keys on the server.
-# You probably don't need this enabled for every server provision.
-# Double check your keys are correct in group_vars/all/users.yml first.
-reset_user_ssh_keys: false
--- a/group_vars/all/security.yml
+++ b/group_vars/all/security.yml
@@ -38,3 +38,8 @@ sshd_password_authentication: false
 ip_whitelist:
   - 127.0.0.0/8
   - "{{ ipify_public_ip | default('') }}"
+
+# Only enable this when you want to reset the SSH keys on the server.
+# You probably don't need this enabled for every server provision.
+# Double check your keys are correct in group_vars/all/users.yml first.
+reset_user_ssh_keys: false
--- a/roles/users/tasks/main.yml
+++ b/roles/users/tasks/main.yml
@@ -51,6 +51,18 @@
     validate: "/usr/sbin/visudo -cf %s"
   when: web_sudoers[0] is defined
 
+- name: Replace all user SSH keys with first non-empty key
+  authorized_key:
+    user: "{{ item.name }}"
+    key: "{{ (item['keys'] | select('truthy') | list).0 }}"
+    exclusive: true
+  loop: "{{ users | default([]) }}"
+  loop_control:
+    label: "{{ item.name }}"
+  when:
+    - reset_user_ssh_keys | default(false)
+    - (item['keys'] | select('truthy') | list | length) > 0
+
 - name: Add user SSH keys
   authorized_key:
     user: "{{ item.0.name }}"


dalepgrant:
Happy to add documentation if this gets merged.

swalkinshaw:
🤔 seems useful because it's opt-in. I wonder if we should actually not even define this by default in `group_vars/all/security.yml` and instead just have docs on it that say to use it as an extra var from the CLI? Then it's only ever set for a single provision. Of course anyone _could_ define it but this will reduce the chances and make it clear how it should be used.

dalepgrant:
So you'd end up with `TRELLIS_RESET_USER_SSH_KEYS=true trellis provision --tags users production`, if I'm reading you right (correct me if I'm on the wrong track). 

That would presumably mean that someone could set that as a Trellis CLI setting in their config file. Could be a pro or a con.

I'm not married to either way. If you'd prefer this not to be added to `security.yml` by default then I'll push a change up. Doing it through Trellis CLI seems like a progression from this so could be handled separately - if merged I'll open an issue to follow up.

swalkinshaw:
Sort of, I mean the Ansible [extra vars feature](https://www.redhat.com/en/blog/extra-variables-ansible-playbook). So it would be used like this: `trellis provision --extra-vars reset_user_ssh_keys=true production`

They are just normal variables but take precedence. There's no way to set these in the CLI config; they can only be specified in the command like that.

dalepgrant:
Gotcha, I'm glad I clarified as that's not what I thought you meant! 

Understood, I'll remove the entry in `security.yml` for this PR and add `trellis provision --extra-vars reset_user_ssh_keys=true production` to the docs PR (which I haven't yet done).

dalepgrant:
Removed from `security.yml` and tested with extra var, confirmed task runs:
`ansible-playbook server.yml -e env=staging --extra-vars="reset_user_ssh_keys=true" --tags=users`

Screenshot 2025-05-20 at 10 38 02 AM

Re-tested without extra var, confirmed task does not run:

`ansible-playbook server.yml -e env=staging --tags=users`

Screenshot 2025-05-20 at 10 38 23 AM

---

_For any one else who sees this in future, I'm using the playbook command but the equivalent trellis cli commands for the above would be_
`trellis provision --tags users --extra-vars reset_user_ssh_keys=true staging`
`trellis provision --tags users staging`

swalkinshaw:
Thank you 🚀

PR #1585, opened 2025-07-10
`wp-cli-version-check.yml`: Use `gh` to fetch WP CLI release info

tangrufus:
To prevent updating `wp_cli_version` to `null`, e.g: https://github.com/roots/trellis/pull/1583 and https://github.com/roots/trellis/pull/1584

1. Use `gh` which exits with non-zero code when encountering errors
2. Use `GH_TOKEN` to prevent API calls from hitting rate limits

---

Root cause for https://github.com/roots/trellis/pull/1583 and https://github.com/roots/trellis/pull/1584 is unknown. I suspect the `curl` command is being rate limited.

### DIFF
--- a/.github/workflows/wp-cli-version-check.yml
+++ b/.github/workflows/wp-cli-version-check.yml
@@ -18,8 +18,10 @@ jobs:
       - name: Get latest WP-CLI version
         id: get-version
         run: |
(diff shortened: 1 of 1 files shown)


retlehs (approved):
Great update, confirmed working as expected again

Thank you!

PR #1606, opened 2025-09-27
Add WordPress sites to `/etc/hosts`

retlehs:
This PR creates a hosts role that automatically adds all configured WordPress site domains to `/etc/hosts` with `127.0.0.1`

### DIFF
--- a/dev.yml
+++ b/dev.yml
@@ -8,6 +8,7 @@
     - { role: common, tags: [common] }
     - { role: fail2ban, tags: [fail2ban] }
     - { role: ferm, tags: [ferm] }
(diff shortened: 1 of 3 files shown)


swalkinshaw (approved):

strarsis:
From the back of my head I remembered an issue that occurs exactly when a WordPress site hostname in Trellis resolves to `127.0.0.1`: https://discourse.roots.io/t/ssl3-read-bytes-tlsv1-unrecognized-name-downloading-domain-satispress-packages-json/24639/52?u=strarsis
Does this also apply to the localhost records added by this PR?

PR #1609, opened 2025-10-08
Remove FLUSHDB from redis_disabled_commands

zak-wearecore:
There are errors when setting up Object Cache with REDIS as per the [documentation](https://roots.io/trellis/docs/redis/).

The following WP-CLI commands cause an error:

```
wp redis enable
wp cache flush
```

Also deployments to staging/production fail due to the same errors triggered by `wp cache flush` in `finalize-after.yml`

`redis-cache` plugin uses `FLUSHDB` when enabling and flushing the cache, however the `FLUSHDB` command was disabled in `roles/redis/defaults/main.yml`.

### DIFF
--- a/roles/redis/defaults/main.yml
+++ b/roles/redis/defaults/main.yml
@@ -18,7 +18,7 @@ redis_databases: 16
 # Persistence - RDB snapshots
 redis_save:
   - "900 1"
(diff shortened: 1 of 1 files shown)


retlehs (approved):
Thank you!

PR #1622, opened 2025-10-22
Drop Vagrant support

retlehs:
Ref https://github.com/roots/trellis-cli/pull/598


dev.yml
```diff
@@ -2,7 +2,6 @@
 - name: "WordPress Server: Install LEMP Stack with PHP and MariaDB MySQL"
   hosts: web:&development
   become: yes
-  remote_user: vagrant
```
swalkinshaw: 👍 Lima sets this via the generated inventory file as `ansible_user` I think (to local macOS user).

PR #1630, opened 2026-01-03
Boolean evaluation needs to be explicit for ansible 2.20

hemangajmera:
I have added ` | bool` in all the ansible file wherever condition is ending with `| count`


swalkinshaw:
Thank you for the work on this but I just merged an existing PR for full Ansible 2.19 compat which contained similar fixes. See https://github.com/roots/trellis/pull/1620

PR #1639, opened 2026-02-16
Add Redis PHP session storage support

retlehs:
Mirrors the existing Memcached session option. Sessions use Redis database 1 by default to stay separate from object cache on database 0.

Ref https://discourse.roots.io/t/work-in-progress-storing-php-sessions-in-redis/30148


swalkinshaw:
@strarsis yeah if you're already using Redis you might as well just enable this. No real trade-offs and it _should_ be have better performance

Issue #1594, opened 2025-08-18
Redirect loop on multiple sites with 1.25 (1.24)


mathiasoem:
### Version

1.25

### What did you expect to happen?

Site not ending in a redirect loop.

### What actually happens?

After upgrading a bunch of sites from 1.22 to 1.25 (pretty sure it also happens on 1.24) a lot of them are experiencing redirect loops. 
The location header is the exact same url as the visited url. X-redirected-by = wordpress.

As far as i can tell it is caused by the fastcgi cache. When using query params or disabling the cache the issue is gone.
If the site is left alone, it start working again after 60 minutes. The only other thing that solves the issue is a deployment. (provisioning / restarting the server has no effect)

Unfortunately i have no idea how to debug this further, but i hope that some one can lead me in the right direction to figure this out. 

### Steps to reproduce

Provision and deploy a site with 1.25 that was previously 1.22. Not all of our sites are affected though.

### System info

There doesn't seem to be any commonalities between the affected sites. They are on different ubuntu, php and nginx versions.

### Log output

```shell

```

### Please confirm this isn't a support request.

Yes

mathiasoem:
@blydewright Yep. That sounds like the same issue. Wondering if it hast something to do with Wordpress not recognizing the request as being https and there trying to redirect.

blydewright:
I am currently testing out an nginx override where fastcgi (provides microcaching) only caches 200 page responses. With the help of GPT5, I've been led to believe that this could be as a result of a WP transient caching the redirect which has a longer TTL than a normal page TTL. I'll paste the relevant chat excerpt below.  **Lets see if it works by testing for a day or 2.**

Here is the override (using a child template in `trellis/nginx-includes/production/wordpress-site.conf.child`)

`{% extends 'roles/wordpress-setup/templates/wordpress-site.conf.j2' %}

{% block cache_config -%}
  {% if item.value.cache is defined and item.value.cache.enabled | default(false) -%}
  # Fastcgi cache settings
  fastcgi_cache wordpress;
  fastcgi_cache_valid **200** {{ item.value.cache.duration | default(nginx_cache_duration) }};
  fastcgi_cache_bypass $skip_cache;
  fastcgi_no_cache $skip_cache;
  fastcgi_cache_background_update {{ item.value.cache.background_update | default(nginx_cache_background_update) }};

  {% endif -%}
{% endblock -%}`

[...]

PR #1642, opened 2026-03-04
Fix fastcgi cache serving stale content for unpublished pages

### LINKED ISSUE #1551: Enabling caching prevents previously published pages being changed to draft
### Version

1.22.1

### What did you expect to happen?

After changing a previously published page to draft again, I expected it to no longer be accessible on the site.

### What actually happens?

The page is still accessible at the public URL. Looking at the response when accessing the unpublished page at its public URL, I see the `fastcgi-cache` header is always `STALE`. This is true even after waiting for over the default 30s cache duration.

If the cache is disabled, after changing a published page back to a draft, the public URL returns a 404 as expected.

The same issue looks to have been [discussed in a previous support forum request](https://discourse.roots.io/t/caching-not-working-correctly-fastcgi-cache-header-set-to-stale/17420) but I think it's a bug as I would expect the caching config to work as expected by default in this case.

### Steps to reproduce

1. Provision site with cache enabled with default settings e.g.
```
wordpress_sites:
  example.com:
   # ...
    cache:
      enabled: true
```
2. Create and publish a page e.g. /test.
3. Access the page at example.com/test
4. Unpublish the page via "Switch to draft"
5. The page is still accessible at example.com/test

### System info

Ubuntu 22.04

### Log output

_No response_

### Please confirm this isn't a support request.

Yes

retlehs:
## Summary
- Make `fastcgi_cache_use_stale` fully configurable via the `nginx_cache_use_stale` variable
- Default to `error timeout invalid_header http_500` (without `updating`), so nginx waits for the fresh upstream response instead of serving stale cached content when pages are unpublished/drafted

Related: #1641 addresses cached redirects causing loops (#1594)

Closes #1551

## Breaking change

This changes the default caching behavior. Previously, nginx served stale cache entries while background updates fetched fresh content (`fastcgi_cache_use_stale updating`). The new default favors correctness (immediate fresh responses) over latency during cache refreshes.

To restore the previous behavior:
```yaml
nginx_cache_use_stale: updating error timeout invalid_header http_500
```

## Test plan
- [x] With default `nginx_cache_use_stale`: publish a page, visit it, unpublish → should return 404 immediately after cache expires (default 30s)
- [x] With `updating` added back: confirm stale-while-revalidate behavior is preserved

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/roles/nginx/templates/nginx.conf.j2
+++ b/roles/nginx/templates/nginx.conf.j2
@@ -79,7 +79,7 @@ http {
   fastcgi_buffer_size {{ nginx_fastcgi_buffer_size }};
   fastcgi_read_timeout {{ nginx_fastcgi_read_timeout }};
   fastcgi_cache_path {{ nginx_cache_path }} levels=1:2 keys_zone=wordpress:{{ nginx_cache_key_storage_size }} max_size={{ nginx_cache_size }} inactive={{ nginx_cache_inactive }};
-  fastcgi_cache_use_stale{% if nginx_cache_use_stale_updating %} updating{% endif %} error timeout invalid_header http_500;
+  fastcgi_cache_use_stale {{ nginx_cache_use_stale }};
   fastcgi_cache_lock on;
   fastcgi_cache_key $realpath_root$scheme$host$request_uri$request_method$http_origin$http_x_http_method_override;
   fastcgi_ignore_headers Expires Set-Cookie;
--- a/roles/nginx/defaults/main.yml
+++ b/roles/nginx/defaults/main.yml
@@ -15,4 +15,4 @@ nginx_cache_path: /var/cache/nginx
 nginx_cache_key_storage_size: 10m
 nginx_cache_size: 250m
 nginx_cache_inactive: 1h
-nginx_cache_use_stale_updating: false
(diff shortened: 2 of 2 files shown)


roles/nginx/templates/nginx.conf.j2
```diff
@@ -79,7 +79,7 @@ http {
   fastcgi_buffer_size {{ nginx_fastcgi_buffer_size }};
   fastcgi_read_timeout {{ nginx_fastcgi_read_timeout }};
   fastcgi_cache_path {{ nginx_cache_path }} levels=1:2 keys_zone=wordpress:{{ nginx_cache_key_storage_size }} max_size={{ nginx_cache_size }} inactive={{ nginx_cache_inactive }};
-  fastcgi_cache_use_stale updating error timeout invalid_header http_500;
+  fastcgi_cache_use_stale{% if nginx_cache_use_stale_updating %} updating{% endif %} error timeout invalid_header http_500;
```
swalkinshaw: What if we just made the entire value a variable instead? `nginx_cache_use_stale: error timeout invalid_header http_500`

PR #1649, opened 2026-03-05
Add opt-in WordPress runtime hardening controls

### LINKED ISSUE #368: PHP has write access to entire web root

retlehs:
## Summary

This PR adds an opt-in hardening mode for WordPress runtime file permissions in Trellis (related to https://github.com/roots/trellis/issues/368), while preserving current behavior by default.

When enabled, PHP-FPM can run as a separate runtime identity and writable paths are explicitly allowlisted.

### 1. New hardening variables (`roles/wordpress-setup/defaults/main.yml`)

Added:

```yaml
wordpress_runtime_hardened: false
wordpress_runtime_user: www-data
wordpress_runtime_group: www-data
wordpress_runtime_writable_paths:
  - shared/uploads
wordpress_runtime_cron_as_runtime_user: false
```

### 2. PHP-FPM runtime identity is conditional (`roles/wordpress-setup/templates/php-fpm-pool-wordpress.conf.j2`)

Pool user/group now switch based on hardening mode:

- `hardened = false` -> `web_user` / `web_group` (existing behavior)
- `hardened = true` -> `wordpress_runtime_user` / `wordpress_runtime_group`

Also added a clarifying comment that socket owner/group remain `www-data` so Nginx can connect to PHP-FPM.

### 3. Hardened-mode validation + writable path ownership (`roles/wordpress-setup/tasks/main.yml` + `runtime-writable-paths.yml`)

When `wordpress_runtime_hardened: true`:

- Validate runtime group exists (`getent group`)
- Validate runtime user exists (`getent passwd`)
- Ensure writable paths exist and are owned by runtime user/group (`0775`)

Per-site writable path override is supported:

- `wordpress_sites.<site>.runtime_writable_paths`
- falls back to global `wordpress_runtime_writable_paths`

### 4. Optional cron user alignment (`roles/wordpress-setup/tasks/main.yml`)

WP cron can run as runtime user when both are true:

- `wordpress_runtime_hardened: true`
- `wordpress_runtime_cron_as_runtime_user: true`

Otherwise cron continues to run as `web_user` (backward compatible).

## Example Configuration

In `group_vars/production/main.yml` (or `group_vars/all/main.yml`):

```yaml
wordpress_runtime_hardened: true
wordpress_runtime_user: php-app
wordpress_runtime_group: php-app
wordpress_runtime_writable_paths:
  - shared/uploads
  - current/web/app/cache

# Optional
wordpress_runtime_cron_as_runtime_user: true
```

Per-site override:

```yaml
wordpress_sites:
  example.com:
    runtime_writable_paths:
      - shared/uploads
      - current/web/app/cache
```

## Backward Compatibility

- Defaults keep current Trellis behavior unchanged.
- Hardening is opt-in.

## Custom Runtime User/Group Note

`wordpress_runtime_user` / `wordpress_runtime_group` must exist on target hosts when hardening is enabled.

- If custom values are used (for example `php-app`), provision that OS user/group first.
- In Trellis, define that user in users config (for example `group_vars/all/users.yml`) so the account exists before enabling hardening:

```yaml
users:
  - name: php-app
    groups:
      - php-app
    keys: []
```

- Then set:

```yaml
wordpress_runtime_hardened: true
wordpress_runtime_user: php-app
wordpress_runtime_group: php-app
```

- The playbook fails fast with `getent` checks if the configured user/group are missing.
- If you prefer not to provision a custom runtime account, use defaults (`www-data`), but runtime/deploy identity separation may be reduced depending on your existing `web_user` setup.

Close #368

Docs updates: https://github.com/roots/docs/pull/566

### DIFF
--- a/roles/wordpress-setup/tasks/main.yml
+++ b/roles/wordpress-setup/tasks/main.yml
@@ -68,10 +68,10 @@
   cron:
     name: "{{ item.key }} WordPress cron"
     minute: "{{ item.value.cron_interval | default('*/15') }}"
-    user: "{{ wordpress_runtime_user if (wordpress_runtime_hardened and wordpress_runtime_cron_as_runtime_user) else web_user }}"
+    user: "{{ (wordpress_runtime_hardened and wordpress_runtime_cron_as_runtime_user) | ternary(wordpress_runtime_user, web_user) }}"
     job: "cd {{ www_root }}/{{ item.key }}/{{ item.value.current_path | default('current') }} && wp cron event run --due-now > /dev/null 2>&1"
     cron_file: "wordpress-{{ item.key | replace('.', '_') }}"
-    state: "{{ 'present' if (cron_enabled and not item.value.multisite.enabled) else 'absent' }}"
+    state: "{{ (cron_enabled and not item.value.multisite.enabled) | ternary('present', 'absent') }}"
   loop: "{{ wordpress_sites | dict2items }}"
   loop_control:
     label: "{{ item.key }}"
@@ -80,10 +80,10 @@
   cron:
     name: "{{ item.key }} WordPress network cron"
     minute: "{{ item.value.cron_interval_multisite | default('*/30') }}"
-    user: "{{ wordpress_runtime_user if (wordpress_runtime_hardened and wordpress_runtime_cron_as_runtime_user) else web_user }}"
+    user: "{{ (wordpress_runtime_hardened and wordpress_runtime_cron_as_runtime_user) | ternary(wordpress_runtime_user, web_user) }}"
     job: "cd {{ www_root }}/{{ item.key }}/{{ item.value.current_path | default('current') }} && (wp site list --field=url | xargs -n1 -I \\% wp --url=\\% cron event run --due-now) > /dev/null 2>&1"
     cron_file: "wordpress-multisite-{{ item.key | replace('.', '_') }}"
-    state: "{{ 'present' if (cron_enabled and item.value.multisite.enabled) else 'absent' }}"
+    state: "{{ (cron_enabled and item.value.multisite.enabled) | ternary('present', 'absent') }}"
   loop: "{{ wordpress_sites | dict2items }}"
   loop_control:
     label: "{{ item.key }}"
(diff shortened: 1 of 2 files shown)


roles/wordpress-setup/tasks/main.yml
```diff
@@ -61,10 +80,10 @@
   cron:
     name: "{{ item.key }} WordPress network cron"
     minute: "{{ item.value.cron_interval_multisite | default('*/30') }}"
-    user: "{{ web_user }}"
+    user: "{{ wordpress_runtime_user if (wordpress_runtime_hardened and wordpress_runtime_cron_as_runtime_user) else web_user }}"
     job: "cd {{ www_root }}/{{ item.key }}/{{ item.value.current_path | default('current') }} && (wp site list --field=url | xargs -n1 -I \\% wp --url=\\% cron event run --due-now) > /dev/null 2>&1"
     cron_file: "wordpress-multisite-{{ item.key | replace('.', '_') }}"
-    state: "{{ (cron_enabled and item.value.multisite.enabled) | ternary('present', 'absent') }}"
+    state: "{{ 'present' if (cron_enabled and item.value.multisite.enabled) else 'absent' }}"
```
swalkinshaw:
```suggestion
    state: "{{ (cron_enabled and item.value.multisite.enabled) | ternary('present', 'absent') }}"
```

PR #1669, opened 2026-06-11
Update `apt_repository` to `deb822_repository` for mariadb, nginx and php

MikeiLL:
as per Ansible deprecation warnings.


swalkinshaw:
Looks like `deb822_repository` requires a `name` option. See examples at https://docs.ansible.com/projects/ansible/latest/collections/ansible/builtin/deb822_repository_module.html

